PR #8, opened 2021-07-12
feat: Add ftdetect for norg filetype

shadmansaleh:
> Worried about lazy loading? It's very hard to do properly because of the way Neovim handles filetypes. I'm planning on adding a PR with support for Neorg files to the Neovim core at some point. 

Support in core isn't necessary for this. You can just add ftdetect for norg format and it'll work .Package managers to execute ftdetect files for lazyloaded plugins too . Lua is suported in core for ftdetect but not all plugin managers have support yet. For example packer doesn't yet run lua ftdetect files for lazyloading . It's planned though https://github.com/wbthomason/packer.nvim/issues/403

### DIFF
--- a/ftdetect/norg.vim
+++ b/ftdetect/norg.vim
@@ -1,0 +1,1 @@
+autocmd BufRead,BufNewFile *.norg setlocal filetype=norg


vhyrro:
Hey, is it alright if I make some changes to the PR? I want to go full lua, and therefore use norg.lua rather than norg.vim for ftdetect :)

shadmansaleh:
You can oviously make changes .

But as I've stated for now vimL is better 
> Lua is suported in core for ftdetect but not all plugin managers have support yet. For example packer doesn't yet run lua ftdetect files for lazyloading

Also since there's yet no way of doing auto commands directly from lua . You'll just have one line of `vim.cmd` .

There's really nothing significant to be gained by going 100% lua . I think it's better to use the best tool for the job . I think we should have it on vimL till `packer` doesn't add support :D

vhyrro:
Gotcha, makes sense. I misunderstood exactly how ftdetect operates, now I get it. I'll merge it as it is, and when packer gets that juicy support we will move to lua. Thanks for these last two pull requests :)

PR #9, opened 2021-07-12
Project: Remove gifs from version control

shadmansaleh:
Cloning this repo on slow connection is pretty terrible .
Basically 99% of that download is gifs that doesn't add anything to plugin instalation.

I've removed gifs from git and  put them in github and put links in README.

I think this should speed up installationa nd updating of this plugin .

@vhyrro what do you think?


shadmansaleh:
@vhyrro Also you'll probably have to remove the gifs from all commits to make .git/ not contain copies of them . And make full clone lighter .

You'll probably need to run something like
```sh
git filter-branch --index-filter 'git rm --cached --ignore-unmatch res/*.gif' HEAD
```
[source](https://stackoverflow.com/questions/35115585/remove-files-completely-from-git-repository-along-with-its-history)

molleweide:
Would it be possible to move them to another branch maybe?

shadmansaleh:
> Would it be possible to move them to another branch maybe?

Why ? The gifs are only used for README . It'll look the same . They will just be loaded from web instead of repo itself .

You can view the Changed Readme [here](https://github.com/shadmansaleh/neorg/tree/remove_gifs) . It's same as main btanches README .

brymer-meneses:
We could use github gists to host the gifs

shadmansaleh:
> We could use github gists to host the gifs

In this branch github is hosting them .

vhyrro:
I think the current way we're hosting via github is perfectly fine. Thanks for this, it should make installation a lot faster!

PR #20, opened 2021-07-15
Add vim-plug to README

syntezoid:
Aloha,
I added how to configure vim-plug with init.vim for initial configuration.


vhyrro:
Thanks for this! I had to reword a few bits n bobs to make it "just right" but other than that I'll be glad to merge

PR #34, opened 2021-07-27
Boot up Neorg on Bufnew

stevearc:
Right now if you `nvim test.norg` it opens up a new file and inserts the `@document.meta` at the top. Similarly, if you open nvim and then `:e existing_file.norg` it will do the same. However, if you open up nvim and `:e new_file.norg`, Neorg is not initialized and the file does not get the `@document.meta`


vhyrro:
Ah, that's something I overlooked. Works great! Thx man. Also big props for targeting `unstable` :P

PR #46, opened 2021-07-31
Add stylua

vuki656:
As discussed on discord I have added stylua to and formatted the whole project with it excluding `completion.lua` as there is a bug in stylua. I added a comment in `.styluaignore` with a reference to the issue.

Stylua is opinionated so it only supports a few configuration options which can be found in `.stylua.toml`. Here is the [table](https://github.com/JohnnyMorganz/StyLua#configuration) with options. Let me know if you prefer other config optins. 

If you have `efm` or `null-ls` configured, you would just need to add stylua to support it and format files. [Here](https://github.com/JohnnyMorganz/StyLua#installation) are the install instructions.


vhyrro:
Looks great, all the options seem right to me :)

I'll just have to update the CONTRIBUTING.md file to let others know to use stylua now, thx for this

PR #51, opened 2021-08-01
Enter Links on <CR>

vhyrro:
As an initial step in helping make the development of Neorg more transparent I've elected to start making draft PRs so people can easily chime in and give thoughts (maybe even contribute ;)).

The idea is simple - whenever the user has their cursor hovering over a link pressing <CR> will jump the user to the location of that link. It could even maybe provide some prompts to ask the user what to do if the link can't be found.

### DIFF
--- a/lua/neorg/modules/core/keybinds/default_keybinds.lua
+++ b/lua/neorg/modules/core/keybinds/default_keybinds.lua
@@ -15,6 +15,8 @@ return function(neorg_leader)
                 -- Keys for managing notes
                 { neorg_leader .. "nn", "core.norg.dirman.new.note" },
 
+                { "<CR>", "core.norg.esupports.goto_link" },
+
                 { "<C-s>", "core.integrations.telescope.find_linkable" },
 
                 { neorg_leader .. "toc", "core.norg.qol.toc.generate.toc" },
--- a/lua/neorg/modules/core/norg/esupports/module.lua
+++ b/lua/neorg/modules/core/norg/esupports/module.lua
@@ -51,7 +51,7 @@ function _neorg_indent_expr()
 end
 
 module.setup = function()
-    return { success = true, requires = { "core.autocommands" } }
+    return { success = true, requires = { "core.autocommands", "core.keybinds" } }
 end
 
 module.config.public = {
@@ -192,6 +192,8 @@ module.load = function()
     if module.config.public.indent_config.realtime.enabled then
         module.required["core.autocommands"].enable_autocommand("TextChangedI")
     end
+
+    module.required["core.keybinds"].register_keybind(module.name, "goto_link")
 end
 
 module.public = {
@@ -293,6 +295,13 @@ module.public = {
             end
         end
     end,
+
+    -- Other functions go here
+
+    goto_link = function()
+        log.warn("<CR> pressed")
+    end,
+
 }
 
 module.on_event = function(event)
@@ -333,8 +342,12 @@ module.on_event = function(event)
         module.public.indent_line()
     end
 
-    if event.type == "core.autocommands.events.bufwrite" then
+    --[[ if event.type == "core.autocommands.events.bufwrite" then
         -- TODO
+    end ]]
+
+    if event.split_type[2] == module.name .. ".goto_link" then
+        module.public.goto_link()
     end
 end
 
@@ -343,6 +356,10 @@ module.events.subscribed = {
         bufenter = true,
         textchangedi = true,
         bufwrite = false,
+    },
+
+    ["core.keybinds"] = {
+        [module.name .. ".goto_link"] = true,
     },
 }
 


vhyrro:
I'm going to first try and work on the TreeSitter PR with a custom lexer and such. If we could detect links using the TS lexer then the job would be insanely easy in terms of extracting information from links. Currently we have to do a tonne of crazy regex magic which would take me ages. Soz for that

mrossinek:
> If we could detect links using the TS lexer then the job would be insanely easy in terms of extracting information from links. Currently we have to do a tonne of crazy regex magic which would take me ages. Soz for that

I have not looked at the details here, but maybe this [markdown TS](https://github.com/ikatyang/tree-sitter-markdown) is helpful? They have a link to an online playground where the showcase a bunch of link syntax :+1:

vhyrro:
There's a reason why I don't go to the markdown treesitter parser for help ;)

Those cursed with the knowledge why will instantly catch what I'm talking about.

vhyrro:
I'm just gonna write here that I didn't die or anything, I'm simply still working on the TS parser :)

Things are going great, I'm excited to announce what we'll be able to do with the new generated syntax trees when I'm done

vhyrro:
Since I'm a bit stuck on the TS parser (but I've managed to implement link detection) I've started writing the barebones. It's a bit messy as of now and I'll clean it up later but hey it extracts stuff properly! Epic

vhyrro:
Ok so the code is a tiny bit messy rn however we have support for entering a link with `<CR>` on "basic" links (i.e. links that only link to an element in the current file). The next thing to implement will be cross-file searching and I'm excited to implement that! Note that if you're gonna want to try this PR out you're also going to have to be running the new experimental TS parser that can be found in the `rewrite` branch in its respective repo. The new experimental parser doesn't have all features implemented yet though so you've been warned :)

I'll post a video of it in action tomorrow for those who aren't bothered trying it out for themselves

vhyrro:
Here's the showcase I promised of what we have so far. As you can see all the basics work fantastic.

https://user-images.githubusercontent.com/76052559/129372633-40a688de-a24e-4c52-a3de-32e21158dd86.mp4

vhyrro:
Alright good news lads. The base features are complete, including cross-file search! Here's the syntax:

```
[a link](*A heading somewhere)
^ This you should be used to, it's just a regular link pointing to a first-level heading in the current document. You don't need to worry about being case sensitive but make sure you keep the punctuation in tact! Escape special chars with a backslash \
```
---
```
[a link](:file1:file2:*A heading somewhere)
```
This is the syntax for searching in other files. What the above syntax says is "search in file1 and file2 in the *current directory of the current file*. This means that if we're inside of `work/index.norg` and we try to search in `important.norg` it'll search in `work/important.norg` because `index.norg` is in the `work` directory. If you want to search from the root of the workspace do `/important.norg` rather than just `important.norg`. You can also omit the `.norg` extension entirely, neorg will figure it out. To summarize:

```
[link to something.norg in the root of the workspace](:/something:*A heading in the something.norg file)
```

### TODO:
- Allow the user to select an action when a link can't be found:
    - Create the file and place the linkable at the top of the file
    - Place the linkable at the top of the file
    - Place the linkable at the bottom of the file
    - Place the linkable in a sibling node above the current link's parent node
    - Place the linkable in a sibling node below the current link's parent node
    - Enter a custom mode that will allow you to freely move around the file and hit `<CR>` to place the linkable, then return back to the location of the link.
    - Do nothing

vhyrro:
I am incredibly proud to show you the latest and greatest thing I've been working on! There's a couple of things I wanna show:

a) Added support for searching for drawers!
b) Added "actions", aka operations that can be performed if a link cannot be found. Currently the only actions that work are `do nothing` (lol) and `attempt to fix the link`. And oh boy let me tell you there's some real witchcraft that happens during the second one. Like I'm not exaggerating I'm really proud of it lol, here's how it works:

### Fixing methods
When you attempt to fix a link, you're provided with two ways of fixing said link. The first is "fuzzy type" fixing, where Neorg will disregard the type of link (i.e. whether it's referencing a first-level heading or a marker or whatever) and just find the first element that most closely matches the misspelled link location. This is probably the most OP one. 

The second type is "strict type" fixing, where Neorg will find the first element that most closely matches the provided destination that is also of the same type. This means that if you're pointing to a second-level heading (`[test](**location)`) it'll only fuzzy search second-level headings.

### How it works under the hood
The algorithm for matching is a multi-layered one. At its base it uses Levenshtein distance to determine how similar the "incorrect" string is to any potential target.
     
It is used to locate accidental spelling mistakes or missing/excessive characters (e.g. it could consider `hello` and `hllo` to be very similar but will consider `hello` and `test` to be very different. On top of that Neorg adds two extra layers in order to find the closest match - but that's kind of mumbo jumbo and I'm not bothered to talk about it. What matters is that IT WORKS and it works surprisingly well (unless you try being very ambiguous with it, like trying to fuzzy match `tt st`, then you can expect a few mistakes if there are several similar matches).

### Showcase
Here's a video showing it in action. Apologies if I feel very disoriented and slow in the video it's just really late and I can't think straight haha, I wanted to get this out to y'all:

https://user-images.githubusercontent.com/76052559/130514159-63e1a5bd-dc54-49f1-a040-ba4d9e87cb02.mp4

Hope you enjoy this feature! As you can see from the video above it can even find matches present in multiple files :D

Issue #69, opened 2021-08-24
Completion Source for `cmp`


vhyrro:
Welp, it happened! I've known about `nvim-cmp` for a pretty long time now but I didn't think the deprecation notice would come this early. Will create a new completion source for it tomorrow so you don't have to use old stuff (ew)

marty-oehme:
I'm not sure where else in the README it exists, but in the wiki page for the completion module - 
https://github.com/nvim-neorg/neorg/wiki/Completion -

it does not provide an example of the necessary configuration posted above. 
If entering a neorg file without the 

```
["core.norg.completion"] = {
    config = {
        engine = "nvim-cmp"
    }
}
```

settings, it will complain about no engine being set.

PR #70, opened 2021-08-24
Integration with nvim-cmp

JoeyGrajciar:
[hrsh7th/nvim-cmp](https://github.com/hrsh7th/nvim-cmp) is set to be successor to nvim-compe

nvim_cmp_neorg
#69

### DIFF
--- a/lua/neorg/modules/core/integrations/nvim-cmp/module.lua
+++ b/lua/neorg/modules/core/integrations/nvim-cmp/module.lua
@@ -1,0 +1,75 @@
+--[[
+        A module for integrating nvim-cmp with Neorg
+        Works with core.norg.completion to provide intelligent completions
+--]]
+
+require("neorg.modules.base")
+
+local module = neorg.modules.create("core.integrations.nvim-cmp")
+
+module.private = {
+  source = {},
+  cmp = {},
+  completions = {},
+}
+
+module.load = function ()
+  local success, cmp = pcall(require, "cmp")
+
+  assert(success, "nvim-cmp not found, aborting...")
+
+  module.private.cmp = cmp
+end
+
+module.public = {
+  create_source = function ()
+
+    module.private.source.new = function ()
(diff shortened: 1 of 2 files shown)


MordechaiHadad:
godly

vhyrro:
Wow, you beat me to it! That's damn awesome. I'll be testing this out today.

mrossinek:
I'll be testing this from now on. Will give feedback later this week :+1: 

- https://github.com/mrossinek/dotfiles/commit/0eeb8980bfd46b7dc28e4f43fc7fdcd1bba7fe56
- https://github.com/mrossinek/dotfiles/commit/5c3ce555e0c36b6dcebe8196c7ba67259b163b69

vhyrro:
Alright, from my testing all looks good! A few things though, in order to keep consistency I'd probably name the source `neorg` rather than `norg` (as it is in the `compe` integration). Also it seems that the completion kind isn't working as intended. I tried looking into it a bit but I'm gonna be honest I don't know exactly what's going on. It seems we can only use prebuilt types provided by cmp?

vhyrro:
I'll gladly merge once the source name inconsistency (norg vs neorg) is resolved if that's alright :)

JoeyGrajciar:
Sure thing will do the change also with some README update and  those types.
Another thing I was thinking about doing is some kind of snippet support e.g.
```
@code $1
  $2
@end
```

vhyrro:
Hi, if that's alright I jumped ahead a bit and did some of the tedious stuff for you - I updated the readme and made it so that the user has to now manually specify a completion engine that they may wanna use :). Also changed the completion source name.

vhyrro:
Yeah, that's the idea. I know that for some it may be a bit tedious because it's one more thing to set up but it's literally 3 LOC. Additionally it becomes less annoying for people who aren't using a supported completion engine and are getting constant errors. It's better to have to enable something than to have to disable something because it's annoying/forcing a certain completion engine down your throat imo.

Issue #55, opened 2021-08-01
DOCUMENTATION UPDATES


vhyrro:
I believe it's important that I keep this issue up until the release of 0.1. As we approach that milestone closer and closer every day I can't possibly keep updating the docs every single time I make a small change. That's why I've decided to just stop updating the documentation until I feel that our feature set is where I want it to be for the first release. After that happens I will then overhaul the README and the wiki. For you this unfortunately means that without some experience in reading code you're gonna have a bit of a bad time if you're using `unstable` :P

If I make a big or significant change then I will post it here. I know it's a bit late but it just occurred to me that this is a good method of informing users about changes without having to update the entirety of the docs lol

I think the most important changes are related to the file format:
- `/this/` is now italic
- `_this_` is underline
- [**NOTE**: This is a temporary change] tags (marked with `@`) need to stay at the beginning of a line. You'll see that if you have realtime indentation enabled Neorg will do this for you. The reason is that until our magical treesitter parser does not become waay more intelligent parsing the contents of tags is an absolute disaster. You have to account for extra indentation, tabs vs spaces etc.. It's horrid.

After all wiki documentation is complete all of the markdown files will be ran through a converter to convert them to vim help files so you can easily access them from the comfort of your favourite editor :)

mrossinek:
Some things we gotta do in terms of docs before releasing 0.1:
- [ ] update module docs/wiki (now auto-generated from source code docs)
- [ ] update README
- [ ] update GIFs
- [ ] update vim help pages
  - [ ] ensure vim help page tags are **sorted**

(I don't aim for this to be a comprehensive list, but at least we have some check points to start from...)

vhyrro:
Honestly I'd say our new Neorg news system is good enough to be able to directly replace this thread - we can broadcast info to thousands of users at once, which is much better than having to get people to subscribe to this almost dead thread anyway haha

I propose we close this issue to prevent it from cluttering up our issues section. Lemme know what you guys think.

cc @mrossinek and @danymat

mrossinek:
Yep, sounds good to me! I'll just go ahead and do this :smile:

PR #80, opened 2021-08-27
Overhaul example.norg

mrossinek:
I overhauled the example to be up-to-date with the `unstable` branch.
I also tried to make it a more easy-going and enjoyable read compared to the full spec.
Hopefully this helps in resolving issues like #55 and #79


vhyrro:
As always awesome work. Indeed it's a great step up!

Issue #81, opened 2021-08-27
[BUG] Assortment of bugs


mrossinek:
### Contributions

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)

### Plugin Version

- [X] I am using the latest version of the plugin

### Neovim Version

- [X] I am using the latest neovim version (0.5)

### Neovim Version

NVIM v0.6.0-dev+1667-g32024787b

### Log Output

Not applicable

### Branch

unstable

### Loaded modules

Here is my setup:
```lua
require('neorg').setup {
    load = {
        ["core.defaults"] = {},
        ["core.highlights"] = {},
        ["core.integrations.telescope"] = {},
        ["core.integrations.treesitter"] = {
            config = {
                highlights = {
                    Heading = {
                        ["1"] = {
                            Title = "+TSTitle",
                            Prefix = "+TSTitle",
                        },
                    },
                },
            },
        },
        ["core.keybinds"] = {},
        ["core.norg.concealer"] = {
            config = {
                icons = {
                    heading = {
                        level_1 = {
                            icon = "◉",
                        },
                        level_2 = {
                            icon = "◎",
                        },
                        level_3 = {
                            icon = "○",
                        },
                        level_4 = {
                            icon = "✺",
                        },
                    },
                },
            },
        },
        ["core.norg.completion"] = {
            config = {
                engine = "nvim-cmp",
            },
        },
        ["core.norg.dirman"] = {
            config = {
                workspaces = {
                    org = "~/Org",
                },
                autodetect = true,
                autochdir = true,
            },
        },
        ["core.norg.esupports"] = {},
    },
}
```

### Expected behavior

This is an assortment of bugs which I encountered while working on #80.

### Escaped markup not always respected
The vim-regex responsible for highlighting attached modifiers can break in cases
where the char is actually escaped:
![screenshot_1630081877](https://user-images.githubusercontent.com/21973473/131167697-43a05d37-f270-45a9-b4e3-c40a5b00e183.png)
- [x] Fixed in #84 

### Concealing of `-` broken
`-` chars at the beginning of lines are being consumed. This breaks unordered
lists, tasks and links as well as the `---` delimiting modifier' appearances.
![screenshot_1630085927](https://user-images.githubusercontent.com/21973473/131167708-d54b715c-5fd1-43ec-a7bb-bded0ca4de57.png)

I believe the reason for this is that pairs like `--` are seen as empty
[...]

PR #84, opened 2021-08-27
Some concealing fixes

mrossinek:
Fixes the first two bugs described in #81.
This also adds the `€` sign to the list of punctuation as previously suggested on Discord.


vhyrro:
Yup, all looks great! Epic

PR #88, opened 2021-09-01
Feature: Dynamically Generate and show Table of Contents

vhyrro:
Helo

This PR is gonna allow you to preview, browse and interact with a dynamically generated table of contents.
As you may or may not know we removed the ability to generate static ToCs a while back. The reasoning is because we favoured the more dynamic version instead:
```
= TOC <Optional custom heading name for the TOC>
```
The drawback that this came with is that now you can't really view the Table of Contents, neither can you jump to locations in a large document efficiently. @mrossinek gave some nice ideas that I will be using regarding the look, feel and usability of the UI, so credits to him! We would like to provide a keybind that will open a vertical split. This split will contain the actual ToC. You will be able to jump to different locations in your document from that ToC, you'll be able to fold different things, see things update in real time, modify the names of headings from within the table of contents, and so on and so forth.

As of right now, this is what we have:
ToC

Very basic, but the barebones are there. I'm going to be expanding this a bunch over the course of the next week, maybe two :). I'll also be helping out with the GTD PR in the meantime so if you don't see activity here for a longer amount of time then now you know.


vhyrro:
Damn, gonna have to rebase lol. Made this PR merge to main rather than unstable

mrossinek:
As discussed on Discord, #159 will supersede this PR.

PR #92, opened 2021-09-05
Basic neorg help-page

mrossinek:
This help-page is essentially a copy of the `EXAMPLE.norg` file.
However, it is accompanied by a `tags` file which vim understands and
parses to extract help tags.
In the future, an automatic generation of this tags file will be added.
For now, we have to update it manually. But the flex we can achieve with
having norg-formatted help pages is definitely worth that! 💪

Peek_2021-09-05_11-17

### DIFF
--- a/doc/neorg.norg
+++ b/doc/neorg.norg
@@ -1,0 +1,221 @@
+@document.meta
+	title: Neorg help-page
+	description: the Neorg vim help-page
+	author: The Neorg-community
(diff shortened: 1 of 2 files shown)


danymat:
Omg this is so cool 
I can't wait to copy pasta the gtd specs lol

vhyrro:
Such epicness :purple_heart:

PR #96, opened 2021-09-06
Conceal all quote levels individually

mrossinek:
screenshot_1631033271

As always, everything is configurable to your hearts content :+1: 

EDIT: new screenshot

### DIFF
--- a/lua/neorg/modules/core/integrations/treesitter/module.lua
+++ b/lua/neorg/modules/core/integrations/treesitter/module.lua
@@ -217,28 +217,28 @@ module.config.public = {
 
         Quote = {
             ["1"] = {
-                [""] = "+Blue",
+                [""] = "+TSPunctDelimiter",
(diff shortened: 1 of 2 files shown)


vhyrro:
This is great. One smol thing though - it's best not to bind to hl groups like Red and Green and Blue. I made this same mistake once and highlights stopped working on several popular colourschemes 😅. We might wanna move to colours that work on all colourschemes instead (like TS highlight groups or something similar)

mrossinek:
Alright fair enough! I will look into that tonight.

I also had to use `quote_level_?` names rather than just `level_?` nested within the `quote` table because I think the nested name must be unique. I was wondering whether instead we can construct a concatenated name for the private icons table like e.g. `heading.level_1` and `quote.level_1`

I can also look into adding that in this PR if that's something you think is worthwhile :+1:

PR #100, opened 2021-09-08
Add more intelligent TODO item toggling

vhyrro:
This PR makes TODO items bind themselves to each other and all in all makes them a lot more intelligent.
showcase

There's still a few issues that need fleshing out:
- [x] The parent element should be marked as pending if one of the children is pending
- [ ] ~~Editing the text doesn't update anything, only using the keybinds does (make this realtime updating as-you-type thing configurable)~~
- [x] Toggling between items with `<C-Space>`

I'll be working on those bit by bit and will try to merge this branch as quickly  as possible. There's a lot that I have to track and do simultaneously but this one should be easy.

### DIFF
--- a/lua/neorg/modules/core/defaults/module.lua
+++ b/lua/neorg/modules/core/defaults/module.lua
@@ -12,5 +12,6 @@ return neorg.modules.create_meta(
     "core.mode",
     "core.norg.qol.todo_items",
     "core.norg.esupports",
-    "core.integrations.treesitter"
+    "core.integrations.treesitter",
+    "core.norg.manoeuvre"
 )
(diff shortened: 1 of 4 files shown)


vhyrro:
Surprise feature!

New module has been added - `core.norg.maneouvre`. Allows you to move different elements up n down with `<M-j>` and `<M-k>`! Move todo items, list items and headings up and down the document. I think that's it for this PR to be honest, I don't feel like dealing with the second TODO item in this PR, we'll start handling realtime updates another day once people start shouting about it. For now I think it's safe to merge into unstable because, well, it's a branch for unstable things :P

PR #101, opened 2021-09-08
Conceal link modifier

mrossinek:
This adds support for concealing the link modifier, `:`. Since tree-sitter will not allow us to do concealing natively we will have to rely on our trusty Vim syntax regions defined via regexes anyways.

### Actual text
screenshot_1631138571
### Concealed
screenshot_1631138580

## Up for discussion
Right now, I implemented the link modifier regex to only require being suff- and prefixed with another markup modifier on the opening and closing link modifiers, respectively. However, this somewhat contradicts the current spec which requires a **non-whitespace character** in front of the opening modifier. However, I find that is a limitation because it will prevent e.g. the following: `This is :*im*:possible!`
I believe, that this is a not uncommon scenario so I am open for suggestions on how to rephrase the spec accordingly. I placed a temporary TODO comment at the relevant position of the markdown file.
As of right now, the spec does not contain a similar limitation for the closing `:`.

_PS: I also did some minor touch-ups on the other conceal elements_

### DIFF
--- a/docs/NFF-0.1-spec.md
+++ b/docs/NFF-0.1-spec.md
@@ -442,11 +442,20 @@ changes:
   This is the only modifier that does not fall into any of the 4 categories of modifiers (attached, detached, trailing and delimiting). The `:`
   symbol is marked the **link modifier**. It exists to link different "categories" of text together.
   The rules for the link modifier are a twist on the rules of the attached modifier, meaning a link modifier must form a pair.
-  [//]: # "TODO: discuss limitations of link modifier as defined below."
-  [//]: # "NOTE: this is mostly a temporary marker to allow me to easily place a review comment on Github."
-  An opening link modifier must be prefixed by a non-whitespace character and must be followed by an attached modifier or by whitespace (if at the end of a word).
-  Note that the link modifier is **not** an attached modifier, so chaining them like `::<attached modifier>` will simply result in the first colon
-  being treated as punctuation. A closing link modifier may only be prefixed with an attached modifier. The syntactical equivalent of markdown's `so*me*thing` will look as such: `so:*me*:thing`.
+  The opening link modifier must be immediately followed by an attached modifier.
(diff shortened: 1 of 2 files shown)


docs/NFF-0.1-spec.md
```diff
@@ -442,6 +442,8 @@ changes:
   This is the only modifier that does not fall into any of the 4 categories of modifiers (attached, detached, trailing and delimiting). The `:`
   symbol is marked the **link modifier**. It exists to link different "categories" of text together.
   The rules for the link modifier are a twist on the rules of the attached modifier, meaning a link modifier must form a pair.
+  [//]: # "TODO: discuss limitations of link modifier as defined below."
+  [//]: # "NOTE: this is mostly a temporary marker to allow me to easily place a review comment on Github."
   An opening link modifier must be prefixed by a non-whitespace character and must be followed by an attached modifier or by whitespace (if at the end of a word).
   Note that the link modifier is **not** an attached modifier, so chaining them like `::<attached modifier>` will simply result in the first colon
   being treated as punctuation. A closing link modifier may only be prefixed with an attached modifier. The syntactical equivalent of markdown's `so*me*thing` will look as such: `so:*me*:thing`.
```
mrossinek, line 449: I'll put up a suggestion myself:
```suggestion
  The opening link modifier must be immediately followed by an attached modifier.
  Accordingly, the closing link modifier must be immediately preceded by one.
  Note that the link modifier is **not** an attached modifier, so chaining
  them like `::<attached modifier>` will simply result in the first colon being
  treated as punctuation.
  There are no other constraints on the placement of link modifiers.
  Let's give some quick examples comparing with Markdown:

| Markdown       | norg             |
|----------------|------------------|
| `so*me*thing`  | `so:*me*:thing`  |
| `*im*possible` | `:*im*:possible` |
| `can*not*`     | `can:*not*:`     |

```
And here is a short concealed example of these words:
screenshot_1631139609

PR #104, opened 2021-09-09
Nested markup highlighting

mrossinek:
:warning: This requires the [attached-modifiers](https://github.com/nvim-neorg/tree-sitter-norg/tree/attached-modifiers) branch of the TS parser!

The TS parser is starting to support attached modifiers. These can be used for highlighting instead of the regex-based syntax regions which we use for concealing. (those will need to stay for the concealing only)

This PR starts to make use of them allowing 2-level nested basic markup which can look like this:
screenshot_1631203984

### Inline code
Markups nested within verbatim modifiers, `, are not being highlighted. This is a design choice which is currently not documented. @vhyrro what do you think about this?

### Current limitation

Right now, we can only support 2-level nesting because the TS parser cannot yet handle nested attached modifiers. Thus, the TS parser only detects the outer-most attached modifier providing highlighting for that.
We can add one additional level of highlighting by exploiting the highlight which we get from the syntax region defining the concealing.

In the future, once the TS parser supports nested attached modifiers, we will make the syntax regions `transparent` because we will be able to get the entire highlight definitions from TS.


vhyrro:
About the verbatim stuff: yeah we shouldn't do highlights inside of those since, well, it's `verbatim` 😄

mrossinek:
This will be superseded by #117. We cannot merge this before the `attached-modifiers` branch of the TS parser is done anyways and we are likely to fix nested attached modifiers on that before merging it.

Issue #98, opened 2021-09-08
[SUGGESTION] Reuse selection popup 


fdschmidt93:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

Hi there!

Cool project :) I've recently migrated and liking it a lot so far.

One thought I've had looking at #60 is to maybe reuse neogits popup builder system, which looks amazing. Ideally, the system could be upstreamed to plenary such that other projects (including neorg) can lever it. I'm suggesting this of course hoping neogit's maintainer would be OK with this (cc @TimUntersberger)

As initial references:

* Base module: https://github.com/TimUntersberger/neogit/blob/master/lua/neogit/lib/popup/builder.lua
* Push popup: https://github.com/TimUntersberger/neogit/blob/master/lua/neogit/popups/push.lua

### Help

No

### Implementation help

I'd be able to, but helping to maintain and adding features to telescope already takes a lot of my time :sweat_smile: 

But maybe I can look into this some time :)

danymat:
Hello, do you have an example view of this popup, and what are the advantages of using it vs the custom made by @vhyrro ?

vhyrro:
We do actually have our own selection popup already implemented if you weren't aware @fdschmidt93 :D

Neogit's popups look amazing but from what I remember they use raw text in a buffer and functions to add highlights whilst Neorg uses pure extmarks to render the text in the popups. I do agree that extracting such a selection popup into plenary would be great though. Neogit's API for constructing popups seems much nicer than Neorg's too so we may be able to mix the best of both worlds into one.

fdschmidt93:
> We do actually have our own selection popup already implemented if you weren't aware :D

Yes, I'm aware :) Sorry if this is a too ambitious suggestion :sweat_smile: I was raising this because over time upstreaming an abstraction to plenary gets ever more difficult. Again, 100% suggestion, just throw this out the window if this is out-of-scope/too ambitious.

> Neogit's popups look amazing but from what I remember they use raw text in a buffer and functions to add highlights whilst Neorg uses pure extmarks to render the text in the popups. 

Also agree extmarks are preferred :)

> Neogit's API for constructing popups seems much nicer than Neorg's too so we may be able to mix the best of both worlds into one.

I fully agree. That's why I brought this up.

[...]

PR #108, opened 2021-09-14
feat: add new UI basics

vhyrro:
This PR is here in order to completely revamp the selection popup UI aspect of Neorg into something a lot more configurable.

This very super duper configurable UI will then be utilized in parts like GTD which will be incredibly important.
Rather than using static tables Neorg will now use the Neogit-style approach of incrementally defining parts of the selection popups and then building that popup in the end. We added a bit of a twist to it though thanks to the usage of extmarks and all round a bunch more extensibility (like configurability of the renderer and individual components like switches or text).

This is still very much a WIP but it's crucial in order to push forward the development of GTD (@danymat has pretty much finished the GTD base, we now need UIs in order to progress).


vhyrro:
We're kind of stuck in an annoying predicament when it comes to the implementation. I'm considering scrapping it and starting afresh because there are just many things that don't add up properly when it comes to implementing it. I think the Neogit-OOP style isn't the best in this case if I wanna implement a generic solution for selection popups. I don't know how exactly to implement it differently though. I'd love to go for a more functional approach but idk what that'd look like.

vhyrro:
Ok this is sort of a weird question but does anyone know how to handle key presses in a "better" way? `vim.api.nvim_buf_set_keymap` only takes in a string and that's fairly tragic considering a lot of the code relies on local functions. Is there a way to get the user input in a non-blocking manner and also a way to invoke a local lua function from it?

danymat:
> Is there a way to get the user input in a non-blocking manner and also a way to invoke a local lua function from it?

I was thinking about using `nvim_buf_attach` in order to inspect the actual key presses. Wdyt?

vhyrro:
As Dany and I discussed on discord it turns out that that wasn't the way to go. In fact there was no way to go: even the not-well-known `vim.on_key` function didn't help us much when it came to parsing inputs in the selection popup. I ended up simply using `vim.api.nvim_buf_set_keymap` in the end. Good news though! The new implementation is very flexible and absolutely awesome! It'll definitely be extensible enough for use in almost any scenario and could potentially be extracted for use in other plugins.

vhyrro:
I'll be extending it more and more until I'm satisfied. Might be done in a day or two.

danymat:
Just finished moving the gtd selection popup to this new api. 
It works very well and I did some adjustments and bug fixes.

Capture d’écran 2021-09-20 à 09 04 28

I'm looking forward to such features:
- Prompt integrated with the selection popup
- Switch flag

danymat:
I created a prompt integrated with the selection popup, that is very configurable. I head that @vhyrro will refactor it a little bit, but the idea is that we reuse the same buffer in order to add a prompt line somewhere, like this code:

```lua
selection
    :title("Test")
    :blank()
    :prompt("Input", callback)
```

This is the created prompt for creating a task:

Capture d’écran 2021-09-20 à 14 37 18

TimUntersberger:
Are there any plans on moving this into a seperate plugin or into plenary so other plugins can reuse this functionality?

I have only seen a shallow overview of the PR, but so far I really like the direction your selection popups are going in.

vhyrro:
@TimUntersberger Yeah we'd love to extract this once we're done and reimplement it for a different codebase like plenary for example, shouldn't be difficult thankfully because there's not much reliance on Neorg-specific code. It'll close #98 if all goes well :)

danymat:
Okay the little mess I did above was me trying to decouple the multifile module, because it was doing a lot of problems.
I will wait for merging unstable into this PR until it is fixed

danymat:
The new UI is quite stable, so I'm merging this PR into `gtd_base` in order to avoid duplicates and dead branches.
We will open new PR's for upcoming changes on the UI.

PR #110, opened 2021-09-17
Overhaul the concealer to use TreeSitter

### LINKED ISSUE #63: Don't highlight norg syntax elements in code blocks
There's currently a bug where using modifiers like markers, headings etc gets highlighted in @code tags, not epic. Will have to use TS to detect the parent node.

vhyrro:
This PR overhauls the concealer module to use TreeSitter queries and nodes to perform concealing rather than parsing the entire document line-by-line and attempting to do regex matches. Thanks to this the code is much smaller and also more flexible. It fixes #63 and is all round much more performant. Exciting stuff!

### DIFF
--- a/lua/neorg/external/helpers.lua
+++ b/lua/neorg/external/helpers.lua
@@ -165,6 +165,16 @@ neorg.utils = {
 
         return reverse_lookup and vim.tbl_add_reverse_lookup(langs) or langs
     end,
+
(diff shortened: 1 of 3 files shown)


vhyrro:
Some other things to consider: maybe make `@code norg` blocks a thing? Since we use TS now we could also perform conceals in that. I wonder how that would work lol

PR #117, opened 2021-09-19
TS based concealing

mrossinek:
Using the workaround explained in [1] we can conceal basic markup
entirely based on the TS syntax tree. While this has some minor edge
cases (especially considering the display of a colorcolumn).

~This commit also introduces a new conceal setting called `volatile`
which will make icons.. well.. volatile! That means, that the concealing
will be lifted when the cursor is above the extmarks start position
making editing possible in a much easier fashion.~

EDIT: rather than using this approach, we decided that it makes more sense to remove the concealing extmarks upon `InsertEnter`.

[1]: https://github.com/neovim/neovim/issues/15706#issuecomment-922253522

screenshot_1636980037

:warning: This requires the `attached-modifier` branch of the TS parser!


The-personified-devil:
To note: From my testing it actually breaks fairly little, not causing issues with splits, and only with colorcolumn and maybe also with after-text-end virtual text.
So I think it should be added as an optional but disabled-by-default feature.

vhyrro:
Alright that's good to hear. When we're done with the `attached-modifiers` branch on the TreeSitter side we'll make this a disabled-by-default thing unless a feature is merged into the neovim core to allow for native TS concealing :)

mrossinek:
Thanks for the feedback @The-personified-devil! Like @vhyrro said we are currently depending on the `attached-modifiers` branch which in turn is awaiting the `new-paragraph-parsing` branch. Once the latter is merged and I have updated the former (and subsequently merged that, too), then I will come back here and update things as you suggested :+1:

mrossinek:
Alright some progress :tada: 

What is left to do:
- [ ] final regex-based concealing to be replaced with TS-based one:
  - [ ] links
  - [ ] trailing link modifier
  - [ ] ink modifier
- [ ] add concealing for sub- and superscripts (suggestions for highlight groups for the time being?)
- [ ] add examples of nested markup to docs
- [ ] (opt.) do not hard-code bold, italic, underline and strikethrough into `highlights.scm` and instead use `+TSStrong`, etc. (?)

mrossinek:
I just merged the `new-links` branch into this PR in preparation for it to be the follow-up to #178 :+1:

mrossinek:
Let's merge this into `new-links`. That way there is only one PR left which depends on https://github.com/nvim-neorg/tree-sitter-norg/pull/14

vhyrro (approved):

PR #127, opened 2021-09-24
feat: conceal ordered lists

mrossinek:
This is a proposal for concealing ordered lists. It works okayish for now but will require https://github.com/neovim/neovim/pull/9496 to be fully functional for deep nesting and to support all planned punctuation wrappers (e.g. `double_parenthesis`) and roman numerals (because these can become quite long).

<details>
<summary>Raw norg file</summary>

```
* Conceal ordered lists

  ~ First item
  ~ Second item
  ~~ Nested item
  ~~ Another nested item
  ~~~ Third level
  ~~~~ Forth level
  ~~~~~ Fifth level
  ~~~~~~ Sixth level
  ~~~ Continuation on third level
  ~ Continuation on First level
  ~~ starting a new nested level
  ~ yet another continuation
  ~~~ jumping directly to the third level
  ~~ and for some reason we now go to the second one 🤷

  ~ This is a new list because it is in a new paragraph


  Concealing can be configured to be any of:
  - numeric: 1, 2, ...
  - latin uppercase: A, B, ...
  - latin lowercase: a, b, ...

  All icon rendering can be wrapped with punctuation methods:
  - no wrapping will not add any punctuation
  - dot: adds a single dot e.g. `1.`, `A.`, `a.`
  - parenthesis: adds a single parenthesis e.g. `1)`, `A)`, `a)`
  - double_parenthesis: adds two parenthesis e.g. `(1)`, `(A)`, `(a)`
  -- Note: for this to be actually useful we will need [anticonceal](https://github.com/neovim/neovim/pull/9496)

  Potential future additions which all are likely to require anticonceal, too:
  - [ ] roman numerals for enumeration (easily requires more than one char)
  - [ ] more punctuation wrappers
```
</details>

### DIFF
--- a/lua/neorg/modules/core/norg/concealer/module.lua
+++ b/lua/neorg/modules/core/norg/concealer/module.lua
@@ -94,54 +94,6 @@ module.private = {
     namespace = vim.api.nvim_create_namespace("neorg_conceals"),
     extmarks = {},
     icons = {},
-
-    -- TODO: find appropriate location
-    -- BUG: placing this inside of module.config.public does not work due to odd require behavior
-    ordered_concealing = {
-        get_index = function(node, level)
-            local sibling = node:parent():prev_named_sibling()
-            local count = 1
-            while sibling and sibling:type() == level do
-                sibling = sibling:prev_named_sibling()
-                count = count + 1
-            end
-            return count
-        end,
-
-        icon_renderer = {
-            numeric = function(count)
-                return tostring(count)
-            end,
-
-            latin_lowercase = function(count)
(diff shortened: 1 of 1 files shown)


mrossinek:
The defaults will use no punctuation until [anticonceal](https://github.com/neovim/neovim/pull/9496) is a thing. I documented how these icon wrappers will work in the future in a comment so that we don't forget lol.

Roman numerals will also only work properly once we have anticonceal so I will leave that addition for a later point in time, too.

Other than that this should now be good to go, @vhyrro :+1:

PR #137, opened 2021-10-03
fix(modules): pass custom config to load_module

aovlllo:
I found a bug related to custom module config loading. Looks it doesn't work as expected

Another bug I've found is that neorg leader is not passed to `default_keybinds.lua`. It always loads default value (I can't say why, but looks lua saves reference to module.config.public on module load and doesn't use it's updated version). main branch doesn't have this problem because it uses `neorg_leader` as argument.

### DIFF
--- a/lua/neorg.lua
+++ b/lua/neorg.lua
@@ -82,9 +82,9 @@ function neorg.org_file_entered(manual, arguments)
     end
 
     -- After all configurations are merged proceed to actually load the modules
-    for name, _ in pairs(module_list) do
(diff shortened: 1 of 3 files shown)


vhyrro:
Hey! Thanks for making this PR, we're currently in a weird spot where modules and module loading is fairly broken lol

We'll be making even further changes to the way the module system functions so this will serve as a great temporary fix. Hopefully that other update I'm working on doesn't screw over the entire plugin :P

Issue #131, opened 2021-09-26
feature: follow link in split


mrossinek:
### Discussed in https://github.com/nvim-neorg/neorg/discussions/129

<div type='discussions-op-text'>

<sup>Originally posted by **rjahrj** September 26, 2021</sup>
After finally learning how to follow a mark I am lost on going back? Is there a keybinding for that or a preferred method. and is there a way to possibly follow a mark in a split? </div>

The suggestion was made to support following links in splits (be that horizontal or vertical). I think this can be easily implemented as an optional argument on `core.norg.esupports.goto_link` and we could then provide various keybindings to do this.

danymat:
Okay guys I don't know about the feasibility of this, but I somewhat created this feature for GTD by using buffer variables.

In GTD, I get all the tasks and their corresponding nodes in the AST. That allows me to populate the views with the tasks, and putting the nodes on their respective lines:

```
- Task 1 -- here I set a buffer variable with { line_nr = node1 }
- Task 2 -- here I set a buffer variable with { line_nr = node2 }
```

This is the code responsible for creating, updating, deleting and doing operations on them (go to location, press tab to show more details, ...):

https://github.com/nvim-neorg/neorg/blob/dab516d10a213282c8a186b6f0fba56191fcb7c4/lua/neorg/modules/core/gtd/ui/displayers.lua#L419-L583

mrossinek:
> Okay guys I don't know about the feasibility of this, but I somewhat created this feature for GTD by using buffer variables. This is somewhat OP because we can potentially move to any node in any buffer.
> 
> In GTD, I get all the tasks and their corresponding nodes in the AST. That allows me to populate the views with the tasks, and putting the nodes on their respective lines:

This is interesting. But this seems to be a rather specific usecase for when you're first building out the buffer contents (at which time you can place these buffer variables).
I would really imagine a much simpler approach (like an optional argument to `goto_link`) which specifies that you can open the target in a window split (vertical, horizontal, or an existing window id).

mrossinek:
As I recall this was part of the `new-links` branch and, thus, should now be available. @vhyrro is there an easy way to map a keybind to this functionality? If yes, I guess we can close this issue.

vhyrro:
Yup! This feature is entirely implemented now thanks to all the new goodies from new-links.
You can hit Alt + \<CR\> to open up a link location in a vertical split.

[...]

PR #159, opened 2021-10-18
ToC as virtual lines

mrossinek:
Potentially supersedes (the currently broken) #88 

Peek_2021-10-20_23-25

Known issues:
- [ ] cannot navigate onto virtual lines (see [comment here](https://github.com/neovim/neovim/pull/15351#issuecomment-946099466)) [nothing we can do about this for now]
- [x] line joining after trailing modifier does not remove said modifier and following indent
- [x] heading icons are not respected (will probably be replaced by list of links anyways)
- [x] heading seems unnecessary now that the ToC gets rendered right below the insertion

Mixing features:
- [x] support extracting ToC into location/quickfix list
- [x] do we still want a sidebar/split view? (selection popup in previous PR is pretty much broken)


mrossinek:
Alright I think it's about time for a short update...

This is based on a very early stage and experimental feature of Neovim 0.6
As such, this will definitely not be enabled by default for now.
However, the ability to also open a ToC in a split and to populate the quickfix or location list is something rather useful imo.
As such, this PR should add some examples + documentation explaining how to use this.

So here are some open todos before merging this:
- [ ] add docs/examples
- [ ] refactor to new link syntax format (depends on #160)

danymat:
We should push this feature (#131) to completion in order for the split Toc be really useful.

danymat:
I found out that calling `display_toc` multiple times does not delete the previous virtual text. And scrolling in the buffer does not delete the virtual text too

mrossinek:
> I found out that calling `display_toc` multiple times does not delete the previous virtual text. And scrolling in the buffer does not delete the virtual text too

Good catch, thanks! I will try to update this PR later this week :+1:

max397574:
This PR is quite stable, this is what's left to do:

- [-] Remove virtual texts on cursor move
- [x] Do not duplicate virtual texts if one is already in place
- [ ] Add the ability to follow the link from the TOC split (cc @vhyrro ?)

danymat:
> Second one done (what should I do when user tries to open a second one? sth like log.warn("ToC already displayed")?)

I think this is a great idea 
> 
> I'm not sure if I've understood the third one correctly
> You want that you have the toc in a split, press on a link and then jump to the heading in the normal file?

Yes, exactly (closing the toc buffer too )

mrossinek:
> Remove virtual texts on cursor move

Why would you want to do this?

>> Do not duplicate virtual texts if one is already in place
>
> what should I do when user tries to open a second one? sth like log.warn("ToC already displayed")?

I would not log a warning. Rather jump back to where the ToC is displayed and don't redo the work.

> Yes, exactly (closing the toc buffer too)

Whether or not to close the buffer needs to be configurable.

max397574:
> > > Do not duplicate virtual texts if one is already in place
> > 
> > 
> > what should I do when user tries to open a second one? sth like log.warn("ToC already displayed")?
> 
> I would not log a warning. Rather jump back to where the ToC is displayed and don't redo the work.

I have an if/else so I can do anything there
I can try to jump to the toc
> 
> > Yes, exactly (closing the toc buffer too)
> 
> Whether or not to close the buffer needs to be configurable.
ok I'll do that

danymat:
> Why would you want to do this?

Oh sorry I though it was a feature (it was like at the beginning of this PR)

max397574:
> > Why would you want to do this?
> 
> Oh sorry I though it was a feature (it was like at the beginning of this PR)

so I won't do it ig

danymat:
Okay guys short update: we can now follow links in TOC split !! 🤗 

I just found a bug by the way: #248 

I will update this a little to be able to keep the split open after jumping (with a custom config)

lua/neorg/modules/core/integrations/treesitter/module.lua
```diff
@@ -804,6 +805,15 @@ module.public = {
             _node = _node:parent()
         end
     end,
+
+    --- Get the root node from a `bufnr`
+    --- @param bufnr number
+    --- @return userdata
+    get_buf_root_node = function(bufnr)
```
mrossinek, line 812: There is already `get_document_root(bufnr)` for this purpose in this same module.

lua/neorg/modules/core/keybinds/default_keybinds.lua
```diff
@@ -60,6 +60,19 @@ module.public = {
                 silent = true,
                 noremap = true,
             })
+            keybinds.map_event_to_mode("toc-split", {
```
mrossinek, line 63: Just adding a comment here in order to comment on the keybinds file: 
I think we need a keybind for the following calls:
- `:lua neorg.modules.get_module("core.norg.qol.toc").display_toc()` -> displays ToC as non-navigable virtual lines
- `:lua neorg.modules.get_module("core.norg.qol.toc").display_toc(true)` -> displays ToC in split
- `:lua neorg.modules.get_module("core.norg.qol.toc").toqflist(false)` -> adds ToC to quickfix list (if bool is true then location list is used, no need for an extra default keybind imo)

These obviously need to be added to normal mode.

lua/neorg/modules/core/norg/esupports/hop/module.lua
```diff
@@ -32,6 +32,116 @@ module.config.public = {
 }
 
 module.public = {
+    --- Follow link from a specific node
+    --- @param node userdata
+    --- @param split string|nil if not nil, will open a new split with the split mode defined (vsplitr...)
```
mrossinek, line 37: `parsed_link` is not documented. What is it exactly?

lua/neorg/modules/core/norg/qol/toc/module.lua
```diff
@@ -0,0 +1,376 @@
+--[[
+    File: Qol-Toc
+    Title: ToC Generation
+    Summary: Generates a Table of Content from the Neorg file.
+    ---
+In order to use this feature, just add a `= TOC` at some place, and call:
+
+- `lua neorg.modules.get_module('core.norg.qol.toc').display_toc()`
+
+Calling display_toc with true (`...display_toc(true)`) will generate a split view of the Toc.
```
mrossinek, line 10: Should also document the `.toqflist()` method

mrossinek: We should also probably add a `Configuration` section

lua/neorg/modules/core/norg/qol/toc/module.lua
```diff
@@ -0,0 +1,376 @@
...
+    find_toc = function()
+        -- Extract any insertion that has a ToC value in it
+        local query = vim.treesitter.parse_query(
+            "norg",
+            [[
+            (insertion
+            (insertion_prefix)
+            item: (capitalized_word) @item
+            parameters: (paragraph_segment)? @parameters
+            (#match? @item "^[tT][oO][cC]$")
```
mrossinek, line 93:
```suggestion
            (#match? @item "^T[oO][cC]$")
```
Needs to be capitalized according to the current NFF spec (insertions are capitalized, variables are lowercase)

lua/neorg/modules/core/norg/qol/toc/module.lua
```diff
@@ -0,0 +1,376 @@
...
+        -- All captures are looped over here
+        for id, node in query:iter_captures(root, 0) do
+            -- Extract the name of the capture from the captures table (this makes it easier to perform comparisons on)
+            local capture = query.captures[id]
+
+            -- If the capture name is "item" then we set the item variable inside of node_data
+            -- It can also optionally be "parameters", in which case that variable will be set too
+            node_data[capture] = node
+
+            if node_data.line == nil then
```
mrossinek, line 121: Kind of a weird edge case: this currently prevents from two `= Toc` in the same document. Not sure if we ever wanted this, but I'd figure I should point it out.

lua/neorg/modules/core/norg/qol/toc/module.lua
```diff
@@ -0,0 +1,376 @@
...
+                    local line, _, _ = node:start()
+                    local heading_text_node = ts.get_first_node("paragraph_segment", 0, node)
+                    local heading_text = ts_utils.get_node_text(heading_text_node, 0)
+
+                    local prefix = string.rep(display_as_links and "-" or "*", heading_level)
+                        .. (display_as_links and "> " or " ")
+                    local text = prefix
+                        .. (function()
+                            if display_as_links then
+                                return "{# " .. table.concat(heading_text, "") .. "}"
```
mrossinek, line 189: Why are you using the magic link target here? What speaks against preserving the heading_level information? We're computing it anyways already and if you use the level here too, you will get the corresponding highlighting in the links, too.

lua/neorg/modules/core/queries/native/module.lua
```diff
@@ -106,7 +106,7 @@ module.public = {
     --- @param bufnr number
     --- @return table
     query_nodes_from_buf = function(tree, bufnr)
-        local root_node = module.private.get_buf_root_node(bufnr)
+        local root_node = module.required["core.integrations.treesitter"].get_buf_root_node(bufnr)
```
mrossinek, line 109:
```suggestion
        local root_node = module.required["core.integrations.treesitter"].get_document_root(bufnr)
```

mrossinek (commented):
Just some observations. Other than that this is shaping up quite nicely :+1:

lua/neorg/modules/core/norg/qol/toc/module.lua
danymat: The configuration section is automatically generated (wiki docgen) with the comments on top of each `module.config.public` parameter

lua/neorg/modules/core/norg/qol/toc/module.lua
danymat: I think this is a good idea to only allow one `= Toc` in the document. What's the point of allowing multiple ones ?

danymat:
Okay this feature is quite ready to be merged. We still need to:

- [ ] Create a command called `Neorg toc close` that will close the split, or remove the inline toc.
- [ ] Change links to headings in toc split
- [ ] Fix: if custom name in toc (`= TOC custom name`), the custom name is not showing in inline

danymat (approved):

vhyrro (approved):

PR #169, opened 2021-10-26
docs(helpfile): improve vim help file

max397574:
- [x] norg code block for nested stuff: so you can see the unconcealed syntax
- [ ] new links


vhyrro:
Aight I think it should be good now! Thanks a lot man :D

PR #177, opened 2021-11-02
Fix workspace matching

dnguyener:
This is similar to 76a621f00b8f631879dd7b75d63c788b40171764.  Get_workspace_match is unable to match 
workspace directories containing special characters.  In my case, its hyphens.


danymat:
Merged, thanks !

PR #184, opened 2021-11-05
Better indentation

mrossinek:
This is also a WIP feature worked on by @vhyrro. Again, I think the improved visibility would be good :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 # Neorg - An Organized Future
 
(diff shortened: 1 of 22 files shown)


danymat:
@mrossinek Is this already pushed in main because of attached-modifiers ?

vhyrro:
@danymat nah this one's still in the works, no merged code here :)

danymat:
Continuation here: #291

PR #190, opened 2021-11-10
update helpfiles to new link format

max397574:


mrossinek (approved):
Thanks @max397574! :+1:

PR #193, opened 2021-11-13
Regex fallback for code blocks

katawful:
~~The bulk of the code is added, but it has severe issues that I'm not sure how to address. For one, the default handling of the 'NeorgTagContent' can't be used at all or we won't see the regex fallback. The other is that only one regex syntax type code block can work at a time. You can't mix syntax within files. The good news is that this doesn't require TS to have additional vim regex enabled~~

~~I am not sure where to go with this PR atm, it seems like there's a real disconnect with how tree-sitter and regex buffers work with each other. The code I adapted was mostly from Vimwiki, which has an extremely robust implementation of nested syntax~~

Multiple languages have been added

Example of regex highlighting fallback:
Screenshot from 2021-11-14 16-42-03

Known issues:

- [x] Only one Vim syntax file is allowed to be loaded at a time - fixed by 0f455fb
- [x] Some languages can break free from code block - fixed by 92af83c
- [ ] Indent dependent languages care about code block indent level - potential upstream issue
- [x] Sometimes won't load colors - fixed by ba4c89b
- [x] Multiple syntax groups load upon highlight refresh - fixed by acd3d3c
- [x] When Neorg updates in any capacity, regex syntax highlighting fails - fixed by 2dc38db

Code Tasks

- [x] Remove possible VimL code for Lua and nvim API calls instead (for speed) - fixed by 8437850 & e0546d3
  - The syntax stuff will need to be changed whenever that gets added to the api


vhyrro:
Awesome! I'll play around with this tomorrow, see if and how I could help. I feel like there must be a way to make this work for mixed languages too.

btw using a `::continue::` label is fine, sometimes there's just no other way of doing it :P

katawful:
So with a bit more searching I figured out *what* was breaking with this feature. The syntax include never actually happens for the second language, and I'm not sure why this is. If this can be fixed then everything else should work just fine

katawful:
Ok, figured out how to do multiple languages. There's only 2 more main problems to address: language escaping from blocks and languages counting code block indents as their own indents

katawful:
Fixed the last major issue and did some code clean up. The only thing left is indentation for indent aware languages (make comes to mind). I don't think I could fix this here, it seems like a general issue with how code blocks work with this plugin. Languages that do break don't leak out still, they just won't highlight properly because they think they're indented if you indent the code block. See below:

I don't see much else I need to do here, I'll leave it to one of the maintainers to go from here. As of right now the code seems pretty stable and fast from my testing, I'll be using it to remake my Vimwiki stuff some more and see how it works, but unless someone else tries this and finds an issue there shouldn't be any problem with the code bug-wise

lua/neorg/modules/core/integrations/treesitter/module.lua
```diff
@@ -32,7 +32,8 @@ module.config.public = {
             },
 
             Parameter = "+TSType",
-            Content = "+Normal",
+            -- this is seemingly needed to fix regex fallback from not showing
```
mrossinek, line 35: I'm not sure where these highlights get applied but this could be related to whether we allow highlight groups to combine. If not, this probably overwrites the regex syntax highlight. Maybe there is a way to fix this somewhere else in our code.

lua/neorg/modules/core/norg/concealer/module.lua
```diff
@@ -304,6 +304,95 @@ module.public = {
...
+            local code_lang = vim.treesitter.parse_query(
+                "norg",
+                [[(
+                (ranged_tag (tag_name) @_tagname (tag_parameters) @language)
+                )]]
+            )
+
+            -- look for language name in code blocks
+            -- this will not finish if a treesitter parser exists for the current language found
+			for id, node in code_lang:iter_captures(tree:root(), 0, from or 0, -1) do
```
mrossinek, line 317: This entire block seems to be indented one level too deep.

lua/neorg/modules/core/norg/concealer/module.lua
```diff
@@ -304,6 +304,95 @@ module.public = {
...
+					-- It's a very good implementation of nested vim regex
+					regex_language = regex_language:gsub("%s+", "") -- need to trim out whitespace
+					local group = "textGroup" .. string.upper(regex_language)
+					local snip = "textSnip"..string.upper(regex_language)
+					local start_marker = "@code "..regex_language
+					local end_marker = "@end"
+
+					-- if our region syntax group exists, quit out
+					-- this stops repeat groups from forming on text update
+					-- NOTE: potenial nvim api call can be used here
```
mrossinek, line 345: Maybe you can check out `nvim_get_hl_by_name`. Other than that nothing comes to mind that would fit here.

lua/neorg/modules/core/norg/concealer/module.lua
```diff
@@ -304,6 +304,95 @@ module.public = {
...
+					    current_syntax = vim.b.current_syntax
+						vim.b.current_syntax = nil
+					end
+
+					-- temporarily pass off keywords in case they get messed up
+					local is_keyword = vim.api.nvim_buf_get_option(0, "iskeyword")
+
+					-- see if the syntax files even exist before we try to call them
+					-- NOTE: this is what fails for the second language
+					-- TODO: replace with non-vimL functions
```
mrossinek, line 363: Are you looking for `nvim_get_runtime_file` here?

lua/neorg/modules/core/norg/concealer/module.lua
danymat: This should not be a problem with automatic stylus formatting

katawful: Yeah, I tried to get it proper myself but it's hard to tell sometimes. I'll take another check when I go through changes again

lua/neorg/modules/core/norg/concealer/module.lua
katawful: This seems a touch broken from my testing, it doesn't actually tell you if a group doesn't exist. It'll always create a new one, unless I'm missing something

lua/neorg/modules/core/norg/concealer/module.lua
katawful: Bah, totally missed that. The API needs to be easier to search sometimes

lua/neorg/modules/core/integrations/treesitter/module.lua
katawful: Yeah I'm not sure what it does exactly. It doesn't seem to break other tag blocks in my testing but I don't exactly feel comfortable removing the feature, it had to have a purpose I feel as is

vhyrro:
Shame that it currently doesn't work on my end :sweat_smile:. I'll try debugging the code a little, it seems to have something to do with when the highlights are applied - they don't seem to be applied at all when the program begins. Modifying some text doesn't apply the highlights as expected either, it creates `textGroupMA`, `textGroupMAK` groups but never a `textGroupMAKE` hl group for whatever reason :(

katawful:
> Shame that it currently doesn't work on my end sweat_smile. I'll try debugging the code a little, it seems to have something to do with when the highlights are applied - they don't seem to be applied at all when the program begins. Modifying some text doesn't apply the highlights as expected either, it creates `textGroupMA`, `textGroupMAK` groups but never a `textGroupMAKE` hl group for whatever reason :(

I noticed something similar but couldn't figure out what it was. It happened when Neorg updated its windows, like when you go through gtd views to view projects, then quit out of that window or when you go between files. What's happening probably is that the syntax still "exists" so my code never recreates the nested syntax. Some sort of check will have to be added. It should work, however, if you just open up a raw Neorg file and do regex code blocks in just that file

vhyrro:
The thing is that it doesn't work at all lol. I managed to get it to work once on accident cause i kept changing random bits of text until it started working, but by default it fails 🤔

katawful:
I can only it to break if I traverse through the jumplist between Neorg files or if open the file manually, though its only with syntax that previously existed. It always seems to load whenever I open the file

katawful:
Ok, I think I've fixed it from not loading properly. It should hopefully load on first go, but it won't also fail when there's a second appearance of the same language or when you reload the file

mrossinek:
Okay I was finally able to get an example to work. Some observations:
- luckily both norg-TS parser branches (`main` and `attached-modifiers` work)
- currently you added your code within the code block dimming method. I had that option disabled which was one reason why I never got this to work before. I suggest you extract this into a separate method instead.
- indentation appears to be a real issue for some languages. I was naively testing with `make` but this appears to be a particularly bad choice since its syntax does not allow any whitespace preceding e.g. its targets

I'm going to take a closer look at this on the weekend but I'm glad to see it working now. Maybe you can extract your code into a separate function in the meantime?

katawful:
Yeah, indentation is going to be a big problem. Where do you suggest putting this? Still a bit unfamiliar with the project

vhyrro:
If i have the time today I'll implement it :)

katawful:
I moved the fallback code to its own function within the concealer module, and added the appropriate calls where needed. Please let me know if this is a good place for it

katawful:
So I was thinking about the problem with how the `TagContent` highlight group seems to overwrite the regex highlighting. Is it possible that this can be disabled for just code blocks that have a syntax file attached to it somehow?

vhyrro:
I suppose we could ditch some highlights like TagContent and other things like that, they don't really add much value. As for the indentation issues I tried quite a few things but none of them seemed to work :(. Does stuff like makefile codeblocks in vimwiki work perfectly fine, or do they have some complex chains of vim highlights to work properly? I don't think it'd be worth making such complex highlights, especially since TS parsers are fairly easy to make and there'll be more and more of them as we go.

katawful:
Vimwiki has the same issue with it's preprocessor (and I assume similar markdown plugins), however the language itself doesn't have much indentation requirements so most users probably wouldn't come across it. That's why I was wondering if it was something that could be fixed within NeoVim itself at point

There's not a lot of indent aware languages, so I think it's best to just make a note somewhere for end users to be aware of

vhyrro:
Yeah I totally agree on this one. It's probably not worth fighting with those indent-dependent langs that hard :P

By the way, do you think this should be disabled-by-default or enabled-by-default? I don't know exactly how much it affects performance and also how many people would use this. I feel we should make it a toggleable flag though in `module.config.public`. Other than that we're ready to merge :D

katawful:
I just pushed a commit that should avoid doing any of the syntax updating if a syntax file is already loaded (since its handled internally by NeoVim, not my code). This should remove most of the performance penalty

I can try throwing a bunch of regex into a file and see what its like

katawful:
[test.txt](https://github.com/nvim-neorg/neorg/files/7582198/test.txt) - if you want to test

I added a bunch of regex (a total of about 2400, so quite a lot) to this file and messed around with my code changes and noticed a few things:   
- With large files, Neorg in general is quite slow. I tested this with and without the regex code function being called within "BufEnter" autocmd (like 10-20ms difference over ~300ms, mostly just from not loading the syntax files)
- Actually editing text isn't too bad, about as good as any large regex highlighted file. Other current unstable Neorg functions aren't much slower, at least from what I can tell

Overall, I don't think having an option for disabling regex highlighting will do anything. Any slowness will be more of a problem with how much Neorg does with its autocommands in really large files

vhyrro:
Awesome, thanks a bunch for trying out those benchmarks. Neorg does get slow with the concealer enabled, yeah :/. It's not really our fault here - there's no way for us to properly apply extmarks without having to reparse the entire buffer on every change.

`vim.api.nvim_buf_attach` and its `on_lines` function doesn't work properly when undoing text, so that's not a viable option. The way nvim-orgmode does it is via a decoration provider. This works very nicely for what they're doing, but when you have to reparse treesitter queries on every redraw things get... tragic, to say the least. I guess our only bet is to either wait for on_lines to start working properly or wait for `vim.*` to be available in libuv threads so we could do this stuff asynchronously.

Until then though we'll live with what we got! Thanks a tonne for working on this. You got anything else on your mind, anything you wanna quickly implement beforehand, or we good to merge? Gimme the green light and the PR will be on its way :)

katawful:
I can't think of anything else that's needed aside from dealing with `TagContent` highlighting

vhyrro:
TagContent got removed in https://github.com/nvim-neorg/neorg/commit/1dac9f6b9bd1084ec9ede118dd2548dbf2fd3dfb, so I don't think it should pose much of a problem anymore, right?

katawful:
Ah yeah, it shouldn't then. I commented it out originally. All is good then!

vhyrro:
Sick! Thanks again :purple_heart:

PR #200, opened 2021-11-21
docs: update README regarding lazyload

ray-x:
To allow lazy load work, might need to 
1) add filetype `neorg` (if init.lua set syntex&filetype detection to be off when launch neovim, or, in any case nvim failed to detect filetype for `neorg`)
2) manually load `treesitter` before setup `neorg`

I think this would solve most problems related to lazy loading. But please comment on if it still failed to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,10 +165,29 @@ You can install through any plugin manager (it can even be vimscript plugin mana
   Here's an example:
 
   ```lua
-  use { "nvim-neorg/neorg", ft = "norg", config = ... }
+  use {
+    "nvim-neorg/neorg", 
+    -- in case you turn off filetype detection when startup neovim
+    setup = vim.cmd("autocmd BufRead,BufNewFile *.norg setlocal filetype=norg"),
+    ft = "norg",
+    config = function()
+      local loader = require"packer".loader
+      -- lazy load treesitter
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -165,10 +165,29 @@ You can install through any plugin manager (it can even be vimscript plugin mana
...
+    "nvim-neorg/neorg", 
+    -- in case you turn off filetype detection when startup neovim, or ft detection failure for norg
+    setup = vim.cmd("autocmd BufRead,BufNewFile *.norg setlocal filetype=norg"),
+    ft = "norg",
+    config = function()
+      local loader = require"packer".loader
+      -- lazy load treesitter
+      if not packer_plugins['nvim-treesitter'].loaded then
+        loader("nvim-treesitter")
+      end
```
vhyrro, line 178: Couldn't this be achieved by making sure neorg is `after = "nvim-treesitter"`? Or does this have some side effect that makes it actually work?

ray-x: You are right. it should work in most cases. I updated the PR.
For me. I do not trust packer that much because it may fail when using `after` `before` in a complicated dependencies.

vhyrro:
Yup, looks nice! Thanks :D

PR #233, opened 2021-12-12
feat: add callback for completions

esquires:
I'd like to enable completion for project tags for this [extension](https://github.com/esquires/neorg-gtd-project-tags) where the set of completions updates changes depending on changes that are made after load time (e.g., a user adds another project/subproject tag).  This small PR enables that to happen. You can see its use in the `add-completion` branch of the extension [here](https://github.com/esquires/neorg-gtd-project-tags/blob/add-completion/lua/neorg/modules/utilities/gtd-project-tags/module.lua#L34)

### DIFF
--- a/lua/neorg/modules/core/norg/completion/module.lua
+++ b/lua/neorg/modules/core/norg/completion/module.lua
@@ -249,7 +249,9 @@ module.public = {
         -- Loop through every completion
         for _, completion_data in ipairs(completions) do
             -- Construct a variable that will be returned on a successful match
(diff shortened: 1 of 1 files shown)


vhyrro:
Looks nice, and I can see it having some cool uses. Thanks!

danymat:
Wow this is such a good idea @esquires !!

What about adding this functionality in core GTD ? So that the user can quickly add `contexts` or `waiting.for` ?

esquires:
Thanks! I am a bit preoccupied over the next couple of weeks but can take a look after that.

PR #234, opened 2021-12-12
Generate gtd completions

danymat:
This is based off the #233 that uses callback for completions. Even if it does not uses this, I though it was a good idea to generate completion for GTD.

I will be merging this as soon as we simplify unstable and main workflow


esquires:
Would it be better to generate this dynamically? For instance if a user creates a new context with this implementation it wouldn’t show up in the completion engine until they reopen vim. 

Also can we add completion for the next few days for when the task starts/stops?

danymat:
It could be better indeed, but I have no clue how to handle this to trigger when a new context is created. And I don't want to trigger this on CursorLeave...


About time.start and time.due, it should be pretty easy to do, with a custom text (e.g tomorrow, 2 days, 3 days, 1week...)

esquires:
> It could be better indeed, but I have no clue how to handle this to trigger when a new context is created. And I don't want to trigger this on CursorLeave...

Good question. No autocommands are needed. You would change this [line](https://github.com/nvim-neorg/neorg/blob/unstable/lua/neorg/modules/core/norg/completion/module.lua#L181)

```lua
complete = {"today", "someday"}
```

to something like

```lua
complete = function(context, prev, saved) module.private.get_contexts() return 
```

It looks like you already have an implementation of `get_contexts` in this pull request so it would just be a matter of wrapping it in a function and calling it in the above. 

I only overwrote the completions in [gtd-project-tags](https://github.com/esquires/neorg-gtd-project-tags/blob/master/lua/neorg/modules/utilities/gtd-project-tags/module.lua#L34) because that seemed to be the only way to overwrite the config (the `completions` table [here](https://github.com/nvim-neorg/neorg/blob/unstable/lua/neorg/modules/core/norg/completion/module.lua#L56) is in `module.public`, not `module.config.public`).

danymat:
>  looks like you already have an implementation of get_contexts in this pull request so it would just be a matter of wrapping it in a function and calling it in the above.

Hmm I'm not that found of this implementation to be called each time we go to completions, because I parse every task and project in order to find contexts.. Maybe I will try to only get relevant values from them, which are `contexts` and `waiting.for` atm.

esquires:
If you are looking for a cached solution then it seems like refreshing when the task is edited via “gte” (adding a check for new contexts) would be fast. It would just involve adding any new contexts to an existing cached table. Alternatively when the projects are displayed the tasks have to be calculated anyway so that is another option

danymat:
> If you are looking for a cached solution then it seems like refreshing when the task is edited via “gte” (adding a check for new contexts) would be fast. It would just involve adding any new contexts to an existing cached table. Alternatively when the projects are displayed the tasks have to be calculated anyway so that is another option

I'm not sure we have to focus on gte for refreshing this, because completion will only be called when we change contexts by hand.

esquires:
Great points! It sounds like dynamic loading may not be a good feature to add at this point then.

danymat:
I did some tests and I saw that It increases Neorg stating time by a lot. 
I disabled it by default, to enable:

```lua
["core.gtd.base"] = {
	config = {
                ...
		custom_tag_completion = false,
                ...
	},
},
```

vhyrro:
Perhaps y'all should look into running the tag completion in a `vim.schedule` at startup if that's even doable in your case. It would at least delay the generation of the completion candidates and make it feel snappier.

danymat:
It does not seem to improve by wrapping with `vim.schedule`..

PR #247, opened 2021-12-15
finish (hopefully) toc

max397574:


max397574:
What this does:
- fixes error with multiple virtual ToCs
- adds a keybinding

What it doesn't:
- links from split not found **atm**

What I want it to do:
- add an easy option for opening (perhaps tab and shift tab for split and virt text)
  - add config for whether to open in split or a virtual text
- add more configurations
  - see above
  - whether to close split after entering link or not

PR #250, opened 2021-12-15
docs: update readme to mention regex fallback for codeblocks

katawful:
Didn't see this changed yet, felt it was important to mention


vhyrro:
You're right, it was important to mention. Thanks for this!

PR #252, opened 2021-12-17
update gtd links to new format

max397574:
should I use
`{# Clarify}[Clarify]`
or only `{# Clarify}`?


danymat:
You can keep the links like that ! Thanks

PR #82, opened 2021-08-27
feat(snippets): support for Neorg snippets

JoeyGrajciar:
Basic snippet support for now `code` and `image` blocks.

- [ ] vsnip
- [x] luasnip
- [ ] moar snippets
- [ ] sane triggers for snippets

Showcase:
luasnip_neorg

PR #255, opened 2021-12-17
Snippet support

JoeyGrajciar:
Reopen of #82

### DIFF
--- a/lua/neorg/external/helpers.lua
+++ b/lua/neorg/external/helpers.lua
@@ -405,6 +405,14 @@ neorg.lib = {
     lazy_string_concat = function(...)
         return table.concat({ ... })
     end,
+
+    get_package_path = function()
+        -- Path to this source file, removing the leading '@'
+        local source = string.sub(debug.getinfo(1, "S").source, 2)
+
+        -- Path to the package root
+        return vim.fn.fnamemodify(source, ":p:h:h:h:h")
+    end,
 }
 
 return neorg.utils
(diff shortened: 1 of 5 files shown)


JoeyGrajciar:
@danymat Wondering if it makes sense to continue with this PR as friendly snippets have `norg` snippets present https://github.com/nvim-neorg/neorg/pull/82#issuecomment-985985916
Other issue is that `core.norg.completion` module interferes with snippets if trigger starts with `@`.

mrossinek:
Okay, just for to understand, is the purpose of this PR to add snippets for common norg format structures?

I'm not sure what the common approach is here. Since snippet engines can differ in their syntax, it seems like quite some overhead to supply snippets for norg for possibly quite many snippet engines. Is that common practice or would rather snippet engines supply those in their source? (I'm not big on snippets so I honestly do not know :sweat_smile:)

danymat:
> I'm not sure what the common approach is here. Since snippet engines can differ in their syntax, it seems like quite some overhead to supply snippets for norg for possibly quite many snippet engines. Is that common practice or would rather snippet engines supply those in their source? (I'm not big on snippets so I honestly do not know 😅)

I think that's the good idea, it is like the implementation of completion sources

danymat:
> @danymat Wondering if it makes sense to continue with this PR as friendly snippets have norg snippets present #82 (comment)
Other issue is that core.norg.completion module interferes with snippets if trigger starts with @.

I think that's a good idea to add support for more snippets engine. I personally use luasnip and I don't know if we can supply norg snippets directly within luasnip repo

danymat:
I don't think we can merge this until we fix the `@` issue

vhyrro:
Hey! Unfortunately we'll be closing this PR. We don't see much value in providing snippets ourselves - this is something that snippet collections like friendly-snippets should do, that way not everything is centralized in Neorg itself.

Hope you understand!

molleweide:
I published 'LuaSnip-snippets.nvim' a couple days ago where I'll try to add neorg snippets writtwn in lua.

PR #257, opened 2021-12-18
clean up helpfiles

max397574:


mrossinek:
Are you cleaning up all help files or is this meant for only this one?

max397574:
Originally I planned only this but there are others who still use the old links so I'll do those as well

mrossinek:
> Originally I planned only this but there are others who still use the old links so I'll do those as well

Cool! While you're at it, could you please also check the contents of the `tags` file? I remember a discussion somewhere someone pointing out that `neorg-foreplay` doesn't work. The target no longer exists but it is still present in the `tags` file.

max397574:
foreplay is no longer a tag

danymat:
Merged, thanks !

PR #258, opened 2021-12-18
Improved Concealing

mrossinek:
Improves the TS-based concealing behavior

### DIFF
--- a/doc/neorg.norg
+++ b/doc/neorg.norg
@@ -350,7 +350,7 @@ consective new lines.
       - You can create ,/italic subscripts/,.
       - If you want to shout in the internet, use */_DOUBLY EMPHASIZED AND UNDERLINED CAPS_/*
 
-      Note: You can:*not*: combine sub- and superscripts like so:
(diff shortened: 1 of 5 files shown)


mrossinek:
There are still performance issues related to concealing (especially upon undoing) but at the time there is nothing we can do about that.
Besides that, this PR does the following:
- enabled whitespace-based concealing as the default
- adds the following configuration options:
    - `config.markup.enabled = true|false` (replaces `config.markup.enable`)
    - `config.markup_preset = "safe"|"brave"` (`safe` is the default (whitespace), `brave` uses the unicode word joiner)
    - migrates all `config.icons.markup` to `config.markup`
- adds the `:lua neorg.modules.get_module("core.norg.concealer").toggle_markup()` function to toggle markup at runtime
- adds the `neorg_leader .. "mt"` default keybinding to the above function
- heavily refactors all markup concealing in general (basically entirely relies on fast TS queries now instead of slower custom render functions)

vhyrro (approved):

PR #259, opened 2021-12-18
core.queries.native: Read file content before parsing nodes

### LINKED ISSUE #225: GTD changes pwd
> Yeah I tried it out. This is actually a different question. `NeorgStart silent=true` do start Neorg. Said you can follow these steps:

> 1. Open Nvim at `~/foo`
> 2. Run `pwd` in command mode => result should be `~/foo`
> 3. Start Neorg via `NeorgStart silent=true`
> 4. Open GTD view by `Neorg gtd views`
> 5. Check `pwd` again in command mode => expected to be `~/foo` but now changed to GTD workspace instead

_Originally posted by @vanhtuan0409 in https://github.com/nvim-neorg/neorg/discussions/224#discussioncomment-1771687_

danymat:
(Used in GTD) This should prevent the system from opening a buffer for each parsed query in core.queries module.

It's opening the file and reading it instead, and using the text to parse.

This should fix #225 

(related: https://neovim.discourse.group/t/get-parser-function-does-open-the-buffer/1382/2?u=danymat, thanks @kristijanhusak)


danymat:
Okay so I got a fairly functional thing here. 

It currently reads the norg file with filesystem method (without opening a buffer), and uses the resulted string to parse the AST with tree sitter. 

However, I don't think the implementation is optimal. I currently use tables as data to keep the files contents in string. But this is not optimal because for GTD modifications (adding or deleting contexts for example), I have to save the file every single time there's a change. 

I though of using a temporary unlisted buffer (so that it does not change any directory). In that way, I could place the file content in there, and parse the buffer afterwards. This would allow me to do changes and saving the resulted modifications in the original file at the end.

I will then close this PR and recreate it with a new one (I don't want to create a mess here).

PR #267, opened 2021-12-20
Better Concealing Performance

vhyrro:
This PR attempts to make the concealer a tonne faster by using incremental updates instead of entire buffer reparses.

There's a reason why we didn't start by using incremental updates to begin with - `nvim_buf_attach`s `on_lines` function did not properly work with undos due to implementation reasons. We initially hoped it'd get fixed but after finding out it won't be patched it seems that we simply needed to take matters into our own hands.

This implementation will use `on_lines` to determine the range in which changes to a document have occurred.
It'll then strictly use that range for reparsing - much better than reparsing the entire buffer.

As for undos though, I don't really have much of an idea yet - i thought of a custom `u` and `U` keybind or an `on_key` handler.
As for handling undone ranges I'm thinking of playing around with `undotree()`, I wonder whether there's anything there that we can use.

### Stuff that we gotta do
- [x] Apply conceals everywhere on `BufEnter`
- [x] Unbind nvim_buf_attach calls when user exits buffer
- [x] Make undo work properly
    - [x] Make `u` and `U` simply reparse the entire document for now, look into `undotree()` later.
- [x] Fix currently broken concealing elements, which is basically everything apart from icons rn

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -71,11 +71,11 @@ body:
         - "Yes, but I don't know how to start. I would need guidance (check question below)"
         - "No"
     validations:
-      required: false
+      required: true
 
   - type: textarea
     attributes:
       label: "Implementation help"
       description: "If you selected yes in the last question please specify what you would need help with in order to implement this?"
-    validations:
-      required: true
+
+
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,12 @@
+name: Linter
+
+on: [push, pull_request_target]
+
+jobs:
+  luacheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: nebularg/actions-luacheck@v1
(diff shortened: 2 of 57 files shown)


vhyrro:
Yo, just a quick update from me - after a lot of work and trial and error I've managed to make the concealer a lot faster on startup now too! Before it was simply performance improvements while editing, but now you can have your file open basically without speed compromises. The solution? Chunks!

Chunks are simply segments of your document with a fixed size (by default 1250 lines). If your file is larger than one chunk then instead of loading the entire file at the same time it incrementally loads the buffer's conceals chunk by chunk at a specified interval. You may be thinking: "man, 1250 lines is still a lot to parse in one go", and you would be right, but we've also "improved" the performance of applying large amounts of conceals in one go by deferring each individual conceal through `vim.schedule` - it works surprisingly well.

What you see below is a video of a 5 thousand line file full of `* Hello World!` statements (that's quite a lot of extmarks) and how well the concealer manages to operate with them:

https://user-images.githubusercontent.com/76052559/147505334-d656e8b3-d4b1-4f46-bde3-f46c18407846.mp4

PR #294, opened 2022-01-03
fix check for no data

esquires:
get_by_var can return either {} or nil. Prior to this pull request the if statement could pass when `data == {}`


danymat:
Good catch, thanks !

PR #296, opened 2022-01-04
move ret_completions after check for match

esquires:
Neorg had slowed down a lot for me recently. This fixed the slowdown I observed (it was calling [this function](https://github.com/esquires/neorg-gtd-project-tags/blob/master/lua/neorg/modules/utilities/gtd-project-tags/module.lua#L41) in [gtd-project-tags](https://github.com/esquires/neorg-gtd-project-tags) every time text changed).

Prior to this commit, when completion_data.complete was a function it would be called every time text was changed. This calculation is unnecessary though when the regex does not match.

Related to #233


danymat:
Legit to do this calculation only when there's a match. Thanks!

PR #308, opened 2022-01-12
Better Keybinds

vhyrro:
This PR is here to make Neorg's keybind core less unintuitive, more flexible and powerful.

TODO:
- [x] Make default leader key `<LocalLeader>`
- [x] Enable default keybinds by default when loading `core.keybinds`
- [x] Stop default keybind code from overwriting user-defined keybinds
- [x] Make a new config option called `hook` which will be directly passed into `neorg.callbacks`
- [x] Implement new functions
	- [x] `keybinds.unmap()`
	- [x] `keybinds.remap()`
	- [x] `keybinds.mapd()`
	- [x] `keybinds.map_event()`
    and other functions once I figure out the full API structure.
- [ ] Keybind presets
	- [ ] `vim` keybind preset, for vim veterans who prefer native vim keybindings over mnemonics
	- [x] `neorg` keybind preset, for our hand picked keybinds that we deemed the most intuitive and memorable
	- [ ] `org` keybind preset, for those transitioning from org-mode
- [x] Add support for keybind requests, aka allow other modules to provide default keybinds if the user wants them.

This PR's desc will be updated as I get new ideas


danymat:
> Make a new config option called hook which will be directly passed into neorg.callbacks

Can you ship this one separately into main or an other PR ? It'll be useful for GTD as well

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -1,10 +1,36 @@
...
+### Setting Up a Keybind Hook
+Want to change some keybinds? You can set up a function that will allow you to tweak
+every keybind bit by bit.
+
+```lua
+["core.keybinds"] = {
+    config = {
+        hook = function(keybinds)
+            keybinds.unmap("norg", "n", "gtd")
+            keybinds.remap_event("norg", "n", "<C-Space>", "core.norg.qol.todo_items.todo.task_done")
```
mrossinek, line 27: In terms of docs: I assume `unmap` and `remap_event` are explained in detail elsewhere?
Even if so, it may be nice to have a comment on the expected arguments right here.

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -100,16 +126,28 @@ module.load = function()
...
+
     -- Prefix for some Neorg keybinds
-    neorg_leader = "<Leader>o",
+    neorg_leader = "<LocalLeader>",
+
+    -- Function to be invoked that allows the user to change their keybinds
+    hook = nil,
+
+    -- The keybind preset to use
+    keybind_preset = "neorg",
```
mrossinek, line 147: So how does the `neorg` preset relate to the `default_keybinds.lua` file? I assume they are identical but the names are mismatched.

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -189,26 +230,58 @@ module.public = {
...
+                    opts = opts,
+                }
+            end,
+
+            map_event = function(neorg_mode, mode, key, expr, opts)
+                payload.map(neorg_mode, mode, key, "<cmd>Neorg keybind " .. neorg_mode .. " " .. expr .. "<CR>", opts)
+            end,
+
+            unmap = function(neorg_mode, mode, key)
+                if neorg_mode == "all" then
```
mrossinek, line 249: Since normally modes are written in lowercase, maybe we can change the spelling of this one special? E.g. `ALL` or `__all__` (Pythonic I know...) but just some suggestions.

mrossinek: Also needs to be documented

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -189,26 +230,58 @@ module.public = {
...
+                payload.map(
+                    neorg_mode,
+                    mode,
+                    key,
+                    "<cmd>Neorg keybind " .. neorg_mode .. " " .. new_event .. "<CR>",
+                    opts
+                )
+            end,
+
+            remap_key = function(neorg_mode, mode, old_key, new_key)
```
mrossinek, line 279: What exactly is the difference to `remap`?
I'm asking cause I saw `map` and `map_event`. But there is `remap`, `remap_event` and this one (`remap_key`)...

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -270,57 +344,81 @@ module.public = {
 
             -- Include the current Neorg mode in the contents
             mode = current_mode,
+            leader = module.config.public.neorg_leader,
```
mrossinek, line 347: Why do we need to duplicate this into the payload? Can't we read this directly from the config when needed elsewhere?

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -270,57 +344,81 @@ module.public = {
...
-                                log.warn(
-                                    "Warning in keybind definition for",
-                                    keybind,
-                                    "- 'mode' key has no effect when paired with the 'action' key."
-                                )
+        neorg.events.broadcast_event(
+            neorg.events.create(module, "core.keybinds.events.enable_keybinds", payload),
+            function()
+                for neorg_mode, neovim_modes in pairs(bound_keys) do
+                    if neorg_mode == "all" or neorg_mode == current_mode then
```
mrossinek, line 382: Another `all` mode case

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -270,57 +344,81 @@ module.public = {
...
+                        for mode, keys in pairs(neovim_modes) do
+                            for key, data in pairs(keys) do
+                                local ok, error = pcall(function()
+                                    if action then
+                                        action(buf, mode, key, data.command, data.opts or {})
+                                    else
+                                        if type(data.command) == "string" then
+                                            vim.api.nvim_buf_set_keymap(buf, mode, key, data.command, data.opts or {})
```
mrossinek, line 390: This makes us depend on Neovim 0.7.X, right?

danymat: I think that latest keycap setting in 0.7.x is `vim.keymap.set` instead

lua/neorg/modules/core/keybinds/module.lua
vhyrro: This is usually just for convenience for users. In the user hook they would have to execute `neorg.modules.get_module_config("core.keybinds").neorg_leader` in order to access it, versus here they can directly index it :P

PR #314, opened 2022-01-15
Unprocessed tasks logic

danymat:
Hello ! 🥳

This is part of the discussion I had with @molleweide in #307.

The idea is summarized below:

> Because we can improve the views a little bit by: Adding a new view called Unprocessed.

It'll group every task or project that is not entirely processed. 
For a task to be entirely processed, it has to be visible. 
- If it's a task: by having one or more contexts 
- If it's a project: by having one or more tasks associated.
In this way, we can remove the inbox file because we could view our unprocessed tasks, and we will not lose any tasks from the views.
In fact, I'm sure it'll be a better feature, because the inbox file is quite difficult to grasp for new users, and still not very usable by advanced users (because moving tasks and projects is not yet implemented in GTD actions).

I will track here every thing I still need to do in order to mark this feature as completed:

~- [x] Remove any allusion of `INBOX`~
- [x] Create two views: `unclarified tasks` and `unclarified projects`
- [x] Remove unclarified tasks and projects from other views

EDIT:

After a very enlightening discord conversation, we made the conclusion to still keep the inbox file, and the `unclarified` tasks and projects should be the mix of inbox PLUS previously unclarified tasks.

### DIFF
--- a/doc/gtd.norg
+++ b/doc/gtd.norg
@@ -232,10 +232,11 @@
 
    **** Changes from base GTD
     ***** Inbox list
-          There is not concept of a inbox list. However, we have what we call `Unclarified` tasks
-          and projects. 
-          - For a task to be processed, it should require at minima 1 context.
-          - For a project to be processed, it should require at minima 1 task.
+          The inbox list is a set of all tasks and projects that are:
+          - In the inbox list (default: `inbox.norg` file)
+          - /Unclarified/, and not in inbox list
+          -- For a task to be clarified, it should require at minima 1 context.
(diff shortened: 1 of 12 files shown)


danymat:
Hello, it seems that this is ready to be merged. 
You can take a look at the `:h neorg-gtd` spec (last modifications listed on top of the file).

2 two views in GTD views (`:Neorg gtd views`):
- Unprocessed tasks
- Unprocessed projects

Feel free to give a feedback in the GH discussion, or post any issue in the Issues page ! 🥳 🎈

PR #328, opened 2022-01-24
feat(gtd): allow count for weekdays

max397574:
this allows to have a time.due/start for example for the 4th monday that will come.
The format is 4mon.

### DIFF
--- a/lua/neorg/modules/core/gtd/queries/helpers.lua
+++ b/lua/neorg/modules/core/gtd/queries/helpers.lua
@@ -52,7 +52,7 @@ module.public = {
         if amount and weekday then
             local date = os.date("*t")
             if values[weekday] > date.wday then
-                date.day = date.day + values[weekday] - date.wday
+                date.day = date.day + values[weekday] - date.wday + (amount - 1) * 7
             else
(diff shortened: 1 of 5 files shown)


danymat:
Nice, It seems that the tests passes. Would you mind adding one or two tests for the new case ?

max397574:
I'll see what I can do

lua/neorg/modules/core/gtd/queries/helpers.lua
```diff
@@ -44,6 +44,21 @@ module.public = {
...
         end
+        local amount, weekday = text:match("^(%d+)(%a%a%a)$")
+        if not vim.tbl_contains(vim.tbl_keys(values), weekday) then
+            weekday = nil
+        end
+
+        if amount and weekday then
+            local date = os.date("*t")
+            if values[weekday] > date.wday then
+                date.day = date.day + values[weekday] - date.wday
```
max397574, line 55: I think I need to put a `+(amount-1)*7` here right?

danymat (approved):

danymat:
merged, thanks !! 🥳

PR #329, opened 2022-01-24
Better Neorg News

vhyrro:
This PR makes the news system for Neorg much more robust by making the implementation more specific and by focusing on user experience more than extensibility.

To use it, you should learn to use the default commands:
- `:Neorg news all` - displays *all* news in one big changelog. Cool if you just want to have a read about Neorg's history and how it's changed over time.
- `:Neorg news old` - displays all news that you've already seen in the past in one big file. To view news for a specific old Neorg version run `:Neorg news old <version>`.
- `:Neorg news new` - displays all the news you *haven't* seen yet in one file. Useful if you simply want to see what's changed after a plugin update. You can also use a version number here as a parameter if you want to see news for a single version (`:Neorg news new <version>`).


lua/neorg/modules/core/norg/news/module.lua
```diff
@@ -18,316 +26,251 @@ module.setup = function()
...
+                if is_version_greater_than(version, module.private.latest_version) then
+                    module.private.latest_version = version
+                end
 
-    local christmas = module.required["core.storage"].retrieve("christmas")
+                module.private.new_news[version] = filepath
+            else
+                module.private.old_news[version] = filepath
+            end
+        end
```
danymat, line 71: Can't you use `neorg.lib.match` here ?

danymat (commented):
what happens if the user rolls back to a previous version ? will he have the Neorg news for this version?

lua/neorg/modules/core/norg/news/module.lua
vhyrro: `match` seems a little overkill here, plain old `if` checks do the trick :)

vhyrro:
> what happens if the user rolls back to a previous version ? will he have the Neorg news for this version?

No, if the user is on e.g. `0.0.9` and rolls back to `0.0.8` then the internal neorg.mpack file will still have `0.0.9` as the latest version, so the user won't see any news for `0.0.9`.

danymat:
So he will see news for 0.0.8 ?

danymat:
Then I guess we can merge!

PR #337, opened 2022-01-25
[ON HOLD] Prep 0.0.11

mrossinek:
Before releasing Neorg 0.0.11 we should merge these breaking changes.


vhyrro:
There have been so many changes to the treesitter parser and to Neorg that it's not worth maintaining this PR anymore. After the dev branch of the parser is considered "complete" we will *then* start work on upstreaming those changes into Neorg :p

PR #357, opened 2022-02-02
refactor(metagen): update metadata if it already exists

ram02z:
When you run `:Neorg inject-metadata`, it will create another copy of the metadata instead of replacing it. Not sure if this was intended.


vhyrro:
Hey, thanks a bunch for this! Didn't think of doing such a simple thing :)

Issue #365, opened 2022-02-09
Regex syntax code blocks overwrite syntax settings for norg filetype


katawful:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.6.1

### Neorg setup

n/a

### Actual behavior

During the process of loading in syntax files for code blocks that don't use tree-sitter, the entire filetype plugin that NeoVim will use (whether built in or an external plugin) will also load in extra settings such as `foldmethod`, `syntax case`, `s:cpo_save`, etc... These will overwrite whatever syntax settings exist for the norg filetype

### Expected behavior

Regex syntax code blocks do not leak out settings and data

### Steps to reproduce

Add any regex syntax code block to a norg file that has a syntax related setting/variable and see how it changes from what Neorg needs

### Help

Yes

katawful:
Might be worth making a bigger issue of regex code block issues, a lot have been popping up since I've been using Neorg and I'm not exactly happy with the implementation

Issue #315, opened 2022-01-15
Regex highlighting for code blocks can fail in some situations if the syntax file only does 'hi link'


katawful:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.6.1

### Neorg setup

      {:load {:core.defaults {}
              :core.norg.concealer {:config {:markup_preset :safe
                                             }}
              :core.keybinds {:config {:neorg_leader :<Leader>n}}
              :core.norg.dirman {:config {:workspaces {:main "~/Documents/neorg"}
                                          :autodetect true
                                          :autochdir true}}
              :core.gtd.base {}
              :core.norg.qol.toc {}
              :core.norg.journal {:config {:workspace :main
                                           :journal_folder "journal"
                                           :use_folders false
                                           }}
              :core.norg.completion {:config {:engine :nvim-cmp}}
              }

### Actual behavior

With syntax files that use the `hi link groupA groupB` method instead of `hi def link groupA groupB` method, whenever the regex highlight needs to update (like changing color schemes or modifying the code block), the code block will cease to highlight.

### Expected behavior

This doesn't happen

### Steps to reproduce

NeoVim's built in Scala syntax file uses `hi link` format. Simply add an example block like below and change color scheme and the highlighting should go away:

```scala
object Hello {
    def main(args: Array[String]) = {
        println("Hello, world")
    }
}
```

### Help

Yes

### Implementation help

I'm mostly making this issue to remind myself since I'm the one that implemented this, but if anyone else wants to take a stab at it go ahead

vhyrro:
Hm yeah this seems to be a problem. I couldn't find any real way to do it either without breakages. I believe we'll just have to live with this and hope that treesitter starts prevailing slowly but surely.

I wish I had the time but I'd love for someone to just go on a language adding spree and add as many languages as they can to nvim-treesitter so we no longer have to worry about regex so much :P

PR #367, opened 2022-02-10
Various regex code block highlighting improvements

katawful:
Felt displeased with my implementation of regex code blocks, this PR addresses the majority of them

Issues:

- [ ] hi link groups clear when Neorg updates the code block syntax groups - NeoVim issue #315 
- [x] general performance issues, scheduling is perhaps wrong way to go about this
- [x] syntax settings from code blocks leak out #365 
- [x] regex languages aren't available for completion like treesitter languages
- [x] syntax files can "double" which leads to syntax markers to escape - ToC issue?
- [x] create new module for regex code block stuff


katawful:
Performance fixes are almost done, but ultimately even a single well made syntax file will increase CPU time on the main thread by about double when a redraw call is made in my testing. It is a bit better for redraws, and significantly more reliable and less bloated for neorg events however. I have 1 more major performance fix I want to introduce and then it's cleanup time for the most part. This PR might hit 0.1 window but no guarantee

katawful:
Performance fixes are done

katawful:
I can't seem to replicate my issues with other syntax files that do the same thing, so my code is probably ok? Seems to be a bug in how Vim in general can accept some ways to do things, so its a bug in my own syntax files. Not going to mark that issue done until its 100% reliable however

katawful:
Leaking issue seems mostly fixed now. Going to do some more writing in Neorg just to confirm the functionality, but I think I have done everything I possibly can to stop it from ever happening while also allowing all broken language features to still work

katawful:
Everything in my list is done, think its read for review

danymat:
I want to test this, but I don't know what I'm looking out. Can you pinpoint some basic steps for testing, and what I should be aware of ?

katawful:
Ideally there shouldn't be anything out of the ordinary, I've tested this a lot. My suggestion is to get some various regular syntax languages you know of, preferably ones that might contain more features like syntax folding and the like, and see how they work while plainly using Neorg. Like using Telescope or FZF to switch files, using GTD regularly, etc... I haven't done much of that recently

katawful:
Hmm, regular Vim jumps doesn't cause the doubling issue but Neorg hop jumps do (like ToC)

katawful:
Using fzf-lua with full buffer previews seems broken. Carries the preview buffers over. I presume that telescope does the same. Will have to check to fix this

katawful:
Ok, I think I've finally fixed it all for real this time. Barring the conflicts in the concealer module which needs to be fixed, I haven't found any immediate issues

katawful:
Vim regex is not good enough to do everything we need for Neorg, and with no sign of that changing, some use will have to be broken

vhyrro:
:purple_heart:

PR #376, opened 2022-02-20
Improve presenter module

max397574:


lua/neorg/modules/core/presenter/module.lua
```diff
@@ -56,17 +56,70 @@ module.load = function()
...
+    },
 }
 
 module.private = {
     data = {},
     nodes = {},
     buf = nil,
     current_page = 1,
+    ns = nil,
+    display_slide_count = function()
```
danymat, line 89: Please add documentation for this function

lua/neorg/modules/core/presenter/module.lua
```diff
@@ -56,17 +56,70 @@ module.load = function()
...
+        if module.config.public.slide_count.enable then
```
danymat, line 90: Maybe you should do:
```lua
if not module.config.public.slide_count.enable then
  log.warn(...)
  return
end
```

To minimize indentation

lua/neorg/modules/core/presenter/module.lua
```diff
@@ -56,17 +56,70 @@ module.load = function()
...
+            local text = string.format(
+                module.config.public.slide_count.count_format,
+                module.private.current_page,
+                #module.private.data
+            )
+            if module.config.public.slide_count.position == "top" then
```
danymat, line 96: You can use `neorg.lib.match` instead doing if/elseifs

lua/neorg/modules/core/presenter/module.lua
```diff
@@ -185,6 +242,7 @@ module.public = {
         vim.api.nvim_buf_set_option(module.private.buf, "modifiable", true)
         vim.api.nvim_buf_set_lines(module.private.buf, 0, -1, false, module.private.data[module.private.current_page])
         vim.api.nvim_buf_set_option(module.private.buf, "modifiable", false)
+        module.private.display_slide_count()
     end,
 
     close = function()
```
danymat, line 248: Yo should delete the namespace on close

danymat (changes requested):

danymat:
Please add a description of the improvements in your PR

danymat:
Some comments: 

if slide count is at `bottom`, it is showing in the same line of the last line. Causing text to be hidden.

Capture d’écran 2022-02-20 à 20 48 05

It would be better to show on a line just after it.

I propose to stick the the slide_count on top of bottom of the window, and not dependent of the text.
At the moment, if we have a long text, scrolling through the slide will make the slide_count scroll. I would prefer it to be sticked.

Current behavior on a long text: (with `bottom`)

Capture d’écran 2022-02-20 à 20 49 59

PR #385, opened 2022-02-27
Fix/journal file creation

max397574:


max397574:
there are some commits here which shouldn't be here

Issue #384, opened 2022-02-27
:Neorg journal <anything> deletes whatever content was in the journal file


jakobkhansen:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.7.0-dev+1145-gc26ef6bcd4

### Neorg setup

```
require("neorg").setup({
	load = {
		["core.defaults"] = {}, -- Load all the default modules
		["core.norg.dirman"] = { -- Manage your directories with Neorg
			config = {
				workspaces = {
					gtd = "~/Documents/gtd",
				},
			},
		},
		["core.norg.completion"] = {
			config = {
				engine = "nvim-cmp", -- We current support nvim-compe and nvim-cmp only
			},
		},
		["core.norg.concealer"] = {
		},
		["core.keybinds"] = { -- Configure core.keybinds
			config = {
				default_keybinds = true, -- Generate the default keybinds
				neorg_leader = "<Leader>o", -- This is the default if unspecified
			},
		},
		["core.integrations.telescope"] = {},
		["core.gtd.ui"] = {},
		["core.gtd.queries"] = {},
		["core.gtd.base"] = {
			config = {
				workspace = "gtd",
				default_lists = {
					inbox = "inbox.norg",
				},
				exclude = {
					"journal",
				},
			},
		},
		["core.norg.journal"] = {
			config = {
				workspace = "gtd",
			},
		},
	},
})
```

### Actual behavior

Hi, when using the command `:Neorg journal <anything>` the command overwrites whatever was already in the journal file. For example if I do `:Neorg journal today`, write something, then reopen that file with the same command, everything is gone.

### Expected behavior

Expected behavior is to not overwrite whatever was in the file, just open it with the contents.

### Steps to reproduce

`:Neorg journal today`

write something in file.

`:Neorg journal today`

file content is gone.

The buffer could even still be open and the file contents will be emptied.

### Help

No

max397574:
this is because journal always creates a new file
I'm working on a fix for this at the moment.

vhyrro:
Hey! This should hopefully be fixed now. Lemme know if all's good :)

jakobkhansen:
Great, works now! Thanks guys

PR #386, opened 2022-02-27
fix(journal): existing files would get overwritten

max397574:
checks if file exists and opens it if it does
This will close https://github.com/nvim-neorg/neorg/issues/384
Also use correct separator for paths


vhyrro (approved):

vhyrro:
Thx!

PR #389, opened 2022-02-27
Create Export Module

vhyrro:
Y'all better get excited for this PR :P

As we approach 0.1 (we're actually getting close this time I promise) the last thing we need to do is create an export tool for Neorg. The reason isn't as obvious as it may seem on the surface though - it's not because we want to allow exporting to markdown or org (even though this PR will allow that!), but the core reason why this PR exists is to allow for *upgrading of Neorg documents*. 

As our syntax changes rapidly before our stable 0.1 release we want to have a way for users to be able to seamlessly transition between norg syntax revisions without having to manually rewrite their documents. This export module will allow for such a thing to exist, and as such that is why it is crucial for it to be implemented before our first stable Neorg release, so hang tight!


roland-rollo:
My body is ready. Much love for all you, holy warriors! <3

vhyrro:
Oops, gonna have to scratch the current implementation and go with a new one. It was a good idea but the execution was meh and didn't end up scaling well.

vhyrro:
Good news, we're back baby. Better implementation is here, I'd consider the basics of the exporter core complete. The only thing left? Utilizing that very core and creating a bunch of functions that allow norg export to markdown. This will be a test of how well the exporter works, and will also make people excited that they can convert neorg into md :P

vhyrro:
Alright everyone, I'm fairly certain that this is ready to rock and roll! It exports to markdown fine in my testing, if you find any bugs or issues then be sure to file an issue, cause there are definitely things I may have missed or overlooked.

In the meantime, it's time to merge! All documentation and code commenting will be done in one swoop just before the release of 0.1!

vhyrro:
Shit, how do I keep doing this lol

I forgot the most crucial part, writing to the file. I'll deal with that now.

vhyrro:
There we go, *now* I'd consider the export module basis to be complete :D

PR #399, opened 2022-03-12
Adding `core.tutor` module

shift-d:
Basically this module is somekind alternative of `:Tutor`, but only for neorg format.
This lets user discover useful features via interactive guide besides specs.


mrossinek:
Lol, sorry, that was unintended :joy:

vhyrro:
Closing this due to inactivity. It's definitely something we want to have one day!

PR #402, opened 2022-03-19
Treesitter Based Concealing

vhyrro:
This PR is a temporary workaround for treesitter based concealing not working (see https://github.com/nvim-neorg/neorg/issues/292#issuecomment-1073023644). The "brave" preset will now actually physically conceal the text thanks to a discovery I made recently with `matchaddpos()` :P

It may be a little less performant with this update, but it's a tradeoff I'm willing to make until we get the real thing.


vhyrro:
Fantastic news! The real deal has been merged! We can finally have normal concealing! Note that this requires the absolute latest Neovim master/nightly to function :D

PR #407, opened 2022-03-27
Indentation v3

vhyrro:
Third time's the charm... right?

Indentation is stupid difficult like you wouldn't believe - it seems so simple on the surface but there are so many edge cases. That ends with this PR. I've learnt a lot about the most effective techniques to indent Neorg documents, and will try applying all that amassed knowledge here.

Here's a small preview in the meantime:
indents


vhyrro:
One of my main problems right now is actually not the indentation itself, but `'indentkeys'`. Yup, that's right, `'indentkeys'`. They're very limited, and I need to either roll my own lua implementation or try to contribute to Neovim core 😅. Only time will tell what I actually end up doing.

vhyrro:
Alright! We're done here. Is the indentation engine perfect? No. Is it much better than the previous implementation? Heck yes!

If you find bugs with it then be sure to make an issue! As for now I'd say this is ready as an MVP.
As always extensive documentation will be provided just before the release of 0.1, which is nearing close now.

PR #410, opened 2022-03-28
add default text for capture popup

max397574:
contexts and waiting for now has current values as default text so you can edit them


danymat (approved):

danymat:
Merged, thanks !

PR #414, opened 2022-03-29
[GTD] Rewrite capture core

danymat:
Hello everyone, 

Right now GTD seems very stable, with the exception of GTD capture methods.
Why though ?

-  First of all, I was still a newbie when I started out creating this integration. 😁
- But most importantly, capture methods for Neorg gtd feels kinda complicated, let me show you:

Imagine you want to create a task inside a project, and this is the type of task that you want to create:

```lua
- [ ] Wash dishes
``` 

If you want to add to a very simple project, like this one:

```
* Home things
```

It's possible to add it like this:

```
* Home things 
- [ ] Wash dishes 
```

But now imagine you already had a task before, with a context:

```
* Home things 
  #contexts sister
  - [ ] ask her to wash dishes 
```

You can't simply put your task afterwards, because Neorg (and therefore gtd) treats things inside tags as groups, so this:

```
* Home things 
  #contexts sister
  - [ ] ask her to wash dishes 
  - [ ] Wash dishes   
```

Will mean that our task will have `sister` context.

OKAY THEN, what will be done here ? Basically, follow this workflow:

1 . Ask for every information for a task
2. Adds the task to the buffer
3. After the task has been added, verify the location of new task: does it require some operations ?
4. modify the buffer accordingly, and add more information (like new contexts for example)

### DIFF
--- a/lua/neorg/modules/core/gtd/queries/creators.lua
+++ b/lua/neorg/modules/core/gtd/queries/creators.lua
@@ -6,50 +6,55 @@ module.public = {
     --- supported `string`: project|task
     --- @param type string
     --- @param node core.gtd.queries.task|core.gtd.queries.project
-    --- @param bufnr number
     --- @param location table
-    --- @param delimit boolean #Add delimiter before the task/project if true
     --- @param opts table|nil opts
-    ---   - opts.new_line(boolean)   if false, do not add a newline before the content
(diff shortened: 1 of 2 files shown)


vhyrro:
I'm closing this PR because GTD will be rewritten soon anyway and there's no point in having this sit here all day :smile:

PR #422, opened 2022-04-05
added folding for headings

max397574:


vhyrro:
Hi max, the reason I don't think I want this is that headings will be folded by default unless the foldlevel Neovim variables are tinkered with - I wish you could set a priority for folds (perhaps with `(#set! priority x)` but until that's the case I want to be very conservative with what's folded and what isn't. Am tempted to implement a custom fold function like kristijan's nvim-orgmode does it for better results, but I also don't want to give myself more things to do before 0.1 - it'll just delay the release even further :)

max397574:
👍 so I'll close this

zdcthomas:
It seems like this got implemented recently, and I was wondering if there was a way to turn off the folding by default. I'm probably just missing it somewhere, but I'd ideally like to have the concealer turned on but with no (or limited) folding.

PR #424, opened 2022-04-09
fix: errors in fs_write on some systems when exporting a file

katawful:
My Pop OS machine would fail to run the export function, complaining that arg 3 of vim.loop.fs_write is nil instead of a number. Using 0 instead seems to remedy this. My Arch machine works fine


vhyrro:
Interesting, thanks for the fix!

PR #426, opened 2022-04-12
fix(gtd): nil query

esquires:
Prior to this pull request, `query_nodes_from_buf` could return `{}` or `nil`. When it was the latter, this block in `retrievers.lua` in the `get` function would fail:

```lua
        for _, bufnr in pairs(bufnrs) do
            ---@type core.queries.native
            local queries = module.required["core.queries.native"]
            local nodes = queries.query_nodes_from_buf(tree, bufnr)
            vim.list_extend(res, nodes)  -- fails here because nodes is nil
        end
```

The backtrace is 

```
Error executing vim.schedule lua callback: vim/shared.lua:0: src: expected table, got nil
stack traceback:
        [C]: in function 'error'
        vim/shared.lua: in function 'validate'
        vim/shared.lua: in function 'list_extend'
        .../neorg/lua/neorg/modules/core/gtd/queries/retrievers.lua:104: in function 'get'
```

### DIFF
--- a/lua/neorg/modules/core/norg/esupports/indent/module.lua
+++ b/lua/neorg/modules/core/norg/esupports/indent/module.lua
@@ -18,11 +18,7 @@ end
 module.public = {
     indentexpr = function(buf)
         local node = module.required["core.integrations.treesitter"].get_first_node_on_line(buf, vim.v.lnum - 1)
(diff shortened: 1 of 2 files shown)


vhyrro:
Heyo, thanks for the fix! :)

PR #429, opened 2022-04-15
Bump Neovim version to latest stable (0.7)

roland-rollo:
As project use things from 0.7 version, more than 0.6, users have more benefits knowing that newer version of Neovim is better for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Neorg - An Organized Future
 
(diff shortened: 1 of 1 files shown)


max397574:
Not sure about this
Since it doesn't require 0.7 to work

vhyrro:
Heyo, thanks for the PR! Neorg doesn't strictly require Neovim 0.7 just yet, *but* we'll be rewriting several core components to use 0.7 APIs soon - in the meantime this PR will remain open. After the migration we'll then merge this :)

vhyrro:
Well here we go. Neorg now officially requires Neovim 0.7 to operate, merge time!

Issue #419, opened 2022-04-03
Popup help window with relevant combinations


premell:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

I think its very helpful to have some common bindings right in neorg, so you dont have to go to github to check the syntax. I use neorg about once a month and I always find myself coming here to see how to do basic stuff, like filling in a todo list.

I think it would be nice to have a quick look up right in neorg with relevant bindings. I am thinking something like this. ![image](https://user-images.githubusercontent.com/65544203/161428989-799072d5-23a3-4ca0-bb44-85981d372278.png)

I also find nvchads keybinding list very helpful and it could be some inspiration: 
![image](https://user-images.githubusercontent.com/65544203/161429065-f8e8d67c-81fd-4980-9bf6-51ad435dac4e.png)

You could also use treesitter to figure out what binding the user wants to look up. For instance here you can see that user is hovering a todo list, and so you can show the various options for todos (ie [x] for done).
![image](https://user-images.githubusercontent.com/65544203/161429145-7d80fc67-0293-4128-a6ca-1a0438248a70.png)

In this example the user is hovering a bold text. You could create categories with relevant combinations, like "important text". So if they hover bold text, you could give them a list of the different headers etc.
![image](https://user-images.githubusercontent.com/65544203/161429168-c12659a4-eb7d-4262-a9b7-7998101347c1.png)

### Help

No

max397574:
I'll add a cheatsheet module for this

vhyrro:
While this used to be a great idea back in the day, nowadays this should ideally be delegated to some sort of `which-key` implementation (there are a few out there right now), such a feature would just bloat modern neorg :p

Closing!

PR #431, opened 2022-04-18
add cheatsheet

max397574:
closes https://github.com/nvim-neorg/neorg/issues/419


max397574:
for now I just copied that example file
We should change that though

max397574:
@premell is this what you though about?

max397574:
@vhyrro 
the tests that fail have nothing to do with this pr right?

vhyrro:
Tests failing are "normal" :p

If i get some time i'll populate the contents of the cheatsheet to have something more enticing than just a copy of the example file

mrossinek:
I don't think it's a good idea to effectively duplicate the examples file in this format. We already have too many files which need to be kept consistent and with the formal syntax specification moving to the TS repo, the documentation over here will also need to be overhauled. Just my two cents..

max397574:
Yes
for now I just added the example file (added links) because it was mainly about the floating window things etc
Eventually we should write a compact cheatsheet

max397574:
@vhyrro ?

vhyrro:
I've just spent a bit of time compiling a larger cheatsheet that is just a simple helpfile for the time being (`:h neorg-cheatsheet`). In my opinion we should close this PR and we can work from the ground up on a solution that revolves around the new cheatsheet file with features like "show automatically when it's the user's first time installing neorg", a tutorial mode (if you've seen conjure's tutorial then i'd love something along those lines) and so on.

This PR not being merged became kind of a meme at this point but it can't just be here forever :smile:

Issue #432, opened 2022-04-20
Task properties break progress indicator (x of Y, percent)


metalelf0:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.8.0-dev+12-g147cc60d2

### Neorg setup

```lua
require("neorg").setup({
	-- Tell Neorg what modules to load
	load = {
		["core.defaults"] = {}, -- Load all the default modules
		["core.norg.concealer"] = {}, -- Allows for use of icons
     ["core.integrations.telescope"] = {}, -- Enable the telescope module
		["core.norg.dirman"] = { -- Manage your directories with Neorg
			config = {
				workspaces = {
					default  = "~/neorg/notes",
          gtd      = "~/neorg/gtd",
				},
        autochdir = false,
        index = "index.norg",
        last_workspace = vim.fn.stdpath("cache") .. "/neorg_last_workspace.txt" -- The location to write and read the workspace cache file

			},
		},
		["core.norg.completion"] = {
			config = {
				engine = "nvim-cmp",
			},
		},
		["core.gtd.base"] = {
			config = { -- Note that this table is optional and doesn't need to be provided
        workspace = 'gtd'
			},
		},
		["core.keybinds"] = { -- Configure core.keybinds
			config = {
				default_keybinds = true, -- Generate the default keybinds
				neorg_leader = "<Leader>o", -- This is the default if unspecified
			},
		},
    ["core.export"] = {},
    ["core.export.markdown"] = { config = { extensions = "all" } },
	},
})
```

### Actual behavior

<img width="549" alt="Schermata_2022-04-20_alle_08 46 41" src="https://user-images.githubusercontent.com/131911/164168862-9b6a422e-94f7-4015-8e21-5663b3e82605.png">

The "GDR" project has two tasks, but only the first one is counted towards progress. It seems task properties (`#time.due`, `#time.start` and `#contexts`) break parsing the task list.

Moving the second task before the first one fixes the issue, but if I add properties to both tasks, there's no way to count both.

### Expected behavior

In the example above, the virtual text for the task should read `(0 of 2) [0% complete]`.

### Steps to reproduce

Try pasting the following content:

```
* GDR
  - [ ] Rivedere lista incantesimi Petpetor
[...]

PR #433, opened 2022-04-20
fix: concealer breakage on carryover tag set when calculating todo items

metalelf0:
This PR actually fixes issue #432 by using recursion to iterate over all descendant nodes and counting all todo_items.

### DIFF
--- a/lua/neorg/modules/core/norg/concealer/module.lua
+++ b/lua/neorg/modules/core/norg/concealer/module.lua
@@ -473,41 +473,39 @@ module.public = {
             local total = 0
 
             for child_node in start_node:iter_children() do
-              if child_node:type() == "generic_list" then
-                  for todo_item_node in child_node:iter_children() do
-                      if vim.startswith(todo_item_node:type(), "todo_item") then
-                          local type_node = todo_item_node:named_child(1)
+                if child_node:type() == "generic_list" then
+                    for todo_item_node in child_node:iter_children() do
+                        if vim.startswith(todo_item_node:type(), "todo_item") then
+                            local type_node = todo_item_node:named_child(1)
 
-                          if type_node then
-                              local todo_item_type = type_node:type():sub(string.len("todo_item_") + 1)
-                              local resulting_todo_item = results[todo_item_type] or 0
+                            if type_node then
+                                local todo_item_type = type_node:type():sub(string.len("todo_item_") + 1)
+                                local resulting_todo_item = results[todo_item_type] or 0
 
-                              results[todo_item_type] = resulting_todo_item + 1
-                              total = total + (todo_item_type == "cancelled" and 0 or 1)
-                          end
-                      end
-                  end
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/norg/concealer/module.lua
```diff
@@ -66,6 +66,27 @@ local function schedule(func)
...
+            local todo_item_type = type_node:type():sub(string.len("todo_item_") + 1)
+            local resulting_todo_item = results[todo_item_type] or 0
+
+            results[todo_item_type] = resulting_todo_item + 1
+            results.total = results.total + (todo_item_type == "cancelled" and 0 or 1)
+        end
+    end
+end
+
+local function traverse_nodes(root_node, results, func)
```
max397574, line 83: why the `func` parameter
this function is only used here and always the same function is used

lua/neorg/modules/core/norg/concealer/module.lua
```diff
@@ -66,6 +66,27 @@ local function schedule(func)
     end)
 end
 
+local function add_to_counters_if_todo_node(node, results)
+    if vim.startswith(node:type(), "todo_item") then
+        local type_node = node:named_child(1)
+
+        if type_node then
+            local todo_item_type = type_node:type():sub(string.len("todo_item_") + 1)
```
max397574, line 74: seems really weird to do that like this
just use the number or  (cleaner imo) a pattern

lua/neorg/modules/core/norg/concealer/module.lua
metalelf0: I passed `func` as a parameter to make it more extendable. You can pass a custom function and have the code count items depending on different rules, or do different stuff on nodes. I get your point though, so if you prefer to keep it simple I can remove the func parameter and make it use the `add_to_counters_if_todo_node` function directly.

lua/neorg/modules/core/norg/concealer/module.lua
metalelf0: I just copy-pasted the original implementation. Not sure why it was done this way, maybe @vhyrro can tell us?

lua/neorg/modules/core/norg/concealer/module.lua
max397574: I think you could
we can later change this if we want
but I see no reason why we'd do this

lua/neorg/modules/core/norg/concealer/module.lua
vhyrro: Magic numbers in my opinion are way worse than less readable code - using a pattern is overkill to me (why an entire regex system to cut a string?). I'm pretty annoyed that `len()`'s return value is incompatible with `:sub()` and requires a `+ 1`, but it's lua what can ya do :(

vhyrro:
Looks good to me! Code is nice, can't seem to find any bugs with it - thanks a bunch for the PR!

PR #434, opened 2022-04-21
feat: replace beta print in `:Neorg module list` with a UI

max397574:


max397574:
Screenshot 2022-04-21 at 18 35 36

at the moment this looks like this
The styling of the window is easy to change.
Highlights can also be added
So feel free to suggest things

danymat:
You should make the UI non editable

max397574:
Screenshot 2022-04-21 at 20 21 13

looks like this now

PR #440, opened 2022-04-30
feat(gtd/queries): allow sorting by state

max397574:


max397574:
tests aren't failing because of this pr I'm 99.9% sure

vhyrro:
No they're not caused by the PR i don't think. Thanks!

Issue #442, opened 2022-04-30
Attempt to call nil field `tree_map_rec` when trying to use TOC features


zdcthomas:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.7.0

### Neorg setup

```lua 
require("neorg").setup({
	load = {
		["core.defaults"] = {},
		["core.norg.qol.toc"] = {},
		["core.norg.concealer"] = {},
		["core.norg.esupports.metagen"] = {
			config = {
				template = {
					{
						"title",
						function()
							return vim.fn.expand("%:p:t:r")
						end,
					},
					{ "description", "" },
					{ "authors", "Zachary Thomas" },
					{ "categories", "" },
					{
						"created",
						function()
							return os.date("%Y-%m-%d")
						end,
					},
					{ "version", require("neorg.config").version },
				},
			},
		},
		["core.integrations.nvim-cmp"] = {},
		["core.norg.journal"] = {
			config = {
				strategy = "flat",
			},
		},
		["core.norg.completion"] = {
			config = {
				engine = "nvim-cmp",
			},
		},
		["core.norg.dirman"] = {
			config = {
				workspaces = {
					wiki = "~/irulan",
				},
			},
		},
	},
})
```

### Actual behavior

In a `.norg` file which contains an `= TOC`, when I run `Neorg toc inlilne/split/toqflist` I get an error which doesn't appear in the `norg.log` file in `stdpath('data'). The error text is 
```
Error executing vim.schedule lua callback: ...art/neorg/lua/neorg/modules/core/norg/qol/toc/module.lua:253: attempt to call field
 'tree_map_rec' (a nil value)
stack traceback:
        ...art/neorg/lua/neorg/modules/core/norg/qol/toc/module.lua:253: in function 'generate_toc'
        ...art/neorg/lua/neorg/modules/core/norg/qol/toc/module.lua:281: in function <...art/neorg/lua/neorg/modules/core/norg/qo
l/toc/module.lua:267>
        ...art/neorg/lua/neorg/modules/core/norg/qol/toc/module.lua:390: in function 'on_event'
[...]

PR #443, opened 2022-05-01
fix(treesitter): readd deleted functions 

max397574:
(fixes https://github.com/nvim-neorg/neorg/issues/442)


vhyrro:
Thanks! I'll probably make the toc module use a different function either way but this is good for now lol

PR #458, opened 2022-05-15
feat(neorgcmd): select next unspecified command argument

AckslD:
First of, I realised that I should maybe have first reach out and checked if this makes sense with the current vision. But you can see this more as a draft/suggestion and if you don't like it no problem :)

I started trying out `neorg` a few days ago and already loving it. When starting out one thing that bothered me a bit was when I was trying to run a command, let's say `:Neorg news` just to see what it does, I get an error with something like:
```
Unable to execute neorg command under name event_name - minimum argument count not satisfied. The command requires at least
```
of course I can then try again using `<Tab>` so see what the options are. But to reduce that little bit of friction I thought it would be great if instead of giving the error, `neorg` would instead query the user for the next argument using `vim.ui.select` (which can for example be call out to eg `telescope` based on the users config). This is what this MR attempts to achieve.

See the following gif for how the current MR can work:

neorgcmd

Feel free to let me know what you think and if this makes sense :)

Thanks for the awesome work :tada:

### DIFF
--- a/lua/neorg/modules/core/neorgcmd/module.lua
+++ b/lua/neorg/modules/core/neorgcmd/module.lua
@@ -137,21 +137,25 @@ end
 ---@param args table #previous arguments of the command Neorg
 ---@param choices table #all possible choices for the next argument
 local _select_next_cmd_arg = function(args, choices)
-    local current = string.format('Neorg %s', table.concat(args, ' '))
+    local current = string.format("Neorg %s", table.concat(args, " "))
 
     local query
 
     if vim.tbl_isempty(choices) then
-        query = function(...) vim.ui.input(...) end
+        query = function(...)
+            vim.ui.input(...)
+        end
     else
-        query = function(...) vim.ui.select(choices, ...) end
+        query = function(...)
+            vim.ui.select(choices, ...)
+        end
     end
 
     query({
         prompt = current,
     }, function(choice)
         if choice ~= nil then
-            vim.cmd(string.format('%s %s', current, choice))
+            vim.cmd(string.format("%s %s", current, choice))
         end
     end)
 end
@@ -329,7 +333,7 @@ module.public = {
             -- performs a recursion on the data of each keybind versus the completions of each keybind. This means
             -- that keybinds with many arguments wouldn't be registered and would instead cause this function to
             -- loop. The only solution is to query completions for the next item here.
-            local completions = _neorgcmd_generate_completions(_, string.format("Neorg %s ", table.concat(args, " "))
+            local completions = _neorgcmd_generate_completions(_, string.format("Neorg %s ", table.concat(args, " ")))
             _select_next_cmd_arg(args, completions)
             return
         end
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/neorgcmd/module.lua
```diff
@@ -132,10 +132,26 @@ function _neorgcmd_generate_completions(_, command)
...
+    end
+    vim.ui.select(choices, {
+        prompt = current,
+    }, function(choice)
+      vim.cmd(string.format('%s %s', current, choice))
+    end)
+end
+
 module.load = function()
     -- Define the :Neorg command with autocompletion and a requirement of at least one argument (-nargs=+)
```
max397574, line 152: update this

lua/neorg/modules/core/neorgcmd/module.lua
```diff
@@ -297,14 +313,8 @@ module.public = {
         end
 
         -- If our recursion depth is smaller than the minimum argument count then that means we have not supplied enough arguments
```
max397574, line 315: update this

vhyrro:
Hey! This is interesting actually, I'll take a look at this when i get home :)

AckslD:
One could also add a custom eg `BACK` choice which takes you back one level, so if you're at the choices for arguments to `Neorg gtd` it will take you back to the choices for `Neorg`.

lua/neorg/modules/core/neorgcmd/module.lua
```diff
@@ -297,14 +314,9 @@ module.public = {
...
```
max397574, line 316: shouldn't this be
> the user has not supplied enough arguments

vhyrro:
From my testing it works nicely apart from prompts - consider `:Neorg export to-file my-file.md`. It's impossible to enter `my-file.md` as the code always defaults to a `select` call - perhaps we need some extra logic here to check for those - also, how are optional parameters handled, can they even be handled at all? If we can resolve these issues this would be an awesome addition :p

AckslD:
@vhyrro I made it so that for prompts without a finite set of choices for the argument it uses `vim.ui.input` instead of `vim.ui.select`. Let me know if that works as you'd expect.

Re: optional parameters, do you have an example for that? How do you think that should work? Maybe it could be fine to say that if a user wants to supply optional parameters they need provide the full command directly?

vhyrro:
Right after a while I'm finally back and ready to review again!

You're right, dealing with optional values makes no sense (how would you know when to display the selection screen, maybe the user didn't want to supply the argument?).

I did find a rather weird bug though which I'll look into rn - typing `:Neorg keybind` and selecting e.g. `norg` from the `select` screen causes it to loop and display the same selection again, and then proceeds to error out that keybind `norg` does not exist lol!

vhyrro:
Wow I'm totally stumped here - `ref_definitions` doesn't contain the right data when executing `:Neorg keybind all` and breaks :thinking:

How does it work under normal conditions then? lol

vhyrro:
Hope you don't mind - I pushed a few commits that fix the main problems, it's now ready to merge! Thanks a bunch for the PR and the cool idea!

PR #461, opened 2022-05-17
feat(gtd.ui): add keybind to add captured task and directly capture a new one

AckslD:
This adds a keybind `n` in the capture popup which adds the current one to the inbox and directly starts capturing a new one.

I was also thinking it could be nice to have one more option to capture another task with the same context as the current one, any thoughts? Although not fully sure how to do that in a nice way.

### DIFF
--- a/lua/neorg/modules/core/gtd/ui/capture_popup.lua
+++ b/lua/neorg/modules/core/gtd/ui/capture_popup.lua
@@ -84,24 +84,11 @@ module.private = {
                         )
                     end)
                     :flag("<CR>", "Add to inbox", function()
-                        local workspace = neorg.modules.get_module_config("core.gtd.base").workspace
-                        local workspace_path = module.required["core.norg.dirman"].get_workspace(workspace)
-
-                        local files = module.required["core.norg.dirman"].get_norg_files(workspace)
-                        local inbox = neorg.modules.get_module_config("core.gtd.base").default_lists.inbox
-                        if not vim.tbl_contains(files, inbox) then
-                            log.error([[ Inbox file is not from gtd workspace.
-                            Please verify if the file exists in your gtd workspace.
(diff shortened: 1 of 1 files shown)


max397574:
haven't looked at the code yet
but would it be possible that you can press n, a variable gets set and then you can still place it at different places (idk if that makes sense)
so you could at any time in the capturing process press `n` and then e.g. add it to cursor position and the capture a new one

AckslD:
Just so I understand you correctly @max397574, do you mean that you when the variable is enable then if you press eg `<CR>` as normal to add to inbox you still get a a prompt for a new capture, and also for placing it at other positions?

I guess it would then be nice if the variable is visualised in some way, with a checkbox or so. Is it easy to redraw the prompt window?

max397574:
yes you got it
I also though about how we should show it
I think it should be possible to somehow display that value with the selection popup

max397574:
@AckslD is this done?

AckslD:
No sorry, I haven't had time to investigate what you suggested.

max397574:
no worries
I was just looking through some prs an issues

AckslD:
Actually @max397574! Do you have any thoughts on how to achieve that, in particular to display the variable in the prompt? I realised that I'm not really sure how to do it so any tips would be useful :)

max397574:
I'll have a look tomorrow.
We plan to rewrite big parts of the capture functionality anyways.
So we could implement this then.

AckslD:
@max397574 Is there a rough timeframe for this re-write? I'm just wondering if it's worth trying to implement this here or better to just wait as you also mention.

danymat:
Closing this as GTD is no longer active on 1.0.0 neorg. We'll create a PR/discussion that centralizes all GTD talk and needs

PR #467, opened 2022-05-25
feat(hop): open external links asynchronously

bR3iN:
First, thank you for your efforts in this project, `neorg` works great for me so far!

However, a major pain point I encountered while using `neorg` is that external links are currently opened synchronously via a simple call to `vim.cmd("xdg-open [...]")`; so when I follow a link to a website inside `neorg`, it will open my browser and then block until the browser is closed. I see a lot of potential in `neorg` for organizing references to websites and local files, but I think this behaviour makes it frankly unusable for this purpose and I don't think this behaviour is what most users would expect or want.

This PR proposes to instead spawn the processes used to open external links via `luvit` in an asynchronous manner; since `neorg` already uses/depends on `plenary.nvim`, this change comes pretty much for free by simply using `plenary.nvim`'s existing functionality.

### DIFF
--- a/lua/neorg/modules/core/norg/esupports/hop/module.lua
+++ b/lua/neorg/modules/core/norg/esupports/hop/module.lua
@@ -9,6 +9,7 @@ require("neorg.modules.base")
 require("neorg.external.helpers")
 
 local module = neorg.modules.create("core.norg.esupports.hop")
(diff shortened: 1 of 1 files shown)


vhyrro:
damn, I never even realized this was a problem - thanks for the contribution! 💜

PR #476, opened 2022-06-06
use new 0.7 apis and refactor cmd system

max397574:
use new 0.7 command api to define the command
allow the command data to have a `callback` field which is a function that will be executed
This allows to get rid of the `events.subscribed` things

eventually I'll also try to refactor the keybindings system so it integrates with this


max397574:
I'll start a new pr for this where I will rewrite parts of the whole module system

PR #542, opened 2022-08-14
fix: fix state_to_text translation for gtd

cosmicboots:
I noticed a bug when trying to view the project view in GTD. It seems that the helper was improperly mapping state to `onhold` rather than `on_hold`, as it's seen everywhere else in the codebase.


vhyrro:
Good catch, thanks :D

PR #543, opened 2022-08-15
feat(journal): implement template feature

Gian12315:
This implements the templates feature for journal, the templates are local to the workspace used.

### DIFF
--- a/lua/neorg/modules/core/norg/journal/module.lua
+++ b/lua/neorg/modules/core/norg/journal/module.lua
@@ -4,7 +4,7 @@
     Summary: Easily create files for a journal.
     ---
 How to use this module:
-This module creates four commands.
+This module creates five commands.
 - `:Neorg journal today`
 - `:Neorg journal yesterday`
 - `:Neorg journal tomorrow`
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/norg/journal/module.lua
```diff
@@ -13,6 +13,9 @@ With this commands you can open the config files for the dates.
 - `Neorg journal custom`
 This command requires a date as an argument.
 The date should have to format yyyy-mm-dd.
+
+- `Neorg journal template`
```
max397574, line 17: add `:` so it's the same as the other ones

lua/neorg/modules/core/norg/journal/module.lua
```diff
@@ -60,10 +65,23 @@ module.private = {
...
+        end
+
         module.required["core.norg.dirman"].create_file(
             folder_name .. neorg.configuration.pathsep .. path,
             workspace or module.required["core.norg.dirman"].get_current_workspace()[1]
         )
+
+        if
+            not journal_file_exists
+            and use_template
```
max397574, line 80: just use the value `module.config.public.use_template` here and not create a local variable

lua/neorg/modules/core/norg/journal/module.lua
Gian12315: Done, Neorg journal custom also needs a `:`

lua/neorg/modules/core/norg/journal/module.lua
Gian12315: Done

vhyrro:
This is nice! Thanks for implementing it :)

Issue #508, opened 2022-07-10
Mistakes in markdown export


jghauser:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.7.2

### Neorg setup

```
require("neorg").setup({
  load = {
    ["core.defaults"] = {}, -- Load all the default modules
    ["core.export"] = {},
    ["core.export.markdown"] = {
      config = {
        extensions = "all",
        },
     },
  },
})
```

### Actual behavior

```norg
@document.meta
title: test
description: 
categories: [
  this
  that
]
created: 2022-07-10
version: 0.0.11
@end

#contexts something
#other tag
* Some heading

  A link without description: {www.wikipedia.org}
  
  A comment: +a comment here+

  @comment
  a multiline comment
  @end
```
```md
---

title: test
description: 
categories: [
  this
  that
]
created: 2022-07-10
version: 0.0.11
---
somethingtag
# Some heading

A link without description: [](www.wikipedia.org)

A comment: <!-- a comment here -->
```

1.  @document.meta
- Unnecessary new line after the first `---` (I guess it's from fixing the missing new lines that existed in earlier version of the exporter)
- list of values is converted into invalid yaml syntax

2. Links without description
- Yes, there is a link there. But the exporter converts `{link}` to `[](link)`, which is then concealed to invisibility (here in neovim, but elsewhere similar things will happen probably. Not sure what the best solution is as the conversation does seem to do the right thing. However, maybe it's better to see links without descriptions as links where the description is the same as the destination. After all, that is how neorg concealing deals with them too.

3. Multiline comments
- Currently the are omitted in the conversation. It would be more elegant to covert them as follows:
``` md
<--
this is 
a comment
-->

4. Tags
- Seem to gobble up newline above it
- single word tags are eaten up (maybe that's good behaviour)
- double word tags have the second word show up as plaintext (definitely not good behaviour I think)
- maybe converting them, but adding them into comments might be useful

These are all I've noticed so far!


### Expected behavior

Described above

### Steps to reproduce

Convert the above norg file to md (with all extensions enabled).

### Potentially conflicting plugins

_No response_

### Other information

_No response_

### Help

Yes, but I don't know how to start. I would need guidance (check question below)

### Implementation help

_No response_

[...]

PR #546, opened 2022-08-17
fix(export.markdown): always save last parsed link

danilshvalov:
Part of #508

Neorg:
```norg
A link without description: {www.wikipedia.org}
```

Before:
```markdown
A link without description: [](www.wikipedia.org)
```

After:
```markdown
A link without description: [www.wikipedia.org](www.wikipedia.org)
```


vhyrro:
Thanks a bunch for fixing these! I've been very busy recently so sorry if my responses are delayed.

Issue #525, opened 2022-07-20
Generate a TOC over multiple files eg. Journal Index file


shxfee:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

I was using VimWiki before I switched over to Neorg for keeping notes and journaling. In VimWiki they had a feature where I could generate a journal index file that is basically a TOC of all the journal entries I have recorded so far. It looked something like this

```
== 2022 ==

=== December ===
    - [[2021-12-30|2021-12-30 What does infinity sound like?]]
    - [[2021-12-29|2021-12-29 Troubled sleep]]
    - [[2021-12-28|2021-12-28 Too much work]]
    - [[2021-12-27|2021-12-27 Lonesome Evening]]

```
It basically generates links to all the files in the journal folder with the first heading. I was hoping this could be implemented in Neorg. Maybe by extending the currently existing TOC module? or adding this to the journal module maybe?

### Help

Yes, but I don't know how to start. I would need guidance

### Implementation help

I have little experience tinkering with Neovim plugins and Lua. I don't know where this belongs or if it belongs.

danymat:
I think we could integrate this feature into the journal module instead of the toc module. Pinging @max397574 as he is the creator of the module.

max397574:
@vhyrro

vhyrro:
Thanks for the ping. This has since been implemented! Closing the issue :)

svenXY:
I just added ```@document.meta
  title: foobarbaz
  description:
  authors:
  categories:
  created:
  version:
@end
```

at the top of my journal entry, but even then it did not add the title but kept adding the day as the title.

The entry is always:
`{ 2023, 3, 17, "{:$/journal/2023/03/17:}", "17" }`

PR #550, opened 2022-08-21
feat(journal): create a toc from all the journal entries

Gian12315:
Using `:Neorg journal toc`, a TOC file will be created inside the journal folder which links to all entries, the format for the toc file is configurable with a user made function (see example)

Basically, just use the journal module however you want, and whenever you wish to see all your entries, just call the command and the file will be created (and updated)


Gian12315:
This closes #525

vhyrro:
Thank you! :purple_heart:

PR #553, opened 2022-08-22
fix(tests): fix stylua errors

danilshvalov:


vhyrro:
Thank you so very much, I've been meaning to fix this for so long :pray:

PR #558, opened 2022-08-28
fix(todo_items_match_color): `onhold` -> `on_hold`

igorlfs:


vhyrro:
Oops, my bad! Thanks for the fix :D

PR #560, opened 2022-08-29
refactor(export): return table instead values

danilshvalov:
I think using tables instead of a list of values will improve extensibility

### DIFF
--- a/lua/neorg/modules/core/export/markdown/module.lua
+++ b/lua/neorg/modules/core/export/markdown/module.lua
@@ -337,23 +337,27 @@ module.public = {
                             tag_indent = tag_start_column - 1,
                             tag_close = module.config.public.metadata["end"],
                             is_meta = true,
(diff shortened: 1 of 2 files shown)


vhyrro:
This is a good idea for sure. You missed some parts of code (mainly at the top of the file), so I'll quickly fix those up if that's okay :p

vhyrro:
I believe this should be everything. There are some conflicts now so I'll try dealing with them before we merge :)

vhyrro:
Not a fan of the commit message autogenerate by github, but i'll squash w/ merge anyway so nobody shall see :p

Could you try this PR out one last time? See if there are any bugs/any weird behaviours. I couldn't find any so if you can't find any as well then it's merge time!

danilshvalov:
Works well for me

vhyrro:
Awesome, merged!

PR #562, opened 2022-08-30
feat(export): export code blocks

danilshvalov:
New feature: export of code blocks. See example:

### Neorg
```norg
* Test title

  @export markdown
  ## Another title

  **bold** *italic*

  <h3>Test subtitle</h3>
  @end

  @export latex
  I don't want to export it to markdown
  @end

  *bold* /italic/
```

### Markdown
```markdown
# Test title



## Another title

**bold** *italic*

<h3>Test subtitle</h3>




**bold** _italic_
```

### DIFF
--- a/lua/neorg/modules/core/export/markdown/module.lua
+++ b/lua/neorg/modules/core/export/markdown/module.lua
@@ -323,7 +323,7 @@ module.public = {
                             tag_close = module.config.public.metadata["end"],
                             is_meta = true,
                         }
-                elseif text == "export"
+                elseif text == "embed"
                     and node:next_sibling()
                     and ts_utils.get_node_text(node:next_sibling())[1] == "markdown"
                 then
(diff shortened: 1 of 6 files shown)


max397574:
read the wiki page on tangling

danilshvalov:
`@tangle` export code block to file. `@export` add lines to file, thats exported

vhyrro:
I like the idea of this to be fair - but I don't think `@export` is a good tag name.

We actually already made a tag for this, and that's the `@embed` tag. Then you do e.g.:
```
@embed markdown
# My Markdown
@end
```

wdyt?

danilshvalov:
`@export` like export in [org-mode](https://orgmode.org/manual/Quoting-HTML-tags.html):
```org
#+BEGIN_EXPORT html
  All lines between these markers are exported literally
#+END_EXPORT
```

vhyrro:
I don't think I like the sound of that to be honest 😓. I'd much rather have something dual-purpose, like `embed`, than having yet another tag just for something specific to the Neorg plugin. Doesn't `embed` offer the same (if not more) functionality versus `export`?

Imagine you're writing `.norg` documents outside of Neorg, and you want to embed some html or markup within your neorg document literally. The obvious choice would be to use an `@embed` tag in this scenario. `@export` sounds like a meaningless name when put outside of the scope of Neorg itself in my opinion.

danilshvalov:
I applied the changes to `embed` instead of `export`

vhyrro:
Awesome! Thank you :)

PR #563, opened 2022-08-30
fix(core.ui): filetype `neorg` -> `norg`

danilshvalov:
Fixed the error that occurred when following a link in toc:
```
Error executing vim.schedule lua callback: ...org/lua/neorg/modules/core/norg/esupports/hop/module.lua:206: attempt to index local 'node_under_cursor' (a nil value)
```

### DIFF
--- a/lua/neorg/modules/core/ui/module.lua
+++ b/lua/neorg/modules/core/ui/module.lua
@@ -131,7 +131,7 @@ module.public = {
             bufhidden = "hide",
             buftype = "nofile",
             buflisted = false,
-            filetype = "neorg",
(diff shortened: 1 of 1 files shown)


vhyrro:
Awesome, glad to hear this fixes it! I'll have to wait a bit before I merge this one cause I have to do some extensive testing to check whether it didn't accidentally break other UI components in the meantime (I believe there was a reason why we set the default filetype to `neorg`). I'll keep you updated :)

vhyrro:
Alright nothing seems to have broken! Merge time.

Issue #509, opened 2022-07-10
Highlighting and concealing in comments


jghauser:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

I use comments a lot to write notes to myself when editing. I think their rendering could be improved in some ways:

- Highlighting: Multiline comments aren't highlighted properly at all. They should be greyed out just like inline comments. I know you're aware of this and that there's some trouble with writing the right treesitter query, I'm just mentioning it here for completeness' sake.

- Concealing: I think it would actually be very nice to have concealing work in comments. I often have lists in multline comments and it would be great if these were formatted as lists usually are in neorg. Even better if the whole autoindent were to work!

### Help

No

PR #564, opened 2022-08-30
feat(highlights): add multiline comments highlight

danilshvalov:
### Before
before

### After
after


danilshvalov:
Part of #509

vhyrro:
Yaay it works, kudos 😄

PR #565, opened 2022-08-30
feat(config): add `lazy_loading` option

danilshvalov:
I think it would be nice to let the user enable or disable lazy loading

### DIFF
--- a/lua/neorg.lua
+++ b/lua/neorg.lua
@@ -28,7 +28,7 @@ function neorg.setup(config)
     })
 
     -- If the file we have entered has a .norg extension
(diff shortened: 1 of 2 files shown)


vhyrro:
This is a good idea! I'm wondering whether we should perhaps disable lazy loading by default? I've seen it cause more confusion in people more than anything, but I also don't want to break everyone's startup time and increase it by like 100ms 😂 

Do you have any thoughts on this? If you think it should stay as `true` by default, we can merge this as is :)

danilshvalov:
I think we should stay as `true` for now

PR #566, opened 2022-08-30
feat: more typical indent

danilshvalov:
I want to suggest a more typical indentation. Here is an example, where `·` is space:

### Before

```norg
* Heading1
··test text

** Heading2
···another test text

```

### After

```norg
* Heading1
··test text

··** Heading2
····another test text
```

### DIFF
--- a/lua/neorg/modules/core/norg/esupports/indent/module.lua
+++ b/lua/neorg/modules/core/norg/esupports/indent/module.lua
@@ -32,7 +32,6 @@ module.public = {
             return 0
         end
 
-        -- local initial_indent = module.required["core.integrations.treesitter"].get_node_range(node).column_start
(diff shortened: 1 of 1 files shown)


vhyrro:
Hmm, this should definitely be put in a preset imo. I don't think it'd be difficult to make indent presets would it? The indent that's currently used in neorg is the most "common" and also (from my experience asking people) the preferred way of indenting. Org mode also does not indent headings (but that's for a different reason versus just style).

Would you be able to make a simple preset system for indents, or should that be a TODO for future me?

vhyrro:
I still have indent presets on my TODO list, just at a very low priority because GTD is consuming all of my time :p 

Closing this to remove some clutter in the PR list :)

PR #568, opened 2022-08-31
fix(hop): open external links in Windows

eriedaberrie:
Windows' `start` is shell builtin of `cmd.exe` and not an executable, which makes it unusable by `plenary.nvim`'s job module. It also has some strange behavior like launching a new shell instead of properly opening the link if the user surrounds the link in quotes, so I've just decided to instead directly call `url.dll` with `rundll32.exe` (Netrw's `gx` also uses this).


vhyrro:
I see! Fair enough, hope this works for other windows users as well :)

PR #583, opened 2022-09-17
feat: add semantic analyzer

vhyrro:
This PR aims to implement a fully fledged semantic analyzer into Neorg.

With such awesome tech we could do *much* more than currently possible with simple treesitter queries, i.e.:
- Create LSP-like features (enumerate dead links, show link previews etc.)
- Understand correlations between headings
- Zettelkasten (by using the above feature to track backlinks across files)

I'm very excited to take on this challenge, but as you can imagine it's proving itself very difficult. Might take a while, might not. We'll see :p


vhyrro:
Unfortunately I gotta close this one. For large codebases this thing will *hog* memory, and there's no nice way of doing any hassle-free database handling w/ lua. The core for the semantic analyzer will be written in another language, and the .so file will be shipped with Neorg directly and used within lua as a binding.

PR #586, opened 2022-09-22
feat: disable lazy loading by default

vhyrro:
It's time for some changes. Yes, they are breaking changes, I know nobody likes them, but here they are definitely for the better.
If you don't like the change, feel free to use this configuration:
```lua
require("neorg").setup {
	lazy_loading = true,
	load = { ... },
}
```

# Overview
- `:NeorgStart` no longer exists and you always have access to a `:Neorg` command
- Highlights are always active, which means that e.g. highlights in telescope previews work now

# Side effects
- Some commands will error out if you use them outside of a `.norg` file
- Your config's startup time will be increased by just under 30 milliseconds (on my hardware at least)

# Notes
- Apart from reenabling the `lazy_loading` flag, you can also lazy load using your plugin manager via e.g. the `ft = "norg"` flag in packer.
- Work will be done to optimize existing modules for faster startup

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,14 +143,12 @@ You can install it through your favorite plugin manager:
   
 - [Packer (with lazyloading)](https://github.com/wbthomason/packer.nvim):
 
-  Want to lazy load? Know that you'll have to jump through some hoops and hurdles to get
-  it to work perfectly.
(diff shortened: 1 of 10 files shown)


vhyrro:
Ready. Set. Go!

PR #587, opened 2022-09-23
feat(tangle): properly tangle escaped norg code

danilshvalov:
Let's say we need to tangle the `norg` code (with escape char `\`):

```norg
+a.norg+

#tangle b.norg
@code norg
\@math
123 + 456
\@end
@end
```

Without this patch you will get:

```norg
+b.norg+

\@math
123 + 456
\@end
```

After this patch you will get:

```norg
+b.norg+

@math
123 + 456
@end
```


vhyrro:
Nice! Consider this a temporary fix until Neorg gets the `|example` ranged tag support in the current specification.

PR #605, opened 2022-10-05
promote demote things

max397574:


max397574:
for now this pr adds the following:
promote or demote headings, ordered lists, unordered lists and quotes with 
>> and <<
this falls back to left and right shift (normal behavior)
it also adds the functionality to operate on a range (only demoting and promoting, no shifting)
(also reconceals things (implemented reconceal range functionality)
you an use it like this
`vnoremap >> :<cr><cmd>Neorg keybind all core.promo.promote_range<cr>`
the `:<cr>` is required because `Neorg` cmd doesn't accept a range

todo:
reindent things below the headings (when single things and when range)

max397574:
you can tell me if I should change anything or squash commits

PR #606, opened 2022-10-05
docs: update link to specification file

Austin-Ray:
Commit c5345e72ebdd84585eec6272755aa26233431317 removed the specification file from this repository so the link at the start of `:h neorg` 404s. This commit changes to link to point to the specification directory.


vhyrro:
Good catch! Thanks :)

PR #610, opened 2022-10-07
fix: jump to next/prev headings or links

### LINKED ISSUE #609: Jumping to next/prev headings or links doesn't work correctly
### Prerequisites

### Neovim Version

NVIM v0.8.0-1210-gd367ed9b2

### Neorg setup

```
require('neorg').setup{
  load = {
    ["core.defaults"] = {},
    ["core.keybinds"] = {
      config = {
        hook = function(keybinds)
          keybinds.map_event_to_mode("norg", {
            n = {
              { "<Tab>", "core.integrations.treesitter.next.link" },
              { "<S-Tab>", "core.integrations.treesitter.previous.link" },
            },
          },{
            silent = true,
            noremap = true,
          })
        end,
      }
    },
  }
}

```

### Actual behavior

Any link on the line directly below the cursor is skipped.
Any link on the same line after the cursor is skipped.

### Expected behavior

Jump to the next link, regardless of the current cursor position.

### Steps to reproduce

Consider the following file:

```
{:link1:}
{:link2:}
{:link3:} {:link4:} {:link5:}
{:link6:}
```

With the cursor at the start of the document, repeatedly press Tab.

_No response_

### Other information

_No response_

### Help

Yes

### Implementation help

_No response_

jstemmer:
The implementation of the actions that allow you to jump to the next or previous heading/link contains a small bug that causes the matching node on the next line to be skipped. It also ignores matching nodes that occur after the cursor on the same line.

Fixes #609

### DIFF
--- a/lua/neorg/modules/core/integrations/treesitter/module.lua
+++ b/lua/neorg/modules/core/integrations/treesitter/module.lua
@@ -132,7 +132,8 @@ module.public = {
     --- Jumps to the next match of a query in the current buffer
     ---@param query_string string Query with `@next-segment` captures
     goto_next_query_match = function(query_string)
-        local line_number = vim.api.nvim_win_get_cursor(0)[1]
+        local cursor = vim.api.nvim_win_get_cursor(0)
+        local line_number, col_number = cursor[1], cursor[2]
 
         local document_root = module.public.get_document_root(0)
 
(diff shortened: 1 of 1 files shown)


vhyrro:
Thanks a lot for the fix! Works well in my tests :)

PR #612, opened 2022-10-08
fix(concealer): don't update `content_only` variable if `conceal` is false and `adaptive` is true

luck07051:
I think adaptive should be disable when conceal not enable.
Otherwise the behavior when { adaptive = true, conceal = false } kinda weird (only background adaptive).


vhyrro:
Good catch! I agree that your behaviour is the more sensible one.

Issue #615, opened 2022-10-09
Done and undone symbols are swapped in concealer module


greenspaceexplorer:
### Prerequisites

- [x] I am using either 0.7 stable with the `0.0.12` tag or the latest compiled Neovim version
- [X] I am using the latest version of the plugin

### Neovim Version

v0.8.0-1210-gd367ed9b2

### Neorg setup

```lua
require('neorg').setup {
        load = {
                ["core.defaults"] = {},
                ["core.norg.completion"] = {
                        config = {
                                engine = 'nvim-cmp'
                        },
                },
                ["core.integrations.nvim-cmp"] = {},
                ["core.integrations.telescope"] = {},
                ["core.norg.concealer"] = {},
                ["core.norg.dirman"] = {
                        config = {
                                workspaces = {
                                        research = "~/iCloud/Research/Research_stow/Notes/Research", 
                                },
                                autochdir = true, -- Automatically change the directory to the current workspace's root every time
                                index = "index.norg", -- The name of the main (root) .norg file
                        }
                },
        }
}
```

### Actual behavior

The done and undone symbols are swapped in the concealer module.
<img width="377" alt="Screen Shot 2022-10-08 at 9 03 22 PM" src="https://user-images.githubusercontent.com/16415256/194732926-e9fcd603-155e-49e8-94ca-03505e9130c1.png">

### Expected behavior

`- [ ] incomplete`
`- [x] complete`

...not the other way around.

### Steps to reproduce

* Install concealer module
* Make a to-do list in any .norg file

### Help

Yes

max397574:
this is expected
it looks like a font issue
the done tasks should display a tick

greenspaceexplorer:
Was getting fooled by my terminal choice (Alacritty). Also, fooled by `U+00D7` being "incomplete" when `- [x]` is complete.  Changed to OSX terminal to see the following:

<img width="272" alt="Screen Shot 2022-10-09 at 11 12 40 AM" src="https://user-images.githubusercontent.com/16415256/194764895-5bd7f798-df51-424d-bc32-96eca2b25eda.png">

max397574:
looks like your font is missing some symbols then
ig this issue can be closed?

greenspaceexplorer:
[...]

PR #616, opened 2022-10-09
fix: Corrected "done" and "undone" symbols in concealer module

greenspaceexplorer:
Just needed to do a simple symbol swap. Solves #615


max397574:
this isn't an issue of the symbols
it's a font issue
Screenshot 2022-10-09 at 09 06 07
is what it looked like before

greenspaceexplorer:
Actual problem: not all unicode characters being rendered. 
Screen Shot 2022-10-09 at 11 12 40 AM

PR #618, opened 2022-10-09
add `core.execute`

tamton-aquib:
This module is aimed to provide an `execute` command for neorg.

https://user-images.githubusercontent.com/77913442/198702666-8312fb76-c7e6-408e-929c-663c93f38297.mp4

Possible todos:
- [ ] cross platform support.
- [x] spinner for running block.
- [ ] run for
    - [x] compiled (c, cpp, rust, etc)
        - [ ] check extra parameters for wrapping in main function? (+ named blocks)
    - [x] interpreted (python, lua, bash, etc)
- [x] subcommands
    - [x] view (virtual_lines)
    - [x] normal (maybe a better command name, normal lines added to buffer.)
- [ ] add logging?
- [ ] check for timeout / limit number of output lines.
- [x] make non-blockable if possible (weird python behaviour)
- [ ] user config options.
- [x] set lines to separate files and run.
- [x] panic messages instead of empty returns.
- [x] assign state for each running block instead of tracking the current one. (will fix spinners and re-execution bugs)
- [ ] code cleanup.


lua/neorg/modules/core/execute/module.lua
```diff
@@ -73,11 +73,20 @@ module.private = {
             local curr_task = module.private.tasks[id]
             curr_task.spinner = spinner.start(curr_task, module.private.ns)
 
+            if not vim.tbl_isempty(curr_task.output) then
+                vim.api.nvim_buf_set_lines(
+                    curr_task.buf,
+                    curr_task.code_block['end'].row + 1,
+                    curr_task.code_block['end'].row + #curr_task.output + 1,
```
max397574, line 80: shouldn't this be the same as the `start`
if not lines below the code block are overwritten

lua/neorg/modules/core/execute/module.lua
```diff
@@ -101,11 +110,19 @@ module.private = {
     },
 
     init = function()
-        local id = vim.api.nvim_buf_set_extmark(
-            0, -- NICE: check if errors
-            module.private.ns,
-            0, 0, {}
-        )
+        -- IMP: check for existng marks and return if it exists.
+        local cr, _ = unpack(vim.api.nvim_win_get_cursor(0))
```
max397574, line 114: couldn't you just get rid of the `, _` ?
the value would just be ignored
also
I think `cr` isn't a really descriptive variable name

lua/neorg/modules/core/execute/module.lua
```diff
@@ -0,0 +1,274 @@
...
+            subcommands = {
+                view = { args=0, name="execute.view" },
+                normal = { args=0, name="execute.normal" },
+            }
+        }
+    })
+end
+
+module.config.public = require("neorg.modules.core.execute.config")
+module.config.private = {}
```
max397574, line 28: no need to add this if it's unused

lua/neorg/modules/core/execute/module.lua
```diff
@@ -0,0 +1,274 @@
+---@diagnostic disable: undefined-global
+require("neorg.modules.base")
+local spinner = require("neorg.modules.core.execute.spinner")
+
+local module = neorg.modules.create("core.execute")
+local ts = require("nvim-treesitter.ts_utils")
```
max397574, line 6: we got a module for this
not sure what's preferred

lua/neorg/modules/core/execute/spinner.lua
```diff
@@ -0,0 +1,41 @@
+-- TODO: code cleanup
```
max397574, line 1: @vhyrro perhaps add some standard location of things like this? (basically util files)

lua/neorg/modules/core/execute/module.lua
tamton-aquib: > couldn't you just get rid of the , _ ?

Was indexing the first item(row), but was testing with both.

> I think cr isn't a really descriptive variable name

let me know what else you want to call it

lua/neorg/modules/core/execute/module.lua
tamton-aquib: There was some settings on my local repo, will delete the line if its not used 👍🏻

lua/neorg/modules/core/execute/module.lua
tamton-aquib: will discuss and resolve with vhyrro 👍🏻

lua/neorg/modules/core/execute/module.lua
tamton-aquib: This is for deleting the previous output lines if present, havent experienced any bugs yet on my side(on this line).

lua/neorg/modules/core/execute/module.lua
vhyrro: Here you probably want to use `module.required["core.integrations.treesitter"].get_ts_utils()`

vhyrro:
By the way, this module is not forgotten. I think about this every now and again but just have other priorities. The thought of implementing code execution does excite me though heh

laher:
Hi @tamton-aquib ... I just copy-pasted this as an external module on my filesystem, and it's working beautifully for me. Very cool work. I just had to `s/core.execute/external.execute/` in a few places. It's so good for my journal template to have some daily tasks which I can trigger like this - thank you 🥇 

What do you think about re-creating this as an external module & publishing in its own repo, while this core module is under review? It might take a while for this to get merged as a core module, and in the meantime we could collaborate. I & others could fork & contribute. (Or I can publish my copy-paste and credit you, but that would feel like the wrong way round).

I know it belongs in `core` eventually, but I think that if folks could start using it now, then they would & people would help to improve it. 

Thanks again

tamton-aquib:
Heyyo @laher ,

> What do you think about re-creating this as an external module & publishing in its own repo, while this core module is under review? It might take a while for this to get merged as a core module, and in the meantime we could collaborate

I talked to vhyrro on discord and it might take some time for him to review this PR.
So for the time being i guess you could create the fork because i wont probably be able to maintain it 😬  .

> (Or I can publish my copy-paste and credit you, but that would feel like the wrong way round)

This would be better. I guess you could just link this PR instead of credits btw 😅 

PS: i dont know if the code is ugly, so feel free to refactor or anything hehe

laher:
Cool, I'll do that. I haven't written much lua myself, but what you've done is working & simple enough that I could add a couple of languages easily. 

I'll add some linting and stuff and go from there...

I'll pop onto discord for discussion once I have something.

laher:
Hi again, I've been reading the specs and getting my head around tags and norg generally, and I think I know where I'd like to take this.

Repo: https://github.com/laher/neorg-exec

 * I named it `external.exec` because it's 3 whole characters shorter, and I'm lazy. I gave it a GPL3 like neorg because it matches.
 * I think it'd be tidy to use ranged tags for the results sections, so it's easier to address them for re-runs. (could be verbatim or standard, depending on the use case).
 * I'd like to use tags to indicate how to run the code block, something like `#exec cache=5m pwd=.. result.tagtype=|` 
 * also tags to present metadata about the results. exit code, code-hash for cached results, timings, etc.
 * I'd like to use popups instead of virtual lines for the `virtual` mode.

See the planning section for examples of what this might look like: https://github.com/laher/neorg-exec#planning

I'm very new to norg, so I'd appreciate some feedback about how well this tagging approach might work with norg.

OK I'll stop spamming this PR now and move discussion into discord + my repo.
Ta

tamton-aquib:
Closing the PR as the development regarding this features is moved to an external module:
https://github.com/laher/neorg-exec 🎉

PR #637, opened 2022-10-20
fix: do not skip anchors when jumping to next/prev link

jstemmer:
Anchor definitions and anchor declarations were skipped when jumping to the next or previous link. This was a bit surprising as you could jump to links with a custom text for example. Since anchors serve a similar purpose as links, it makes sense to be able to jump to them.

To test this, use the same Neorg setup as described in issue #609, then create a new norg file with the following contents:

```
{:link1:}
{:link2:}[description]
[anchor]
[anchor]{:link3:}
{:link4:}
```

With the cursor on the first line, repeatedly press `Tab`.


vhyrro:
Awesome, works like a charm on my end. Thanks for the PR!

PR #639, opened 2022-10-22
ref: use `vim.api` for autocmds

danilshvalov:


vhyrro:
`vim.api` for the win! Thanks :)

Issue #642, opened 2022-10-25
Regression: folke/todo-comments.nvim comments in norg file are no longer highlighted?


xfzv:
### Prerequisites

- [ ] I am using either 0.7 stable with the `0.0.12` tag or the latest compiled Neovim version
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.8.0

### Neorg setup

```lua
require('neorg').setup {
    load = {
        ["core.defaults"] = {},
        ["core.norg.concealer"] = {
            config = {
                dim_code_blocks = {
                    enabled = true,
                    conceal = true,
                    content_only = true,
                    adaptive = true,
                },
                icon_preset = "diamond",
            },
        },
    },
}
```

### Actual behavior

A while ago, I opened https://github.com/folke/todo-comments.nvim/issues/106 and the issue was fixed by https://github.com/nvim-neorg/neorg/commit/887dc41f508aea1d1c9f195dba18a3aac0a2e1c8.

The problem re-appeared recently, so I re-opened the [original issue](https://github.com/folke/todo-comments.nvim/issues/106#issuecomment-1276649002).

I noticed that `~/.local/share/nvim/site/pack/packer/start/neorg/queries/norg/highlights.scm` no longer had `(inline_comment) @comment` anywhere. I tried to add it back myself and it solved the issue. Not sure why it was removed? Is there another fix I'm not aware of or could this be re-implemented?

In current state, without `(inline_comment) @comment`:

![image](https://user-images.githubusercontent.com/78810647/197784045-9641ec89-9172-4241-a033-45d4540618bb.png)

### Expected behavior

When adding back myself `(inline_comment) @comment` to `~/.local/share/nvim/site/pack/packer/start/neorg/queries/norg/highlights.scm`, the comments in a `.norg` file are correctly highlighted:

![image](https://user-images.githubusercontent.com/78810647/197783686-657906fa-44b7-459e-9e6b-fe71a87f75fa.png)

### Steps to reproduce

1. Install `nvim-lua/plenary.nvim`, `folke/todo-comments.nvim` and Neorg
2. Open a blank `.norg` file
3. Try to add some comments like
```
+PERF:+
+HACK:+
+TODO:+
+NOTE:+
+FIX:+
+WARN:+
``` 
4. They're not highlighted

### Help

No

max397574:
if adding the query mentioned above fixes the issue you could open a pull request for it

max397574:
yes

PR #643, opened 2022-10-25
fix: `folke/todo-comments.nvim` comment highlighting in `.norg` files

xfzv:
- Fixes https://github.com/nvim-neorg/neorg/issues/642  ; https://github.com/folke/todo-comments.nvim/issues/106
- Original commit: https://github.com/nvim-neorg/neorg/commit/887dc41f508aea1d1c9f195dba18a3aac0a2e1c8


vhyrro:
Thanks for this! I checked and this shouldn't interfere with the previous highlight query for the same node so here's hoping :smile:

Issue #559, opened 2022-08-28
Treesitter errors "W18: Invalid character in group name" "Unable to parse"


jwbaldwin:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.8.0-dev-922-g42e9fe7d9

### Neorg setup
```
# custom.p.neorg.lua
local neorg = require("neorg")

neorg.setup {
  load = {
    ["core.defaults"] = {},
    ["core.gtd.base"] = {
      config = {
        workspace = "project"
      }
    },
    ["core.norg.concealer"] = {
    },
    ["core.norg.completion"] = {
      config = {
        engine = "nvim-cmp"
      }
    },
    ["core.norg.dirman"] = {
      config = {
        workspaces = {
          project = "~/neorg/project",
        }
      }
    }
  }
}

#nvchad install of plugins see https://nvchad.com/config/plugins
  ["nvim-neorg/neorg"] = {
    after = "nvim-treesitter",
    config = function()
      require "custom.p.neorg"
    end
  },
```

### Actual behavior

1. Start up Neovim
2. run `:NeorgStart`
3. Get the following error:
```
W18: Invalid character in group name
W18: Invalid character in group name
```
4. Press escape to continue to empty index.norg
5. Try to edit and recieve the following error:
```
...ua/neorg/modules/core/integrations/treesitter/module.lua:503
Unable to parse the current document's syntax tree :(
```
6. If I run `:Neorg inejct-metadata` the error from 5. never appears again unless I leave the `.norg` file and try something like `:Neorg gtd views`

Other information:
- Packer install contains `after = "nvim-treesitter"`
- No syntax highlighting happens on the document EXCEPT code snippets work
- `:highlight` doesn't show any Neorg group
- output from `:checkhealth nvim-treesitter` looks good:
```
nvim-treesitter: require("nvim-treesitter.health").check()
========================================================================
## Installation
  - OK: tree-sitter found 0.20.0 (e85a279cf29da1b08648e27214dda20a841e57c8) (parser generator, only needed for :TSInstallFromGrammar)
  - OK: node found v18.7.0 (only needed for :TSInstallFromGrammar)
  - OK: git executable found.
[...]

PR #649, opened 2022-11-02
Update README.md

ssiccha:
I'm using NVIM v0.8.0-dev+177-gcf68f0a51 and got the same errors as in https://github.com/nvim-neorg/neorg/issues/559. It's nice that that issue is pinned, so that I found the issue relatively quickly. Stating this in the README is still a bit easier to discover.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,7 @@ will also be updated to the correct revision.
 
 ### Troubleshooting Treesitter
 - Not using packer? Make sure that Neorg's `setup()` gets called after `nvim-treesitter`'s setup.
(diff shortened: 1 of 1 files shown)


max397574:
you should use 0.8 stable then it should work. You're on an older 0.8 version which perhaps doesn't have the patch yet.

ssiccha:
>you should use 0.8 stable then it should work. You're on an older 0.8 version which perhaps doesn't have the patch yet.

Good to know, thanks! Then I'll update my Neovim and can get rid of the tag fix. In that case I'd reword the PR slightly.

max397574:
imo the only thing that could perhaps be changed is "neorg requires at least nvim 0.8 **stable** to operate"

ssiccha:
Ah, I see. I wasn't aware that recently there was an official 0.8 release.

ssiccha:
In that case I'll close the PR. I forgot that I installed a dev version of Neovim. :grin:

PR #664, opened 2022-11-19
Upgrade tool

vhyrro:
Need I say more.

TODOs:
- [ ] Some extra documentation
- [ ] "Upgrade all workspaces" switch

But those will be done later, let's merge this right here, right now :p

### DIFF
--- a/docgen/init.lua
+++ b/docgen/init.lua
@@ -204,7 +204,7 @@ docgen.get_module_top_comment = function(path)
 
     return {
         buf = buf,
-        comment = comment
+        comment = comment,
     }
 end
 
@@ -686,7 +686,9 @@ docgen.generate_md_file = function(buf, path, comment, main_page)
 
                 if not vim.tbl_isempty(api) then
                     -- sort api in order to not shuffle each time we want to commit
-                    table.sort(api--[[@as table]])
+                    table.sort(
+                        api --[[@as table]]
+                    )
                     for function_name, item in pairs(api) do
                         if type(item) == "function" then
                             table.insert(results, "- `" .. function_name .. "`")
@@ -978,7 +980,10 @@ docgen.generate_md_file = function(buf, path, comment, main_page)
                 table.insert(output, item)
             end
         elseif type(item) == "table" then
-            local query = docgen.get_module_queries(buf--[[@as number]], item.query)
+            local query = docgen.get_module_queries(
+                buf, --[[@as number]]
+                item.query
+            )
 
             if query then
                 local ret = item.callback(query)
--- a/lua/neorg/modules/core/defaults/module.lua
+++ b/lua/neorg/modules/core/defaults/module.lua
@@ -27,5 +27,6 @@ return neorg.modules.create_meta(
     "core.promo",
     "core.storage",
     "core.syntax",
-    "core.tangle"
+    "core.tangle",
+    "core.upgrade"
 )
--- a/lua/neorg/modules/core/fs/module.lua
+++ b/lua/neorg/modules/core/fs/module.lua
@@ -1,0 +1,39 @@
+--[[
+    File: Filesystem
+    Title: Module for Filesystem Operations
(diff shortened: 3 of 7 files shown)


jghauser:
I've tried this on one file so far and I get the following error:

```
Error executing vim.schedule lua callback: .../packer/start/plenary.nvim/lua/plenary/window/border.l
ua:219: String cannot contain newlines
stack traceback:
        [C]: in function 'nvim_buf_set_lines'
        .../packer/start/plenary.nvim/lua/plenary/window/border.lua:219: in function '__align_calc_c
onfig'
        .../packer/start/plenary.nvim/lua/plenary/window/border.lua:263: in function 'new'
        ...ack/packer/start/plenary.nvim/lua/plenary/popup/init.lua:384: in function 'create'
        ...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:327: in function '_create_window
'
        ...ck/packer/start/telescope.nvim/lua/telescope/pickers.lua:397: in function 'find'
        ...er/start/dressing.nvim/lua/dressing/select/telescope.lua:134: in function 'select'
        .../packer/start/dressing.nvim/lua/dressing/select/init.lua:63: in function ''
        vim/_editor.lua: in function <vim/_editor.lua:0>
```

And then an empty float opens (I assume it's supposed to ask about backup?). It upgrades the file (without creating a backup file as far as I can tell).

Also, in the converted file, list items aren't indented (I need to enter/exit insert mode at the relevant line to get it to update it). Maybe this is because I have a custom indent tweaks...

vhyrro:
Yeah I assume you're using e.g. dressing.nvim? The way external plugins overwrite `vim.ui.select` is pretty bad - not only does the prompt disallow newlines (which are allowed in the base implementation) it also seems like the call is asynchronous (from the behaviour others have told me about), rendering the whole select call useless.

Unsure what I can do about it at this point in time :/

max397574:
I already discovered that issue
problem is with telescope-ui-select
1. because prompt title only takes a string (so you can't make it a table (telescope)) and this cannot contain newlines (plenary)
2. that's also a telescope-ui-select issue
3. that's a real bug

nothing that really can be done about 1 and 2
perhaps open an issue there

jghauser:
Ok, thanks for the information! It might be good to default to creating a backup, so that if something goes wrong in the way it did in my config, files aren't just overwritten. It wasn't a problem for me personally, as I could just undo the changes, but default backups might save someone else a scare ;).

vhyrro:
@jghauser I think I might've fixed this completely (with support for custom `vim.ui.select` UIs). Mind checking it out on the latest version pls?

jghauser:
I'm afraid this has made things worse :laughing:. What happens now is that I can see a notification asking me whether I want to create a backup. But no UI opens and I can't seem to interact with neovim at all anymore at that point and need to force close the app.

Actually, even the notification doesn't show up if I disable "desktop-notify.nvim" (I use it to have notifications display as system notifications, and disabled it to see if it was causing issues. But that doesn't seem to be the problem.)

vhyrro:
Oh no :flushed:

vhyrro:
Okay @jghauser I assume you were trying the directory export right? Should be fixed now, made a logic error that triggered an infinite loop lol

jghauser:
I actually tried a file export. While I'm happy my report led to a fix with directory exports, I'm afraid my issue still exists. No float or other UI is opened at all... and now that you mention infinite loop, I think there's still one hidden somewhere, at least my nvim process is taking up 100% after entering the upgrade command ;)

vhyrro:
Okay, take three - I removed the async-specific code - it was the cause for the whole infinite loop thing. Now I wonder if we'll encounter the same problem with it beginning the upgrade before waiting for the user to select an option. Let me know :smile:

jghauser:
:partying_face:  this fixed it! 

Upgrade only happens once I select something, and everything seems to be working as expected as far as I can tell :)

Issue #670, opened 2022-11-20
No highlighting in presenter mode


TornaxO7:
### Prerequisites

- [X] I am using either 0.7 stable with the `0.0.12` tag or the latest compiled Neovim version
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.9.0-dev-5e1687e-dirty Build type: RelWithDebInfo LuaJIT 2.1.0-beta3 Compilation: /usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions     -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security     -fstack-clash-protection -fcf-protection -Wp,-U_FORTIFY_SOURCE -Wp,-D_FORTIFY_SOURCE=1 -DNVIM_TS_HAS_SET_MATCH_LIMIT -DNVIM_TS_HAS_SET_ALLOCATOR -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/tornax/.cache/paru/clone/neovim-git/src/build/cmake.config -I/home/tornax/.cache/paru/clone/neovim-git/src/neovim-git/src -I/usr/include -I/home/tornax/.cache/paru/clone/neovim-git/src/build/src/nvim/auto -I/home/tornax/.cache/paru/clone/neovim-git/src/build/include Compiled by tornax@TORNAX-MAIN  Features: +acl +iconv +tui See ":help feature-compile"     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/usr/share/nvim"  Run :checkhealth for more info

### Neorg setup

```lua
local neorg = require("neorg")
neorg.setup({
    load = {
        ["core.defaults"] = {},
        ["core.norg.dirman"] = {
            config = {
                workspaces = {
                    main = "/main/Dokumente/NeorgNotes",
                }
            }
        },
[...]

PR #672, opened 2022-11-21
fix: set filetype when creating new norg buffer

TornaxO7:
This PR fixes [my issue](https://github.com/nvim-neorg/neorg/issues/670).

image

### DIFF
--- a/lua/neorg/modules/core/norg/journal/module.lua
+++ b/lua/neorg/modules/core/norg/journal/module.lua
@@ -136,10 +136,7 @@ module.private = {
             workspace_path .. "/" .. folder_name .. neorg.configuration.pathsep .. path
         )
 
-        module.required["core.norg.dirman"].create_file(
-            folder_name .. neorg.configuration.pathsep .. path,
-            workspace
-        )
+        module.required["core.norg.dirman"].create_file(folder_name .. neorg.configuration.pathsep .. path, workspace)
 
         if
             not journal_file_exists
--- a/lua/neorg/modules/core/norg/journal/module.lua
+++ b/lua/neorg/modules/core/norg/journal/module.lua
@@ -102,7 +102,7 @@ end
 module.private = {
     --- Opens a diary entry at the given time
     ---@param time number #The time to open the journal entry at as returned by `os.time()`
-    ---@param custom_date string #(Optional) A YYYY-mm-dd string that specifies a date to open the diary at instead
+    ---@param custom_date? string #A YYYY-mm-dd string that specifies a date to open the diary at instead
     open_diary = function(time, custom_date)
         local workspace = module.config.public.workspace or module.required["core.norg.dirman"].get_current_workspace()[1]
         local folder_name = module.config.public.journal_folder
--- a/lua/neorg/modules/core/norg/journal/module.lua
+++ b/lua/neorg/modules/core/norg/journal/module.lua
@@ -102,11 +102,11 @@ end
 module.private = {
     --- Opens a diary entry at the given time
     ---@param time number #The time to open the journal entry at as returned by `os.time()`
-    ---@param custom_date string #A YYYY-mm-dd string that specifies a date to open the diary at instead
+    ---@param custom_date string #(Optional) A YYYY-mm-dd string that specifies a date to open the diary at instead
     open_diary = function(time, custom_date)
-        local workspace = module.config.public.workspace
-        local folder_name = module.config.public.journal_folder
-        local template_name = module.config.public.template_name
+        local workspace_c = module.config.public.workspace
+        local folder_name_c = module.config.public.journal_folder
+        local template_name_c = module.config.public.template_name
 
         if custom_date then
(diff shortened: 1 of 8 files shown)


lua/neorg/modules/core/ui/module.lua
```diff
@@ -298,6 +298,13 @@ module.public = {
             config = { config, "table", true },
             opts = { opts, "table", true },
         })
+
+        if config == nil then
+            config = { ft = "norg" }
+        else
+            config.ft = "norg"
+        end
+
```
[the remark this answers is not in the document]
danilshvalov: Maybe it would be better to do it this way:
```lua
config = config or {}
config.ft = "norg"
```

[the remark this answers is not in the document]
TornaxO7: dayum, good idea!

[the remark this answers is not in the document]
TornaxO7: done!

TornaxO7:
hm.... I don't really understand why I'm getting here a linter error :thonk:

danilshvalov:
I think you shouldn't use `*_c` variables. Just overwrite the existing variables (without `local`). For example:
```lua
local workspace = module.config.public.workspace
-- ...
workspace = workspace or module.required["core.norg.dirman"].get_current_workspace()[1]
```

Or, where possible, immediately write down the value of the variable:
```lua
local workspace = module.config.public.workspace or module.required["core.norg.dirman"].get_current_workspace()[1]
```

lua/neorg/modules/core/norg/journal/module.lua
```diff
@@ -102,11 +102,11 @@ end
 module.private = {
     --- Opens a diary entry at the given time
     ---@param time number #The time to open the journal entry at as returned by `os.time()`
-    ---@param custom_date string #A YYYY-mm-dd string that specifies a date to open the diary at instead
+    ---@param custom_date string #(Optional) A YYYY-mm-dd string that specifies a date to open the diary at instead
     open_diary = function(time, custom_date)
-        local workspace = module.config.public.workspace
-        local folder_name = module.config.public.journal_folder
-        local template_name = module.config.public.template_name
+        local workspace_c = module.config.public.workspace
```
max397574, line 107: why renaming all those variables?

TornaxO7: because the linter is complaining about it: https://github.com/nvim-neorg/neorg/actions/runs/3531239229/jobs/5924180898#step:3:116

max397574: that warning means you should just remove the `local` keyword

TornaxO7: oh bruh, ok

TornaxO7: @max397574 should I just remove the `local` then? Isn't this creating a global variable?

danilshvalov: If the variable was previously declared, then this will write the new value to the variable. The global variable will not be created. See my comment above.

TornaxO7: I'm still getting the linter error ._.

danilshvalov: I'm not sure, but it seems CI is incorrectly configured. It seems it is checking the main branch, not the PR. But I'm not sure about that.

lua/neorg/modules/core/norg/journal/module.lua
```diff
@@ -102,9 +102,9 @@ end
 module.private = {
     --- Opens a diary entry at the given time
     ---@param time number #The time to open the journal entry at as returned by `os.time()`
-    ---@param custom_date string #A YYYY-mm-dd string that specifies a date to open the diary at instead
+    ---@param custom_date string #(Optional) A YYYY-mm-dd string that specifies a date to open the diary at instead
```
max397574, line 105: just use string|nil as the type and remove that optional

max397574:
did you run stylua in the terminal?
because it seems like you used different setting than were used before

lua/neorg/modules/core/norg/journal/module.lua
vhyrro: or make it optional:
```suggestion
    ---@param custom_date? string #A YYYY-mm-dd string that specifies a date to open the diary at instead
```

TornaxO7: done!

lua/neorg/modules/core/ui/module.lua
```diff
@@ -298,6 +298,10 @@ module.public = {
             config = { config, "table", true },
             opts = { opts, "table", true },
         })
+
+        config = config or {}
+        config.ft = "norg"
```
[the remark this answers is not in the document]
TornaxO7: done! I didn't know about the `tbl_deep_extend`

vhyrro:
Awesome, thanks for the fixes! :purple_heart:

PR #673, opened 2022-11-22
Fix file_exists ignoring cwd/workspace difference

### LINKED ISSUE #662: Journal TOC gets added multiple times
### Prerequisites

### Neovim Version

NVIM v0.8.1

### Neorg setup

```
require('neorg').setup {
	load = {
		["core.defaults"] = {},
		["core.norg.concealer"] = {},
		["core.norg.completion"] = {
			config = {
				engine = "nvim-cmp",
			}
		},
		["core.norg.journal"] = {
			config = {
				workspace = "default",
			}
		},
		["core.norg.dirman"] = {
			config = {
				workspaces = {
					default = "~/Documents/neorg",
				}
			}
		}
	}
}

```

### Actual behavior

Open nvim, type `:Neorg journal today` and toc gets added  regardless of it already existing.

(On another note, setting `journal_folder` to an empty string causes it to never get added.)

### Expected behavior

Table of contents only gets added the first time the file is created.

### Steps to reproduce

Open nvim and open the journal via neorg. Doesn't happen if I first open a norg file inside of ~/Documents/neorg.

_No response_

### Other information

_No response_

### Help

_No response_

### Implementation help

_No response_

Roger-Roger-debug:
Fixes #662 

Now checking for the absolute path to the file. I don't know if this is the right way to do it but it works for me.

### DIFF
--- a/lua/neorg/modules/core/norg/journal/module.lua
+++ b/lua/neorg/modules/core/norg/journal/module.lua
@@ -129,14 +129,16 @@ module.private = {
             time
         )
 
-        local journal_file_exists = false
-        if module.required["core.norg.dirman"].file_exists(folder_name .. neorg.configuration.pathsep .. path) then
-            journal_file_exists = true
-        end
+        local workspace = workspace or module.required["core.norg.dirman"].get_current_workspace()[1]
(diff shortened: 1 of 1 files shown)


vhyrro:
Thanks!

PR #711, opened 2023-01-12
fix: completion for TODO items

HendrikPetertje:
Added the "(" as a trigger character, prevent ( from spawning a ) with the updated regex and updated the completion table.

I'm not super familiar with the new syntax yet.. or the syntax in general; so I'll probably end up reading up on the documentation to see what other completions can be disregarded and what completions should be added. But I'll do that in a future pull request.

### DIFF
--- a/lua/neorg/modules/core/integrations/nvim-cmp/module.lua
+++ b/lua/neorg/modules/core/integrations/nvim-cmp/module.lua
@@ -86,7 +86,7 @@ module.public = {
         end
 
         function module.private.source:get_trigger_characters()
-            return { "@", "-", "[", " ", "." }
+            return { "@", "-", "[", "(", " ", "." }
         end
 
         module.private.cmp.register_source("neorg", module.private.source)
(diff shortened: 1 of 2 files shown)


lua/neorg/modules/core/integrations/nvim-cmp/module.lua
```diff
@@ -86,7 +86,7 @@ module.public = {
         end
 
         function module.private.source:get_trigger_characters()
-            return { "@", "-", "[", " ", "." }
+            return { "@", "-", "[", "(", " ", "." }
```
HendrikPetertje, line 89: Before I would recommend merging this:
Can i safely remove [ as a trigger character for completion start?

HendrikPetertje (commented):
thoughts from self

lua/neorg/modules/core/integrations/nvim-cmp/module.lua
vhyrro: yeah, there are no other constructs where `[` is required as a trigger character in norg. You can safely remove it imo :)

vhyrro:
Thanks a bunch for this!

PR #721, opened 2023-01-24
improvement: overhaul documentation generator

vhyrro:
This is the last "chore" PR before I can start ramping up work on features (calendar, GTD) for real this time!

The documentation in Neorg is okay I guess, but it's by no means anything fancy or useful. This PR aims to not only refactor the codebase into something much more manageable, but also to improve the look and feel of the module documentation found in the wiki, including GIFs, more accessible documentation, and fixed rendering of lua tables.


vhyrro:
Update: things are looking really good so far. Table rendering seems to work as intended!

vhyrro:
Here's a sample of what a module page will look like with the new update: <https://gist.github.com/vhyrro/b34e5ab60fd35060eda2facb40287c7b>

By no means complete, heavily subject to change, feedback appreciated! GIFs etc in the gist are there just to show what it would look like. I'm updating the gist as I make commits here so you can see the effects in realtime.

roland-rollo:
Looks very clean and informative. Me like.

vhyrro:
I'm currently stuck on a very difficult to fix bug, but after that's done we should be ready to merge!

vhyrro:
Aand would you know it I just fixed it. I'm happy to merge this, however there are still a few things that absolutely need doing. Just because the docgen is now readable and shiny does not mean that all of the documentation is now magically better, cause some of the lua comments describing the behaviours of modules are very lackluster to be fair. After this the focus is now to improve the actual documentation itself, because the generator is right where I want it to be :)

PR #734, opened 2023-02-04
docs: improve documentation across the board

vhyrro:
This PR aims to rewrite the documentation for almost all the Neorg modules to be more easily consumable when reading the wiki pages.

This PR may take anything from a few days to a week, we'll see :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 documentation:
 	nvim --headless -u docgen/minimal_init.vim -c "cd ./docgen" -c "source init.lua" -c 'qa'
 
+local-documentation:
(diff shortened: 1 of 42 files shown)


vhyrro:
About halfway done currently

vhyrro:
I'm happy with this as-is. The documentation is much more consumable and approachable now, showing the user what features are present, as opposed to what APIs are present (which they shouldn't care about anyway). Thanks to the new table rendering in the new docgen figuring out configuration options should be simple.

While I have given fancy GIFs to a few modules, this is not the final state. There will be a follow up PR at some point in the future which will deal with giving nice embeds anywhere it can :)

PR #741, opened 2023-02-09
fix(core.norg.journal): fixes #736 , now generates TOC correctly

### LINKED ISSUE #736: Unable to Create TOC in Journal
### Prerequisites

### Neovim Version

NVIM v0.8.2

### Neorg setup
```
require("neorg").setup{
    load = {
        ["core.defaults"] = {}, -- Loads default behaviour
        ["core.norg.journal"]  = {},
        ["core.norg.dirman"] = { -- Manages Neorg workspaces
            config = {
                default_workspace = "notes",
                workspaces = {
                    notes = "~/notes",
                },
            },
        },
    },
}
```

### Actual behavior

Error executing luv callback:
...azy/neorg/lua/neorg/modules/core/norg/journal/module.lua:231: Unable to generate TOC for directory 'journal' - ENOENT: no such file or directory: journal/
stack traceback:
        [C]: in function 'assert'
        ...azy/neorg/lua/neorg/modules/core/norg/journal/module.lua:231: in function <...azy/neorg/lua/neorg/modules/core/norg/journal/module.lua:230>

### Expected behavior

I expect a TOC to be generated

### Steps to reproduce

1. From your home directory, create a journal entry for today, add some text, save it.
    ```
    :Neorg journal today
    ```
1. In the same journal entry, generate a TOC.  The error should have been generated after the command is ran.
    ```
     :Neorg journal toc update
    ```

I've managed to reproduce this only with Neorg installed.

### Other information

I believe the error is occurring as its checking for the directory `journal` in my current directory.  

### Help

Yes

### Implementation help

_No response_

dark-ether:
the failure in generating the journal TOC came from assuming that the
  user's current workspace is the journal's workspace, which may not be
  true if a journal has a workspace set in the configuration. 
i tested with my config so it fixes the bug, the changes are small so hopefully no bugs were created

### DIFF
--- a/lua/neorg/modules/core/norg/journal/module.lua
+++ b/lua/neorg/modules/core/norg/journal/module.lua
@@ -200,6 +200,8 @@ module.private = {
         local workspace = module.config.public.workspace
             or module.required["core.norg.dirman"].get_current_workspace()[1]
         local index = neorg.modules.get_module_config("core.norg.dirman").index
+        local workspace_path = module.required["core.norg.dirman"].get_workspace(workspace)
+        local workspace_name_for_links = module.config.public.workspace or ""
         local folder_name = module.config.public.journal_folder
 
         -- Each entry is a table that contains tables like { yy, mm, dd, link, title }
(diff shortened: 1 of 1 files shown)


vhyrro:
Thanks for this! :)

PR #743, opened 2023-02-11
feat(dirman): add new `use_popup` option for `dirman`

sensasian:
- Allow users to use the vim's default input prompt in cmdline instead of popup prompt provided by neorg ui when creating a new note

### DIFF
--- a/lua/neorg/modules/core/norg/dirman/module.lua
+++ b/lua/neorg/modules/core/norg/dirman/module.lua
@@ -517,8 +517,7 @@ module.on_event = function(event)
 
     -- If the user has executed a keybind to create a new note then create a prompt
     if event.type == "core.keybinds.events.core.norg.dirman.new.note" then
-        local use_popup = module.config.public.use_popup
-        if use_popup then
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/norg/dirman/module.lua
```diff
@@ -513,18 +517,27 @@ module.on_event = function(event)
...
-            center_x = true,
-            center_y = true,
-        }, {
-            width = 25,
-            height = 1,
-            row = 10,
-            col = 0,
-        })
+        local use_popup = module.config.public.use_popup
+        if use_popup then
```
max397574, line 521: no need to crate a variable if it's only used once
just do `if module.config.public.use_popup then`

sensasian: Thanks for pointing it out! 👍🏼 I was using the variable to print and verify user config value is indeed working properly, then forgot to refactor.

vhyrro:
Thanks for this! :)

PR #746, opened 2023-02-12
feat(core.norg.esupports.metagen) merge template from user with default template

dark-ether:
For generating a workspace summary i was planning on checking every .norg file metadata. So i configured metadata to auto i then wanted  my template to have categories: [ ] and description: TODO ADD but have everything else unchanged. 

When i added a template it just substituted the default one, which for my use case is undesired and if we assume that there is a metadata format most use cases a person will just override a name instead of removing old ones and using unrelated new ones. 

So this probably is the desired behaviour but if needed we could just have a configuration flag for whether to replace or merge the template i just didn't do that as it seemed unnecessary. but if wanted i can modify this PR to add this feature.

### DIFF
--- a/lua/neorg/modules/core/norg/esupports/metagen/module.lua
+++ b/lua/neorg/modules/core/norg/esupports/metagen/module.lua
@@ -36,6 +36,8 @@ module.config.public = {
     delimiter = ": ",
 
     -- Custom template to use for generating content inside `@document.meta` tag
+    -- Fields are merged with default fields. To prevent a field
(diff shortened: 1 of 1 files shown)


vhyrro:
Oh my this PR completely slipped my radar and has been sitting here for months. I'm really sorry! It has been sitting here for so long in fact that it has been since fixed, you may specify your template with the format `{ "title" }` to keep the default generator for that value or you can override it with `{ "title", "My Title" }`.

Going to close this, feelsbadman :|

PR #749, opened 2023-02-14
fix: `:Neorg journal today` would fail on alternative path separators

TheLeoP:
Note: I'm using the stylua version compatible with lua 5.2 as stated in the CONTRIBUTING.md, but it is still changing the format of the document on parts of the code I didn't touch.

This should fix #725


vhyrro:
Hah it's fine, thanks for the fix!

PR #750, opened 2023-02-14
feat: add `dirman.summary` module

dark-ether:
it took me way too long to do this, really lost all my lua skills. there are various ways to improve but for a first version i think it is good. this code is probably undercommented, so if anything is unclear please ask.
possible improvements,
- allow user to configure which heading gets the summary
- allow user to configure which file gets the summary
- filter files by metadata
- move query logic to core.queries and make a new function there
- add a comment that gron. substitution is to allow files like xx.norg.norg
- expose some logic as lua functions
- make calling :Neorg generate-workspace-summary also call :w
- make the links relative to workspace root so they work for files in subdirectories
as this is around 100 lines i opted for implementing only generating the summary for now.

### DIFF
--- a/lua/neorg/modules/core/norg/dirman/module.lua
+++ b/lua/neorg/modules/core/norg/dirman/module.lua
@@ -472,6 +472,9 @@ module.public = {
         file:close()
         return true
     end,
+    get_index = function()
+        return module.config.public.index
+    end,
 }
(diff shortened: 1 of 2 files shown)


vhyrro:
This is great, thanks for taking the time to do this :)

PR #784, opened 2023-03-27
fix: neovim 0.9 vim.treesitter.parse_query deprecation

nagua:
Hello everyone,

neovim 0.9 deprecated the `vim.treesitter.parse_query` call a few days ago.
You should now use `vim.treesitter.query.parse` instead.

I introduced a new `neorg.utils.ts_parse_query` function that will check the
neovim version used and use the right call. I replaced all usage of `parse_query`
with the new function.

Have a great day everyone!

### DIFF
--- a/lua/neorg/external/helpers.lua
+++ b/lua/neorg/external/helpers.lua
@@ -9,7 +9,7 @@ neorg.utils = {
     --- @param query_string string # Query in s-expr syntax
     --- @return any # Parsed query
     ts_parse_query = function(language, query_string)
-        if vim.fn.has("nvim-0.9") == 1 then
+        if vim.treesitter.query.parse then
             return vim.treesitter.query.parse(language, query_string)
         else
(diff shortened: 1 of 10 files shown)


lua/neorg/external/helpers.lua
```diff
@@ -4,6 +4,17 @@
 --]]
 
 neorg.utils = {
+    --- A version agnostic way to call the neovim treesitter query parser
+    --- @param language string # Language to use for the query
+    --- @param query_string string # Query in s-expr syntax
+    --- @return any # Parsed query
+    ts_parse_query = function(language, query_string)
+        if vim.fn.has("nvim-0.9") == 1 then
```
max397574, line 12: checking like this will break things for people with an outdated nightly (like me)
you should do `if vim.treeistter.query.parse then`

vhyrro:
Thanks a lot for the PR! Fixed that bit that max was talking about and now we should be good :)

PR #788, opened 2023-04-02
ref(calendar): extract logic, fix bug with time reformatting

andreadev-it:
This fixes the issue that I've talked about recently (when the current date is higher than the previous/following month max day)
It also contains an improvement for the leap year check.


vhyrro:
thx! :heart:

PR #790, opened 2023-04-06
ref(calendar): extract rendering logic, add and improve keybinds

andreadev-it:
This commit completes the extraction of the rendering logic out of the `create_calendar` function. This allows us to re-render the view sections as needed without too much extra code. It also simplifies the keybindings logic, and allowed the easy integration of "j" and "k" keybindings, which goes to the previous and next week, respectively.
The new keybinding logic also fixes an issue with leap years. Previously, if the month of february was rendered with 29 days, and then we moved to a different year where february only had 28 days, it would break the logic and not render correctly.

### DIFF
--- a/lua/neorg/modules/core/ui/calendar.lua
+++ b/lua/neorg/modules/core/ui/calendar.lua
@@ -247,6 +247,154 @@ module.private = {
         end
     end,
 
+    render_month_array = function (ui_info, date, options)
+        -- Render the first weekday banner in the middle
+        local weekday_banner = module.private.render_weekday_banner(ui_info, 0, options.distance)
+        module.private.render_month_banner(ui_info, date, weekday_banner)
+        module.private.render_month(ui_info, date, date, weekday_banner)
+
+        local blockid = 1
+
+        while math.floor(ui_info.width / (26 + options.distance)) > blockid * 2 do
+            weekday_banner = module.private.render_weekday_banner(ui_info, blockid, options.distance)
+
+            local positive_target_date = reformat_time({
+                year = date.year,
+                month = date.month + blockid,
+                day = 1,
+            })
+
+            module.private.render_month_banner(ui_info, positive_target_date, weekday_banner)
+            module.private.render_month(ui_info, positive_target_date, date, weekday_banner)
+
+            weekday_banner = module.private.render_weekday_banner(ui_info, blockid * -1)
+
+            local negative_target_date = reformat_time({
+                year = date.year,
+                month = date.month - blockid,
+                day = 1,
+            })
+
+            module.private.render_month_banner(ui_info, negative_target_date, weekday_banner)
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/ui/calendar.lua
```diff
@@ -247,6 +247,154 @@ module.private = {
...
+            ui_info.buffer,
+            module.private.namespaces.logical,
+            extmark_id,
+            {}
+        )
+
+        vim.api.nvim_win_set_cursor(ui_info.window, { position[1] + 1, position[2] })
+    end,
+
+    render_view = function (ui_info, view, date, previous_date, options)
```
andreadev-it, line 353: This function is new. It checks what it needs to be updated in the view based on the difference between the current and previous date.

lua/neorg/modules/core/ui/calendar.lua
```diff
@@ -303,198 +451,57 @@ module.public = {
...
-                        next_extmark_id,
-                        {}
-                    )
-                    vim.api.nvim_win_set_cursor(window, { next[1] + 1, next[2] })
+                    local new_date = reformat_time({
+                        year = current_date.year,
+                        month = current_date.month,
+                        day = current_date.day + 1
+                    })
+                    module.private.render_view(ui_info, view, new_date, current_date, options)
```
andreadev-it, line 467: This is what fixes the bug for the leap years. Previously, we were trusting the extmarks to move through the days, and february could have had 29 extmarks even if the year wasn't a leap year. Now it relies more on the date, this way even though february could have 29 extmarks, it will never use the last one if it isn't a leap year.

lua/neorg/modules/core/ui/calendar.lua
```diff
@@ -303,198 +451,57 @@ module.public = {
...
+                        year = current_date.year,
+                        month = current_date.month,
+                        day = current_date.day - 1
+                    })
+                    module.private.render_view(ui_info, view, new_date, current_date, options)
+                    current_date = new_date
+                end,
+            })
+
+            vim.api.nvim_buf_set_keymap(buffer, "n", "j", "", {
```
max397574, line 484: I guess you should use `vim.keymap.set` here

lua/neorg/modules/core/ui/calendar.lua
```diff
@@ -303,198 +451,57 @@ module.public = {
...
+                        year = current_date.year,
+                        month = current_date.month,
+                        day = current_date.day + 7
+                    })
+                    module.private.render_view(ui_info, view, new_date, current_date, options)
+                    current_date = new_date
+                end,
+            })
+
+            vim.api.nvim_buf_set_keymap(buffer, "n", "k", "", {
```
max397574, line 496: here as well

max397574 (changes requested):

andreadev-it:
Good catch, those were the functions used before, but now the modern counterpart is better :)

PR #796, opened 2023-04-10
ref(calendar): refactored extmark creation, fixed single month view and added month limit

andreadev-it:
This PR is mainly aimed at fixing the calendar when the view is too small to show more than one month. Currently, this is breaking the code because other months extmarks are not removed and will be displayed on top of the visible ones.
I've added a function to clear the extmarks of the months that are outside of the current view, and this fixes the issue.

The visible months are now limited to 11 (the center one + 5 on the sides) because having more than 12 months will break a lot of the current code.

I've also slightly refactored the extmark creation functions to make them more DRY.

### DIFF
--- a/lua/neorg/modules/core/ui/calendar.lua
+++ b/lua/neorg/modules/core/ui/calendar.lua
@@ -386,7 +386,6 @@ module.private = {
         month_max = month_max > 12 and (month_max - 12) or month_max
 
         local clear_extmarks_for_month = function (month)
-            print("Clearing extmark for month " .. month)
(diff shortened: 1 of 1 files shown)


andreadev-it:
Wow, that was fast 😄 
Thank you vhyrro for your reviews and checks, I do like contributing to this feature. 

Have a nice day

vhyrro:
Thanks to you for helping out so much! I recently started using octo.nvim again so the speed at which I review smaller PRs has increased by 200% haha

PR #798, opened 2023-04-11
ref(calendar): mode integration (WIP)

andreadev-it:
This PR proposes a way to integrate various modes into the calendar.

There are two parts to this proposal. 
The first part is how to give the selected date / range back to the code that requested it. I propose an implementation that make use of callbacks. If a callback is passed to the `create_calendar` function as an option, that callback will be called when the current calendar mode decides to.
What do you think about this idea? Another alternative would be to use events, but for such a simple task, it seemed overkill. We could also have both, if needed, or a better system that I couldn't think of.

Here it is how to set a callback with the proposed code:
```lua
module.public.select_date({
    callback = function(selected_date)
        vim.pretty_print(selected_date)
    end
})
```

The second part is the mode integration.
I've currently created a "select_date" and a "select_range" mode. In this proposal, modes are functions that return a table with state variables and hooks that are called from the view.

For example, here it is the code for the "select_date" mode:

```lua
['select_date'] = function (callback)
    return {
        name = 'select_date',
        on_select = function (_, date)
            if callback then
                callback(date)
            end
            return true
        end
    }
end,
```

The `on_select` hook is called from the view whenever the `<cr>` key is pressed. In this case, it simply calls the callback (if present) and returns `true`, which means "close the calendar window", or "I'm done".

For the select range, I've also added a `get_day_highlight` hook, which allows the mode to customize how days are shown in the calendar. This lets the mode "disable" all dates that precede the first selected date of the range.

Let me know what you guys think about this. This is highly Work In Progress, and I've created this PR mostly to discuss the implementation details with everyone, so that we're all happy with the end result 🙂

### DIFF
--- a/lua/neorg/modules/core/ui/calendar.lua
+++ b/lua/neorg/modules/core/ui/calendar.lua
@@ -7,11 +7,11 @@ local function reformat_time(date)
     return os.date("*t", os.time(date))
 end
 
-module.setup = function ()
+module.setup = function()
     return {
         requires = {
-            "core.ui"
-        }
+            "core.ui",
+        },
     }
 end
 
@@ -38,7 +38,7 @@ module.private = {
 
     modes = {},
 
-    get_mode = function (name, callback)
+    get_mode = function(name, callback)
         if module.private.modes[name] ~= nil then
             local cur_mode = module.private.modes[name](callback)
             cur_mode.name = name
@@ -52,7 +52,6 @@ module.private = {
         logical = vim.api.nvim_create_namespace("neorg/calendar/logical"),
         decorational = vim.api.nvim_create_namespace("neorg/calendar/decorational"),
     },
-
 
     current_mode = {},
 
@@ -264,7 +263,7 @@ module.private = {
                 day_highlight = module.private.current_mode:get_day_highlight({
                     year = year,
                     month = month,
-                    day = day_of_month
+                    day = day_of_month,
                 }, day_highlight)
             end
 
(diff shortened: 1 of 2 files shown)


vhyrro:
I like the implementation! One thing I would point out is maybe we might want to move the modes out of `module.config.public` and create functions like `calendar.add_mode(name, data)` with the state being tracked in `module.private` instead? I doubt a user will ever go out of their way to define custom modes in the configuration because there'll be no other code to use that mode anyway.

If the user is enough of a power user to want to set their own custom modes, they'll probably use `neorg.modules.get_module("core.ui.calendar").add_mode(...)` either way :)

andreadev-it:
Hi @vhyrro ,
Thanks for your feedback! I agree and have adjusted the code, but it seems that I can't use the `module.load` function in the calendar module. I would have used it to add the default modes at loading time.
I noticed that the module is imported (rather then required) from the `core.ui` module. If I change it into a requirement for `core.ui` the load function gets called (other errors show up however).
It seems that the load function doesn't work on imports, but honestly I didn't even know that imports were a thing, so I do not know what would be the best way to initialize it here...

andreadev-it:
Hi @vhyrro ,
I currently edited my code converting the calendar from being an extension of the "core.ui" module into being its own seperate module. This has the following advantages:
- it allows to use the load function to add default modes
- it allows us to require other modules to be used in the calendar. I've been working on extracting the view into a separate module (it will be in the next PR as a proposal)

Let me know what you guys think about it

andreadev-it:
Perfect, yes I think it's ready to be merged now 👍

PR #805, opened 2023-04-19
feat: warn access to `core.norg` modules instead of breaking

pysan3:
@vhyrro I read your post on the Discord server and found out that all `core.norg.foo` modules are moved to `core.foo`.

I know you made a major version jump, but even with that,
Please do not break the plugin immediately.

Instead, warning the user should be way better.

Please considering accepting this PR.

Best


NTBBloodbath (approved):
Thank you so much! 💜

PR #814, opened 2023-04-22
feat: add `core.pivot` for toggling list types

vhyrro:
This PR adds a simple but handy module for toggling the type of a list from unordered to ordered via two keybinds, `toggle-list-type` and `invert-list-type`.

You'll see this make a presence in the next YT video!


max397574:
what's the difference between toggle and invert?

vhyrro:
check the documentation ;)

PR #815, opened 2023-04-23
ref(calendar): refactored view outside of calendar module

andreadev-it:
This PR is just a refactor to propose a way to separate the handling of different views and modes from the actual drawing. Views are extracted into their own modules (in this case, inside the folder `core.ui.calendar.views`) and can add themselves to the calendar module through the `public.add_view` method. This allows other plugins (like GTD, for example) to create their own views and add them to the calendar through an API.

Let me know if you guys like this, the API is pretty simple. The `add_view` function of the calendar module accepts two arguments: the view `name` and the view `details`. The details have to include a setup function that will be called with the following arguments: `ui_info`, `mode` and `options`.

This is how it looks to add the monthly view:

`core/ui/calendar/views/monthly.lua`:
```lua
module.public = {

    view_name = "monthly",

    setup = function (ui_info, mode, options)
        -- ...
    end
}

module.load = function ()
    module.required['core.ui.calendar'].add_view(module.public.view_name, module.public)
end
```

The new functions that were added are:
`calendar.public.add_view`
`calendar.private.get_view`
`monthly.public.setup`

It also slightly changes the api for the modes. The closing of the calendar window is now automatic whenever the callback that requested the date selection is called. This means that the calendar mode are not required to return `true` in order to close the calendar. They can now return `true` if they want the view to be redrawn (that's how the mode `on_select` return value should be considered after this change)

One thing to note: since the calendar views could be very different in how they work (based on the goal they're trying to achieve), I think that the drawing of the last line, that shows how to get the help menu and also shows the view name, should be handled by the view itself. That's of course my opinion, and I'll be happy to see what everyone else think about this whole PR 🙂

Have a nice day!


vhyrro:
Yes!! I really like this refactor, makes things much more readable.

You're right, it would be better to make the current view/help components a per-view thing instead of a global thing, I'm happy with such a change later down the line :)

One very small nitpick: be sure to run `make format` every now and again, or set up a git hook to do it for you - just for the sake of consistency :p

PR #816, opened 2023-04-24
fix(export): `gsub` export links that contain `#`, `?`. closes #807

### LINKED ISSUE #807: bug: Markdown `export` links that contains `#`, `?` etc
### Prerequisites

### Neovim Version

0.9.0

### Neorg setup

```lua
local M = {
  "nvim-neorg/neorg",
  ft = "norg",
  dependencies = {
    "nvim-treesitter/nvim-treesitter",
    "nvim-treesitter/nvim-treesitter-textobjects",
    "nvim-cmp",
  },
  build = ":Neorg sync-parsers",
  cmd = "Neorg",
  default_workspace = "Notes",
}

local function load_plugins()
  return {
    ["core.defaults"] = {},
    ["core.export"] = {},
    ...
  }
end

M.config = function()
  require("neorg").setup({
    load = load_plugins(),
  })
end

return M
```

### Actual behavior

When you have a `norg` file like this, where there is a `?` inside the link,

```txt
From No Org to Neorg - Workflow, Links | #2
- {https://www.youtube.com/watch?t=17m44s&v=Bi9JiW5nSig}
```

when you run `:Neorg export to-file foo.md`, the result will be,

```markdown
From No Org to Neorg - Workflow, Links | #2
- [https://www.youtube.com/watch?t=17m44s&v=Bi9JiW5nSig](https://www.youtube.com/watch%3ft=17m44s&v=Bi9JiW5nSig)
```

where **the `?` is converted to `%3f`.**

The same goes with `#` -> `%23`.

I know this is a result of URL encoding, but I believe these characters should remain the same and not be converted since they come up a lot in URLs.

### Expected behavior

- `#`, `?` in links not encoded on export.

### Steps to reproduce

```txt
From No Org to Neorg - Workflow, Links | #2
- {https://www.youtube.com/watch?t=17m44s&v=Bi9JiW5nSig}
```

Run command
`:Neorg export to-file foo.md`

_No response_

### Other information

_No response_

### Help

No

### Implementation help

_No response_

pysan3:
Hi, thanks for the nice plugin!

As I mentioned in #807, several characters in links inside exported markdown files are escaped and do not link to the desired position.

Several examples are...

- `{https://www.youtube.com/watch?v=j4lTvIGRhmw}["?" => "%3f"]`
  - ["?" => "%3f"](https://www.youtube.com/watch%3fv=j4lTvIGRhmw)
- `{https://github.com/nvim-neorg/neorg/#-installationquickstart}["#" => "%23"]`
  - ["#" => "%23"](https://github.com/nvim-neorg/neorg/%23-installationquickstart)

where if you follow the links above, they do not jump to the correct link, due to the conversion.

This PR aims to fix / workaround the output from `vim.uri_from_fname()` to fix this issue.

Hope this helps.
Best,

### DIFF
--- a/lua/neorg/modules/core/export/markdown/module.lua
+++ b/lua/neorg/modules/core/export/markdown/module.lua
@@ -118,15 +118,6 @@ module.load = function()
     module.config.public.extensions = neorg.lib.to_keys(module.config.public.extensions)
 end
 
-module.config.private = {
-    url_escapes = {
-        ["%%3f"] = "?",
-        ["%%23"] = "#",
-        ["%%5b"] = "[",
-        ["%%5d"] = "]",
-    },
-}
-
 module.config.public = {
     -- Any extensions you may want to use when exporting to markdown. By
     -- default no extensions are loaded (the exporter is commonmark compliant).
@@ -574,13 +565,6 @@ module.public = {
 
                 if state.is_url then
                     state.is_url = false
-                    local url = vim.uri_from_fname(output[#output - 1]):sub(string.len("file://") + 1)
-                    if url ~= output[#output - 1] then -- at least something has changed
-                        for key, value in pairs(module.config.private.url_escapes) do
-                            url = string.gsub(url, key, value)
-                        end
-                    end
-                    output[#output - 1] = url
                     return output
                 end
 
--- a/lua/neorg/modules/core/export/markdown/module.lua
+++ b/lua/neorg/modules/core/export/markdown/module.lua
@@ -118,6 +118,15 @@ module.load = function()
     module.config.public.extensions = neorg.lib.to_keys(module.config.public.extensions)
 end
 
+module.config.private = {
+    url_escapes = {
+        ["%%3f"] = "?",
+        ["%%23"] = "#",
+        ["%%5b"] = "[",
+        ["%%5d"] = "]",
+    },
+}
+
 module.config.public = {
     -- Any extensions you may want to use when exporting to markdown. By
     -- default no extensions are loaded (the exporter is commonmark compliant).
@@ -565,7 +574,13 @@ module.public = {
 
                 if state.is_url then
                     state.is_url = false
-                    output[#output - 1] = vim.uri_from_fname(output[#output - 1]):sub(string.len("file://") + 1)
+                    local url = vim.uri_from_fname(output[#output - 1]):sub(string.len("file://") + 1)
+                    if url ~= output[#output - 1] then -- at least something has changed
+                        for key, value in pairs(module.config.private.url_escapes) do
+                            url = string.gsub(url, key, value)
+                        end
+                    end
+                    output[#output - 1] = url
                     return output
                 end
 


d-r-a-b:
This seems like the wrong fix. URL encoding is meant to be done on URL components, and just searching for certain characters you don't want escaped and replacing them is very fragile.

What is the use case in the original code for calling `vim.uri_from_fname` here? If `link_target_url` is actually always a URL, it should already be well-formed and the proper fix is probably never to call `vim.uri_from_fname` at all. Was the intention to provide a "fix" for poorly written links of the form `{/absolute/path}` which should have been written `{/ /absolute/path}` according to norg-spec?

 To know if it is always a URL, probably will need to check the grammar of `link_target_url` which causes the `is_url` condition in the recollector. It would be defined somewhere in [nvim-neorg/tree-sitter-norg](https://github.com/nvim-neorg/tree-sitter-norg) probably in scanner.cc.

pysan3:
> To know if it is always a URL, probably will need to check the grammar of link_target_url which causes the is_url condition in the recollector.

I agree but this solution is way beyond the scope of this PR imo.

I know this PR is not a silver bullet and will have many edgecases, but I wanted to point out that the current situation (with no fix) is really annoying.

In the sake of `vim.uri_from_fname`, according to [Wikipedia](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), the symbols that have _meanings_ in uri are `:, /, ?, #, [, ], and @`.

The result of `vim.uri_from_fname(":/?#[]@") == "file://:/%3f%23%5b%5d@"` and therefore I created the `url_espaces` table.

d-r-a-b:
> I agree but this solution is way beyond the scope of this PR imo.

In what way is understanding where a bug is coming from outside the scope of a PR to fix that bug? If I'm correct about the behavior of `link_target_url ` the actual fix also doesn't require more code, it actually requires much less: just delete https://github.com/nvim-neorg/neorg/blob/336416f6c41777a4025cc80b8a085e21e758931f/lua/neorg/modules/core/export/markdown/module.lua#L568

d-r-a-b:
The current code couldn't be to support your example since 概要 is in the anchor (part after `#`) and dies with current export anyways. I'm guessing you are saying for theoretical support of something like [http://ko.wikipedia.org/wiki/위키백과](https://ko.wikipedia.org/wiki/위키백과) which has UTF in the page itself as opposed to as part of anchor or a query.

More generally, it doesn't make any sense to do an URL encoding step when moving from norg to markdown. If you are happy with `{https://qiita.com/honeytrap15/items/5a2d60444795bf92c1c0#概要}[link]` in your norg, then you should be happy with `[link](https://qiita.com/honeytrap15/items/5a2d60444795bf92c1c0#概要)` in your markdown.  Firefox has supported links with Unicode transparently since at least 2010. If you are running a browser that's over a decade old and can't support that style of link, you need to have `https://qiita.com/honeytrap15/items/5a2d60444795bf92c1c0#%E6%A6%82%E8%A6%81` in your norg to begin with anyways. The current code actually breaks norg export to markdown when you have properly encoded UTF URLs like `{https://qiita.com/honeytrap15/items/5a2d60444795bf92c1c0#%E6%A6%82%E8%A6%81
}[link]`, which then become `[link](https://qiita.com/honeytrap15/items/5a2d60444795bf92c1c0%23%25E6%25A6%2582%25E8%25A6%2581)` because it tries to reencode an encoded URL. (so all the % become %25). Your patch also would do nothing to address this.

pysan3:
> If you are running a browser that's over a decade old and can't support that style of link, you need to have {...} in your norg to begin with anyways.

I guess your right.

Fixed to simply do nothing at all in the latest commit!

Thanks for your opinions :)

d-r-a-b:
@pysan3, I think you may have pushed to the `main` branch of your fork instead of the PR branch, since I don't see it linked to this PR.

pysan3:
I am so sorry!!

Merged the fix from my main to this branch: a9d055907406daa606670c2c9c1fe8bfe98b7845.

And the final change after all should be a single line deletion.

Thanks for checking @d-r-a-b !

vhyrro:
Thanks a lot for the fix!

Issue #795, opened 2023-04-10
stacker overflow when increasing header level


champignoom:
### Prerequisites

- [X] I am using either 0.7 stable with the `0.0.12` tag or the latest compiled Neovim version
- [X] I am using the latest version of the plugin

### Neovim Version

both v0.9.0 & v0.10.0-dev-24+g3c724fe1f

### Neorg setup

```lua
require('neorg').setup {
	load = {
		["core.defaults"] = {}, -- Loads default behaviour
		["core.norg.concealer"] = {}, -- Adds pretty icons to your documents
		["core.norg.dirman"] = { -- Manages Neorg workspaces
			config = {
				workspaces = {
					notes = "~/notes",
				},
			},
		},
	},
}
```

### Actual behavior

```
Error executing Lua callback: ...packer/opt/neorg/lua/neorg/modules/core/promo/module.lua:127: Vim:Error executing lua callback: stack overflow
stack traceback:
        [C]: in function '__index'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:131: in function '_log'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:410: in function '_iter_regions'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:750: in function '_edit'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:810: in function 'fn'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:309: in function 'for_each_child'
        ...0/share/nvim/runtime/lua/vim/treesitter/languagetree.lua:809: in function '_on_bytes'
        ...r/neovim/0.9.0/share/nvim/runtime/lua/vim/treesitter.lua:65: in function <...r/neovim/0.9.0/share/nvim/runtime/lua/vim/treesitter.lua:64>
        [C]: in function 'nvim_buf_set_text'
        ...packer/opt/neorg/lua/neorg/modules/core/promo/module.lua:127: in function 'promote_or_demote'
        ...packer/opt/neorg/lua/neorg/modules/core/promo/module.lua:174: in function 'promote_or_demote'
        ...
        ...packer/opt/neorg/lua/neorg/modules/core/promo/module.lua:174: in function 'promote_or_demote'
        ...packer/opt/neorg/lua/neorg/modules/core/promo/module.lua:174: in function 'promote_or_demote'
[...]

Issue #803, opened 2023-04-18
[BUG]: promote and demote handle space incorrectly


sandersantema:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

9033d5eccc505d31b4a95d9fd23a7e2a8d2f282c (`--version` doesn't say much but this is the latest commit hash of the source from which I built nvim)

### Neorg setup

```lua
require('neorg').setup({
  load = {
    ['core.defaults'] = {},
    ['core.highlights'] = {
      config = { highlights = { links = { file = '+@text.uri' } } },
    },
    ['core.integrations.telescope'] = {},
    ['core.itero'] = {},
    ['core.norg.concealer'] = {},
    ['core.norg.dirman'] = {
      config = {
        workspaces = { notes = '~/Documents/Notes' },
        default_workspace = 'notes',
        index = 'index.norg',
      },
    },
    ['core.promo'] = {},
  },
})
```

### Actual behavior

Let `|` denote the cursor, `_` whitespace (explicitly to make the example more clear) and all other characters as usual execute `...` should be interpreted as a mapping i.e. execute `-<CR>` in insert-mode means type a `-` and then press enter.

* case 1:
`-_|` 
execute `<C-t>` in insert-mode
`--|` (result)
`--_|` (expected)

* case 2:
`--|` 
execute `<C-d>` in insert-mode
`- |` (result)
`-|` (expected)

* case 3:
`<empty buffer>`
execute `-<C-t>`
`-0>>|` (result)
`--|` (expected)

Similar behavior for `--<C-d>`

### Expected behavior

Described in actual behaviour.

### Steps to reproduce

Described in actual behaviour.

### Other information

It seems that case 3 is the result of the treesitter tree not being updated until leaving insert mode if you do the same and leave insert mode first before executing `<C-t>`, `0>>` is not inserted.

I  think case 1 & 2 are caused by some inconsistent capturing of whitespace around the `-`'s. I could also get some other weird behavior like the cursor ending up between the `-`'s but I couldn't reproduce reliably and I think solving the problems above should solve those weirder behaviors as well.

### Help

Yes, but I don't know how to start. I would need guidance (check question below)

### Implementation help

[...]

PR #817, opened 2023-04-25
refactor(promo): rewrite module to fix several bugs

champignoom:
Fixes https://github.com/nvim-neorg/neorg/issues/795 and https://github.com/nvim-neorg/neorg/issues/803 and https://github.com/nvim-neorg/neorg/issues/822

### DIFF
--- a/lua/neorg/modules/core/promo/module.lua
+++ b/lua/neorg/modules/core/promo/module.lua
@@ -144,7 +144,7 @@ module.public = {
         end
 
         apply_recursive(node, function(c)
-            if module.public.get_promotable_node_prefix(c) then
+            if module.public.get_promotable_node_prefix(c) == prefix then
                 return adjust_prefix(c)
             end
             return true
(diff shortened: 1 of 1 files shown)


vhyrro:
Very nice, thanks a lot for taking the time to do this! A future TODO would be to add descriptive error messages for when a possible assertion fails, so the user doesn't just get a traceback. Would you like to do that in this PR, or should we just make this a TODO for another day and merge now? I'm happy with either option tbh :)

champignoom:
Sure but the assertions are meant to help the developers rather than the users, so the error message might be only as readable as "internal error encountered, please create an issue" plus a table of all involved values.

pysan3:
FYI, this push is failing the luacheck test over on the main branch...
https://github.com/nvim-neorg/neorg/actions/runs/4832513182/jobs/8611381927

image

vhyrro:
Yep I'm aware, I'm fixing them up as we speak :)

Issue #833, opened 2023-05-03
undo leaves concealer artifact


champignoom:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

v0.9.0

### Neorg setup

```lua
require('neorg').setup {
	load = {
		["core.defaults"] = {}, -- Loads default behaviour
		["core.norg.concealer"] = {}, -- Adds pretty icons to your documents
		["core.norg.dirman"] = { -- Manages Neorg workspaces
			config = {
				workspaces = {
					notes = "~/notes",
				},
			},
		},
	},
}
```

### Actual behavior

```
◉
```
One header1 icon remains.

### Expected behavior

Buffer should become empty.

### Steps to reproduce

type `* a<Esc>u`

### Other information

Similar to https://github.com/nvim-neorg/neorg/issues/831,, this problem doesn't exist prior to https://github.com/nvim-neorg/neorg/commit/3009adf2cf48aedcbb309d0765e0fbbb64a0fdf4.

### Help

Yes

vhyrro:
This will be a very difficult problem to solve because `nvim_buf_attach` does not properly report undo operations, and this is a known issue without a real expected fix on the Neovim side iirc as it would require a large refactor in the codebase :|

Issue #831, opened 2023-05-03
incomplete concealing


champignoom:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

v0.9.0

### Neorg setup

```lua
require('neorg').setup {
	load = {
		["core.defaults"] = {}, -- Loads default behaviour
		["core.norg.concealer"] = {}, -- Adds pretty icons to your documents
		["core.norg.dirman"] = { -- Manages Neorg workspaces
			config = {
				workspaces = {
					notes = "~/notes",
				},
			},
		},
	},
}
```

### Actual behavior

```
◉ a
** b
```

### Expected behavior

```
◉ a
 ◎ b
```

### Steps to reproduce

type `* a<CR>** b<Esc>`

### Other information

This problem doesn't exist prior to https://github.com/nvim-neorg/neorg/commit/3009adf2cf48aedcbb309d0765e0fbbb64a0fdf4, so this might be a side effect of https://github.com/nvim-neorg/neorg/commit/ffabc4cc8bc36bbfbd96eb5d40e681d204627b02, potentially together with https://github.com/nvim-neorg/neorg/commit/45388fc0478e8d1273bd80789e7e1af1df76458.

### Help

Yes

Issue #821, opened 2023-04-27
concealer broken when promoting first element


champignoom:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

v0.9.0

### Neorg setup

```lua
require('neorg').setup {
	load = {
		["core.defaults"] = {}, -- Loads default behaviour
		["core.norg.concealer"] = {}, -- Adds pretty icons to your documents
		["core.norg.dirman"] = { -- Manages Neorg workspaces
			config = {
				workspaces = {
					notes = "~/notes",
				},
			},
		},
	},
}
```

### Actual behavior

With cursor on the first line with 'a', press '>>' to promote it, then the text becomes messed up:
```
** ◉
   1
***  ◎
    2
******    ✺
```
With the patch of the promo module from https://github.com/nvim-neorg/neorg/pull/817, the text will still be messed up but slightly differently:
```
*◉ a
   1
* ◎ b
    2
*   ✺ c
```
so this is more likely a concealer issue.

### Expected behavior

not messed up:
```
 ◎ a
   1
  ○ b
    2
     ⤷ c
```

Note that, if an empty line is inserted before the first line so that everything starts from the second line:
```

** a
   1
*** b
    2
****** c
```
promoting on the second line with 'a' won't reproduce the bug and everything behaves as expected.

### Steps to reproduce

Open test.norg, type
```
* a
  1
** b
   2
**** c
```
then ESC to enter normal mode, the text being concealed to
```
◉ a
  1
 ◎ b
   2
   ✺ c
```

### Other information

related: https://github.com/nvim-neorg/neorg/issues/292#issuecomment-1523194575

### Help

Yes

champignoom:
The reason seems to be that the "bufwinenter" event is artificially considered to be at position (1,0), which costs the concealing quota on the first line.

champignoom:
[A bigger problem might ](https://github.com/nvim-neorg/neorg/blob/3009adf2cf48aedcbb309d0765e0fbbb64a0fdf4/lua/neorg/modules/core/concealer/module.lua#L1552-L1554)

This is likely not achieving what it's supposed to achieve, because it's located under the `if event.type == "core.autocommands.events.bufwinenter"` branch, where `event.cursor_position` is always `{1,0}`.

Maybe it was meant to be `module.private.debounce_counters[start + 1]`?

vhyrro:
[...]

PR #834, opened 2023-05-03
refactoring concealer

champignoom:
I'm trying to solve the concealer problems (https://github.com/nvim-neorg/neorg/issues/821, https://github.com/nvim-neorg/neorg/issues/831, https://github.com/nvim-neorg/neorg/issues/833, etc.) systematically, but the current implementation is not very easy to deal with. Not few lines have large (up to 18) indentation levels.

https://github.com/nvim-neorg/neorg/blob/a295adf945a215d84e7a6252faf1b7a4dc9db8f7/lua/neorg/modules/core/concealer/module.lua#L386

This is a tracker PR of the refactoring of concealer and probably of the fixes as well.

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -1547,7 +1547,6 @@ module.on_event = function(event)
                 has_conceal = (
                     vim.api.nvim_win_is_valid(event.window)
                         and (vim.api.nvim_win_get_option(event.window, "conceallevel") > 0)
-                    or false
                 )
 
                 if mode ~= "i" then
@@ -1600,10 +1599,8 @@ module.on_event = function(event)
                         module.private.largest_change_end = _end
                     end
 
-                    module.private.largest_change_start = start < module.private.largest_change_start and start
-                        or module.private.largest_change_start
-                    module.private.largest_change_end = _end > module.private.largest_change_end and _end
-                        or module.private.largest_change_end
+                    module.private.largest_change_start = math.min(start, module.private.largest_change_start)
+                    module.private.largest_change_end = math.max(_end, module.private.largest_change_end)
                 end
             end,
         })


champignoom:
### Features
It's so far instantly responsive, consistent, and mostly backward compatible. With a more precise tracking of changes with phantom extmarks on each line, only changed parts that falls within the dozens of currently visible lines are rerendered, so there's no more performance tweaks.

### Bug fixes
- https://github.com/nvim-neorg/neorg/issues/821
- https://github.com/nvim-neorg/neorg/issues/831
- https://github.com/nvim-neorg/neorg/issues/833
- https://github.com/nvim-neorg/neorg/issues/839
- https://github.com/nvim-neorg/neorg/issues/844
- https://github.com/nvim-neorg/neorg/issues/846
- https://github.com/nvim-neorg/neorg/issues/848

### Breaking changes
- There's no more "enabled" in configs as it makes the wiki noisy. To disable part of a config, change e.g. `pending = { ... }` to `pending = false and { ... }`
- There's no more list_1..6 or heading_1..6. To specify the icons, just write e.g. `heading = { icons = { "!", "@", "#", "$" } }`, where "!" is the icon for level 1, "@" for level 2, etc. With this table, headings of level >6 can be specified. Same for ordered list, unordered list, and quote.
- There's no more "performance" section in config, as performance tweaks are dropped altogether.

### TODO
with decreasing urgency:
- [ ] BufRead is not triggered when editing new files. Another event is need. I'm not sure why/whether this hasn't been a problem.
- [ ] More testing.
- [ ] Comments and documentation.
- [ ] A new command for rerendering, ideally bound to ctrl-L.
- [ ] https://github.com/nvim-neorg/neorg/issues/739 extended marks for quotes.

max397574:
how does the performance compare to the old implementation?

vhyrro:
Oh superb I'll get to reviewing this once i get home! 

@max397574 champignoom says it's instantly responsive now, will test.

max397574:
just tested it
it feels much faster than previous implementation
also many of the redrawing issues (after deleting, pasting, yanking, moving, undoing etc) no longer occur
only thing
it looks like the icon preset no longer works for me

champignoom:
> how does the performance compare to the old implementation?

There should be no more lags like [here from the latest tutorial](https://youtu.be/InmnEt_vbx4?t=345).

For "large files", since only visible part is rendered, file size doesn't matter, so there should be no more lags apart from the one caused by treesitter itself. Besides, I still consider "large file optimization" kind of a myth until a reasonable set of benchmarks is proposed.

champignoom:
> it looks like the icon preset no longer works for me

That's indeed among the tops in the TODO above. This PR should only be mergeable when the first 4 are done.

champignoom:
> it looks like the icon preset no longer works for me

Fixed.

max397574:
I don't like the idea of having a default `<c-l>` mapping
this is such a common key to be mapped

champignoom:
> Fantastic work! The performance can really be felt - no stutters and no visual artifacts that I could find so far.
> 
> A few little things that I found while testing:
> 
> * [ ]  Empty lines in code blocks are not highlighted
> * [ ]  Code block highlight disappears when moving over them in insert mode. This could be considered intended, but I'd argue that the old behaviour of retaining the highlight is the least jarring :p
> * [ ]  Regular `{links}` are no longer concealed. I assume this is because you conceal the heading parts of `{* heading}`, and that causes issues when the content is concealed? EDIT: Apparently this is the case in `main` too? I either have dementia or I vividly recall it working normally before.
> 
> This plus the other todos and this is looking really good!

The first two problems are fixed.

For the third one, could you give a concrete example of such link and the expected concealing behavior?

aldanor:
> I don't like the idea of having a default <c-l> mapping
> this is such a common key to be mapped

Indeed. IIRC, `<c-l>` is mapped by default by most nvim 'distributions' (e.g. in lazyvim, it's bound to "switch to the window to the right").

max397574:
I haven't looked into how this works (and I won't)
but would it be possible to keep in mind that we got this now https://github.com/neovim/neovim/pull/20130
we probably can't use it yet since we can't expect people to be on latest nightly but perhaps make it easy to be able to use it later (@vhyrro )

champignoom:
This should be useable now.

Except that the configuration customization is still buggy and underdocumented. Something from the configuration system itself must be fixed first (e.g. https://github.com/nvim-neorg/neorg/issues/863), which is the job of another PR.

vhyrro:
Hope you don't mind, I just pushed an extra fix that does two things:
- Removes the necessity of placing two extmarks for code blocks
- Fixes a fun edge case bug with partially empty lines in code blocks, causing the concealer to just be confused.

I tested it pretty extensively and to my knowledge it doesn't introduce any bugs :sweat_smile:.

After this I think I'll dive in to fix luacheck screaming and after that I'm fine with merging this module as it is :), we can introduce extra documentation and fixes as we go.

vhyrro:
It's time to merge! @champignoom just wanted to say this is really impressive work, the concealer is something I myself wouldn't even want to touch with a metre long stick lol. Thank you :)

PR #836, opened 2023-05-03
fix(concealer): stop concealer if buffer is not loaded

d-r-a-b:
Many potential causes for buffer to be unloaded, but `:h vimgrep` is a big one. Without this fix, `:vimgrep` can get caught in an infinite loop of errors because the concealer wants to keep going until everything is done, but that can not happen if `:vimgrep` does not find a match and immediately deletes the buffer.


vhyrro:
Thanks! :D

PR #849, opened 2023-05-06
refactor: use new icons for `done`/`undone` as per nerd fonts 3.0

obreitwi:
The icons used to conceal 'done' and 'pending' tasks were [moved to new code points](https://github.com/ryanoasis/nerd-fonts/blob/master/changelog.md#breaking-2-material-design-icons-codepoints). This commit fixes them according to [this translation table](https://github.com/ryanoasis/nerd-fonts/issues/1059#issuecomment-1404891287).

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -838,13 +838,13 @@ module.config.public = {
 
             done = {
                 enabled = true,
-                icon = "",
+                icon = "󰄬",
                 query = "(todo_item_done) @icon",
             },
 
             pending = {
                 enabled = true,
-                icon = "",
+                icon = "󰥔",
                 query = "(todo_item_pending) @icon",
             },
 


vhyrro:
Oh dang, thanks for making us aware of this. I wonder - how long will it take before this change propagates to everyone? I'm not sure how fast the release cycle is for fonts in various package managers, moreso when you start talking about other operating systems like Mac or Windows. I'm afraid that merging this right away will cause some long term confusion, so perhaps we should wait a few days before the merge?

obreitwi:
Yeah, I would suspect the change will take a while to propagate. For example, the [font I use](https://archlinux.org/packages/community/any/ttf-dejavu-nerd/) was patched earlier only earlier this week.

Is there an FAQ or trouble shooting guide it could be added to? Or a compatibility setting that could be used to switch the used icon…
Because I imagine I won't be the only one that's greeted with blank rectangles instead of done/pending tasks.

Afaik, there is no way to check if a code point is present in the font currently used by the terminal, so it will require user action… :/

vhyrro:
I see quite a lot of other projects adapt to nerd fonts 3.0 so I guess it's our time to do so now too! Thanks for the PR :)

PR #850, opened 2023-05-06
docs: add Dependencies.md

d-r-a-b:
Per our discussion in the discord, this PR creates a Dependencies.md which describes the dependencies and some of the many pitfalls involved in getting the basic Neovim + Terminal things set up correctly.

It also (messily, perhaps), creates a subfolder `res/wiki/static` in which files can be placed and which will get copied into the wiki. I haven't had a chance to test that this works properly, so it definitely needs appropriate review before it gets merged.

### DIFF
--- a/docgen/docgen.lua
+++ b/docgen/docgen.lua
@@ -2,10 +2,24 @@ local docgen = {}
 
 -- Create the directory if it does not exist
 docgen.output_dir = "../wiki"
+docgen.static_dir = "../res/wiki/static"
 pcall(vim.fn.mkdir, docgen.output_dir)
 
 -- Copy static wiki resources into the wiki
-pcall(vim.fn.system, "cp -r ../res/wiki/static/* " .. docgen.output_dir)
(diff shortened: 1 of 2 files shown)


vhyrro:
Awesome stuff. One thing I would recommend is to use `vim.loop.fs_copyfile(old_path, new_path)` to make everything cross platform. Other than that I quite like the static idea for things that absolutely won't change.

d-r-a-b:
don't merge this yet, definitely messed something up

d-r-a-b:
Should be better?

vhyrro:
Just tested and it works on my end, so it should hopefully work on the CI VM too! Thanks for this :)

d-r-a-b:
Looks like it didn't kill the CI 😅 The Packer stuff is still basically empty since I definitely don't know anything about it really. IIRC you said you might be able to fill it in a bit -- sorry for forgetting to remind you when I submitted the PR! I just saw the section again when checking the Wiki to make sure it went in Ok

PR #858, opened 2023-05-08
ref(calendar): implemented distance (monthly view) and run make format

andreadev-it:
This very small PR presents a fix for the distance option that gets passed to the monthly view. It also improves a little on code readability in that function.


vhyrro:
Awesome :D

Issue #865, opened 2023-05-12
`Neorg update-metadata` does not use custom datetime format from `core.esupports.metagen` config


dasupradyumna:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.10.0-dev

### Neorg setup

``` lua
require('neorg').setup {
    load = {
      ['core.defaults'] = {},
      ['core.completion'] = { config = { engine = 'nvim-cmp' } },
      ['core.dirman'] = { config = { workspaces = { journal = '~/journal', rrc = '~/RRC' } } },
      ['core.esupports.metagen'] = {
        config = {
          tab = '  ',
          template = {
            { 'title', function() return vim.fn.expand '%:t:r' end },
            { 'description', '' },
            { 'author', 'dasu pradyumna' },
            { 'categories', '' },
            { 'created', function() return os.date '%d %b %Y %H:%M' end },
            { 'updated', function() return os.date '%d %b %Y %H:%M' end },
            -- { 'version', require('neorg.config').norg_version }, -- error here, why?
          },
          update_date = false, -- BUG: "update-metadata" uses default datetime format (not config)
        },
      },
    },
  }
```

### Actual behavior

`Neorg inject-metadata` respects the datetime format provided in config, and it works as expected. But `update-metadata` seems to be hardcoded to use the default datetime format.

After running `Neorg inject-metadata` -
![bug_neorg_metagen_datetime](https://github.com/nvim-neorg/neorg/assets/45595032/f17c5f70-81d6-4818-9346-5733424cca31)

After running `Neorg update-metadata` -
![bug_neorg_metagen_datetime2](https://github.com/nvim-neorg/neorg/assets/45595032/144ab0ea-ea55-49ac-b2e0-9cb35d7ea470)

### Expected behavior

`update-metadata` should behave like `inject-metadata` and respect custom datetime format in user config.

### Steps to reproduce

1. Open a `norg` file in NeoVim.
2. Execute `Neorg inject-metadata`.
3. Execute `Neorg update-metadata`.
4. **created** and **updated** times will have different formats.

### Potentially conflicting plugins

None

### Help

Yes, but I don't know how to start. I would need guidance (check question below)

### Implementation help

Ensure that `update-metadata` respects the datetime format from `core.esupports.metagen`'s `config.template[6]`.

katawful:
Please try #881

PR #882, opened 2023-05-18
fix(metagen): update generation to use user config for `updated` tag

katawful:
Reupload due to bad branch


dasupradyumna:
I'm the author of #865, I can confirm this PR works for me.
Thanks @katawful ~

vhyrro:
Thanks a lot for the fix!

PR #887, opened 2023-05-19
docs(README): add treesitter fix for macos

howarddo2208:
Fix the documentation for #74, required C++11 instead of C++14, took me quite a while in the issue section

PR #891, opened 2023-05-21
fix: TSInstall issues on macOS, hopefully once and for good

aldanor:
I've noticed #887 and figured maybe we can do a bit bitter.

Pointing people at installing external c++ compilers like gcc (especially via homebrew which they may not be using) in order to use neorg is not the most welcoming thing to put in a readme 🙂

(Also note that the "point CC at a compiler with C++11 support" part of the readme is not entirely incorrect, because native `clang` _does_ have C++11 support; it's just not the default.)

Here's a solution that I've tested myself on two different macOS boxes, does _**not**_ require any additional dependencies and should work with system-provided `clang`.

This did require digging a bit into treesitter installer internals (unfortunately) so it may potential break with future treesitter releases, however that part of TS didn't seem to have changed much at all, so it's not too likely.

### DIFF
--- a/lua/neorg/modules/core/integrations/treesitter/module.lua
+++ b/lua/neorg/modules/core/integrations/treesitter/module.lua
@@ -683,6 +683,45 @@ module.public = {
     end,
 }
 
+-- this fixes the problem of installing neorg ts parsers on macOS without resorting to using gcc
+local function install_norg_ts()
+  if vim.fn.has 'macunix' == 1 then
+    -- https://github.com/nvim-neorg/tree-sitter-norg/issues/7
+    -- (we have to force clang to c++11 mode on macOS manually)
+
+    local shell = require 'nvim-treesitter.shell_command_selectors'
+    local install = require 'nvim-treesitter.install'
+
+    -- save the original functions
+    local select_executable = shell.select_executable
+    local compilers = install.compilers
+
+    -- temporarily patch treesitter install logic
+    local cc = 'clang++ -std=c++11'
+    ---@diagnostic disable-next-line: duplicate-set-field
+    shell.select_executable = function(executables)
+      return vim.tbl_filter(function(c) ---@param c string
+        return c ~= vim.NIL and (vim.fn.executable(c) == 1 or c == cc)
+      end, executables)[1]
+    end
+    install.compilers = { cc }
+
+    -- install norg parsers
+    local ok, err = pcall(function()
+      install.commands.TSInstallSync['run!'] 'norg'
+    end)
+
+    -- no matter what, restore the defaults back
+    shell.select_executable = select_executable
+    install.compilers = compilers
+
+    -- if an error occurred during install, propagate it up
+    if not ok then error(err) end
+  else
+    vim.cmd [[ TSInstall! norg ]]
+  end
+end
+
 module.on_event = function(event)
     if event.split_type[1] == "core.keybinds" then
         if event.split_type[2] == "core.integrations.treesitter.next.heading" then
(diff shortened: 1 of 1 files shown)


LangdalP:
I cannot comment on the quality of this fix, but I can add that I am one of the users that got slightly annoyed that I could not install the neorg parser with the built in Clang, which does have C++11 support. I actually started using nvim-orgmode because of it, however I must admit that neorg looks attractive. My point is that this fix is very welcome from the user perspective!

vhyrro:
Ah, this is very nice! Thank you for taking the time to figure this out. I don't have any Mac machines to test for reference, so we'll see if the fix works for everyone once it propagates. If so, I can finally remove all the extra steps required to use Neorg on mac :D :purple_heart:

ddbrierton:
`:Neorg sync-parsers` does indeed work without error on macOS which is great. Thanks to all for the fix. However the fact that `:TSInstallSync norg` fails is kind of a problem, because every time there is a `nvim-treesitter` update (and there have been several in the last few days) it spews out error messages related to this, and any Mac user who is unfamiliar with the history of this problem is likely to think that something has become seriously broken with their `neorg` install.

molleweide:
I am using MacOS 13 on macbook air M1:

`Neorg sync-parsers` yields:

```vim
Downloading tree-sitter-norg...
Creating temporary directory
Extracting tree-sitter-norg...
Compiling...
Could not load parser for norg: "...0.8.2/share/nvim/runtime/lua/vim/treesitter/language.lua:26: no parser for 'norg' language, see :help treesitter-parsers"
Unable to install norg parser.
[nvim-treesitter] [2/3, failed: 1] Downloading tree-sitter-norg_meta...
...
```

ddbrierton:
> Could not load parser for norg: "...0.8.2/share/nvim/runtime/lua/vim/treesitter/language.lua:26: no parser for 'norg' language, see :help treesitter-parsers"

@molleweide have you tried upgrading neovim? You appear to be running 0.8.2. The latest release is 0.9.1.

molleweide:
> > Could not load parser for norg: "...0.8.2/share/nvim/runtime/lua/vim/treesitter/language.lua:26: no parser for 'norg' language, see :help treesitter-parsers"
> 
> @molleweide have you tried upgrading neovim? You appear to be running 0.8.2. The latest release is 0.9.1.

yup i tried with 0.9.1 now and it sais the same.

molleweide:
> I am using MacOS 13 on macbook air M1:
> 
> `Neorg sync-parsers` yields:
> 
> ```viml
> Downloading tree-sitter-norg...
> Creating temporary directory
> Extracting tree-sitter-norg...
> Compiling...
> Could not load parser for norg: "...0.8.2/share/nvim/runtime/lua/vim/treesitter/language.lua:26: no parser for 'norg' language, see :help treesitter-parsers"
> Unable to install norg parser.
> [nvim-treesitter] [2/3, failed: 1] Downloading tree-sitter-norg_meta...
> ...
> ```.
> ```

I got Norg to build for myself now:

1. I was using too old a nvim-treesitter version (probably the main issue)
2. I had old packer package leftovers, (including old nvim-treesitter version) under `XDG_DATA/nvim/site/packer/..` and `nvim-treesitter`s function for obtaining the parser install dir for some reason returned this old packer install of nvim-treesitter.

Now, after cleaning up all those leftovers, running `Neorg sync-parsers` worked perfectly.

aldanor:
> I think it's **important to pointing people at installing external c++ compilers like gcc in order to use neorg** is very important to put in the readme. Currently it's very vague and difficult to debug for beginners and people probably care 0 if their mac uses clang or gcc for compiling. 

@0xFlo Have you tried reading this PR's description? The whole point here was that you _don't need to install_ any external compilers unless your environment is already messed up in some peculiar way. To repeat, you don't need gcc for it to work.

PR #896, opened 2023-05-24
feat(todo-items): add missing "need input" icon and action

fecet:
The 1.0 spec https://github.com/nvim-neorg/norg-specs/blob/main/1.0-specification.norg including "need input" todo status:

https://github.com/nvim-neorg/norg-specs/blob/f4d696ebfe9c227fc912ad4249c6b34f71f4ace4/1.0-specification.norg#L573

But the current module do not include this status so I create this PR.

### DIFF
--- a/lua/neorg/modules/core/keybinds/keybinds.lua
+++ b/lua/neorg/modules/core/keybinds/keybinds.lua
@@ -46,6 +46,10 @@ module.config.public = {
                     -- ^mark Task as Important
                     { leader .. "ti", "core.qol.todo_items.todo.task_important" },
 
+                    -- Marks the task under the cursor as "need_input"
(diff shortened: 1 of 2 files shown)


vhyrro:
Thanks for the fix! What about `<LocalLeader>ta` for "ambiguous" or "agnostic" (which are both synonyms of `uncertain`) instead? `n` as a keybind for "*n*eeds input" doesn't make much sense imo, after that it's time to merge :)

fecet:
Thanks for your input! "ambiguous" is much more suitable, I just not find out other name at the moment :laughing:

vhyrro:
Awesome, thank you! :purple_heart:

vhyrro (approved):

PR #905, opened 2023-05-27
docs(README): make lazy.nvim install instructions more standard

TheBearodactyl:
Using ```config = function()``` is generally preferred over using ```opt```, so the section for lazy.nvim should be using 
```config = function()``` instead of ```opt```

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,24 +204,26 @@ You can install it through your favorite plugin manager:
 
   ```lua
   require("lazy").setup({
-      {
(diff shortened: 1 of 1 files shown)


vhyrro:
It does make things consistent with the other installation instructions, that's for sure. About `opt` vs `config`, to my knowledge they work exactly the same way with no regressions, so there's no real "good practice". Thanks anyway! :)

Issue #863, opened 2023-05-12
`icons` setting in `core.concealer` module does not override / combine with `icon_preset` setting


dasupradyumna:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.10.0-dev

### Neorg setup

``` lua
require('neorg').setup {
  load = {
    ['core.defaults'] = {},
    ['core.concealer'] = {
      config = {
        -- icon_preset = 'diamond',
        -- BUG: preset+override not working?
        icons = {
          heading = {
            level_1 = { icon = '❖' },
            level_2 = { icon = ' ◈' },
            level_3 = { icon = '  ◇' },
            level_4 = { icon = '   ⋄' },
            level_5 = { icon = '    ⟡' },
            level_6 = { enabled = false },
          },
        },
      },
    },
    ['core.dirman'] = { config = { workspaces = { journal = '~/journal' } } },
  },
}
```

### Actual behavior

I only wished to change the header icons while the `icon_preset` is set to "diamond". But if I set the `icon_preset`, the `icon` settings are ignored; and if i remove the `icon_preset`, the `icon` settings work but remaining icons belong to the "basic" preset.
Only the `level_6 = { enabled = false }` takes effect (heading 6 gets disabled), but other icons are still fetched from "basic".

Is this intended behavior?

### Expected behavior

I thought the setting the `icon_preset` and configuring icons would merge both settings, i.e getting "diamond" icons by default but only override the header icons.

### Steps to reproduce

Just open a `norg` file in NeoVim.

### Potentially conflicting plugins

None

### Help

None

PR #908, opened 2023-05-29
fix(config): fix config.custom

champignoom:
Fixes https://github.com/nvim-neorg/neorg/issues/863.


champignoom:
This is a dirty fix, as can be seen from `assert(not config)` which is based on the observation that the `config` of `neorg.modules.load_module(module_name, parent, config)` might not have been correctly used.

vhyrro:
@champignoom are we fine to undraft and merge this, or are there other things that need doing? Is the assertion a blocker?

champignoom:
@vhyrro It had diverged from main. Now it should be mergeable.

PR #926, opened 2023-06-02
feat(keybinds.lua): add `desc` fields to task keybinds

sardonicsloth:
this is to support proper desc= options for which-key, not clear if this is the right place to do this or not. I also don't like setting opts = defaults in the loop but i couldn't think of another way with out changing the rest of the code.

### DIFF
--- a/lua/neorg/modules/core/keybinds/module.lua
+++ b/lua/neorg/modules/core/keybinds/module.lua
@@ -316,18 +316,23 @@ module.public = {
                     for neovim_mode, keymaps in pairs(keys) do
                         -- Loop though all the keymaps in that mode
                         for _, keymap in ipairs(keymaps) do
+                            if keymap[3] and type(keymap[3]) == "table" then
+                                for extraopt, value in pairs(keymap[3]) do
+                                    opts[extraopt] = value
+                                end
+                            end
                             -- Map the keybind and keep track of it using the map() function
                             payload.map(
                                 mode,
                                 neovim_mode,
                                 keymap[1],
-                                "<cmd>Neorg keybind "
-                                    .. mode
-                                    .. " "
-                                    .. table.concat(vim.list_slice(keymap, 2), " ")
-                                    .. "<CR>",
+                                "<cmd>Neorg keybind " .. mode .. " " .. keymap[2] .. "<CR>",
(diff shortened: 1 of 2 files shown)


katawful:
The descriptions feel too ambiguous, hard to know what they are doing. I would suggest adding a verb

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -316,18 +316,23 @@ module.public = {
...
                                 keymap[1],
-                                "<cmd>Neorg keybind "
-                                    .. mode
-                                    .. " "
-                                    .. table.concat(vim.list_slice(keymap, 2), " ")
-                                    .. "<CR>",
+                                "<cmd>Neorg keybind " .. mode .. " " .. keymap[2] .. "<CR>",
                                 opts
                             )
+                            opts = {
```
katawful, line 332: Shouldn't this be above `payload.map()`?

sardonicsloth: This is a "reset" of the options to the default passed. The original function in keybinds.lua send a list of keybinds that is looped through and then a set of options { silent=true, noremap=true } but it doesn't allow for setting the descriptions per map. This is the best workaround i can think of with out changing everything about the function. Let me know if you want a better way

vhyrro:
Thanks for the PR! I think some extra work needs to be done here in order to make everything work with neorg, but these require some extra changes to the API. The fix to allow `{ desc = ...}` is good on paper, but Neorg reserves this as a "parameters" section for keybinds, where there may be any amount of parameters given. I think a better call would be to allow the table to receive a `opts = {}` field which overrides the opts of the mapping function. An example would be:
```lua
{ leader .. "tu", "core.qol.todo_items.todo.task_undone", opts = { desc = "Undone" } }
```
And then revert the change to the mapping function to use an `opts` field instead.

The descriptions could be more along the lines of `Mark Task as Done`, `Mark Task as Uncertain` and so on. If you don't beat me to it in about half an hour I can work on this a little bit :)

vhyrro:
Beat you to it :p

Thank you for the PR! I'll go on to add descriptions for the rest of the keybinds now :D

PR #928, opened 2023-06-03
fix(summary)!: fix norg links, use first heading as title if found

laher:
Fixes and improvements following on from https://github.com/nvim-neorg/neorg/pull/927 . I noticed a bug in the link destinations, so I went ahead and did some of what @vhyrro suggested.

- [x] Fix norg links ( `{:index.norg:}[My Index]` => `{:index:}[My Index]` ) - lol, funny I didn't notice this last time.
- [x] Links now use paths relative to the workspace root. `{:index:}` rather than `{:/Users/me/notes/work/index:}` . (It assumes you're in the workspace root, which _seems_ OK?)
- [x] When there's no `metadata.title`, it queries TS to find the first heading text. The TS query allows heading1 or heading2, and it checks matches up to 20 lines into the doc (e.g. if there's metadata but no title). IDK if this is right or if the query could be improved, but it seems like an OK starting point.
- [x] When there's no metadata or headings (e.g. an empty file), then the title becomes `Index` rather than `Index.norg`.
- [x] The summary now includes files which have no metadata (because we can present something useful now).
- [x] Breaking change - rename the strategy `metadata` to `default` because now it uses a mixture of metadata and headings/filenames now.

My summary looks like this now (as before, I only have scraps of document metadata, and some files are completely empty):

```norg
** My Heading Which My Cursor Was On
*** Homebase
   - {:index:}[Work Notes]
*** Uncategorised
   - {:help:}[Help]
   - {:inbox:}[Inbox]
   - {:main:}[Neorg Notes]
   - {:neorg:}[Neorg Adventures]
   - {:nostuff:}[Nostuff]
   - {:projects:}[Projects - Medium Term]
   - {:journal/index:}[2023]
   - {:journal/template:}[Template]
   - {:journal/2023/05/04:}[04]
   - {:journal/2023/05/05:}[05 Tues]
   - {:journal/2023/05/06:}[6]
   - {:journal/2023/05/07:}[7]
   - {:journal/2023/05/08:}[8]
   - {:journal/2023/05/18:}[18]
```

### DIFF
--- a/lua/neorg/modules/core/summary/module.lua
+++ b/lua/neorg/modules/core/summary/module.lua
@@ -61,14 +61,16 @@ module.load = function()
                     heading_query = neorg.utils.ts_parse_query("norg", heading_query_string)
                 end
                 -- search up to 20 lines (a doc could potentially have metadata without metadata.title)
-                for _, heading in heading_query:iter_captures(document_root, bufnr, 1, 20) do
-                    local start_line, _ = heading:start()
-                    local lines = vim.api.nvim_buf_get_lines(bufnr, start_line, start_line + 1, false)
-                    if #lines > 0 then
-                        local title = lines[1]:gsub("^%s*%*+%s*", "") -- strip out '*' prefix (handle '* title', ' **title', etc)
-                        if title ~= "" then -- exclude an empty heading like `*` (although the query should have excluded)
-                            return title
-                        end
+                local _, heading = heading_query:iter_captures(document_root, bufnr)()
+                if not heading then
+                  return nil
+                end
+                local start_line, _ = heading:start()
+                local lines = vim.api.nvim_buf_get_lines(bufnr, start_line, start_line + 1, false)
+                if #lines > 0 then
+                    local title = lines[1]:gsub("^%s*%*+%s*", "") -- strip out '*' prefix (handle '* title', ' **title', etc)
+                    if title ~= "" then -- exclude an empty heading like `*` (although the query should have excluded)
+                        return title
                     end
                 end
             end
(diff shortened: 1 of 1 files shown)


vhyrro:
Hell yeah. About the workspace root - you can opt for the `{:$/index:}` syntax, which *guarantees* that the file exists from the current workspace's root.
It's an extra character but it'll ensure that in case you create a summary from a file that isn't `index.norg` that everything will still point to the right place!

vhyrro:
Sorry if you see a million "starting review" notifications i'm playing around with how octo.nvim does things :p

lua/neorg/modules/core/summary/module.lua
```diff
@@ -40,16 +40,54 @@ module.load = function()
...
+                             )
+                             (heading2
+                                 title: (paragraph_segment) @next-segment
+                             )
+                         ]
+                     ]]
+                    heading_query = neorg.utils.ts_parse_query("norg", heading_query_string)
+                end
+                -- search up to 20 lines (a doc could potentially have metadata without metadata.title)
+                for _, heading in heading_query:iter_captures(document_root, bufnr, 1, 20) do
```
vhyrro, line 64: Check between lines `1` and `20` is alright, but I think we might be better off altogether by just checking for the first occurrence in the file? I can imagine some niche scenarios where you have a long introductory paragraph to a document and only then do you have the title in a first level heading >20 lines down.

If you agree this is the way to go then the simplest way is to ditch the `for` loop altogether and invoke the function manually to get the first result like this:

```suggestion
				local _, heading = heading_query:iter_captures(document_root, bufnr)()
                
                if not heading then
                	-- do something
                end
                
                -- continue using the `heading` node normally
```

vhyrro:
Resolved a conflict w.r.t the latest `main` commit through the github UI :+1:

lua/neorg/modules/core/summary/module.lua
laher: done - thanks for the tip, I didn't know about using `(..)()` to iterate over the first item only.

laher:
> Hell yeah. About the workspace root - you can opt for the `{:$/index:}` syntax

Done, ta

vhyrro:
Just tested and all seems to work great. Thanks a lot for the contributions!

PR #945, opened 2023-06-14
fix(docgen): inline `esupports.metagen` template function definitions

jneidel:
So they show up in the wiki.

## Current state
Here in the [wiki](https://github.com/nvim-neorg/neorg/wiki/Metagen) for the default implementation in the template it only show the `function()` part:
screenshot_23-06-14_18:50:40
That it is only saying `function()` is just confusing.

## Desired state
I would like to see the full function definition, so I can copy over the ones that I like and adjust what I don't like in my config:
screenshot_23-06-14_19:03:18

### DIFF
--- a/lua/neorg/modules/core/esupports/metagen/module.lua
+++ b/lua/neorg/modules/core/esupports/metagen/module.lua
@@ -41,16 +41,17 @@ module.config.public = {
         -- The title field generates a title for the file based on the filename.
         {
             "title",
-            function()
-                return vim.fn.expand("%:p:t:r")
-            end,
+            function() return vim.fn.expand("%:p:t:r") end,
         },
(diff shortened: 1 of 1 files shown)


jneidel:
@vhyrro You're doing a great job with the youtube videos! Without them the project would be significantly harder to get into.

max397574:
this is nice ig
but I think this is also something that should be fixed in the docgen

jneidel:
> I think this is also something that should be fixed in the docgen

That would probably be a better solution in the long-term :smile:

vhyrro:
Thanks for the temporary fix! Functions with an immediate `return` will be resolved in the docgen soon enough, I added it to my TODO list :)

Issue #941, opened 2023-06-10
neorg calendar: Make calendar work with smaller resolutions


TJ-Adams:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.10.0-dev-2980+gb8a2220f5-Homebrew Build type: Release LuaJIT 2.1.0-beta3     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/opt/homebrew/Cellar/neovim/HEAD-b8a2220/share/nvim"  Run :checkhealth for more info

### Neorg setup

neorg.setup(
    {
        load = {
            ["core.defaults"] = {},
            ["core.concealer"] = {},
            ["core.integrations.telescope"] = {},
            ["core.manoeuvre"] = {},
            ["core.journal"] = {},
            ["core.ui.calendar"] = {},

            ["core.export"] = {
                config = {
                    export_dir = "/tmp/neorg_export/"
                }
            },
            ["core.completion"] = {
                config = {
                    engine = "nvim-cmp",
                }
            },
            ["core.dirman"] = {
                config = {
                    workspaces = {
                        work = "~/.local/notes/work",
                        personal = "~/.local/notes/personal",
                    }
                }
            },
            ["core.keybinds"] = {
                    config = {
                            hook = function(keybinds)
                                    -- Keybinds to make moving sections up and down easily
                                    keybinds.map("norg", "n", "gj", "<cmd>Neorg keybind norg core.manoeuvre.item_down<cr>")
                                    keybinds.map("norg", "n", "gk", "<cmd>Neorg keybind norg core.manoeuvre.item_up<cr>")
                                    keybinds.map("norg", "n", "]s", "<cmd>Neorg keybind norg core.integrations.treesitter.next.heading<cr>")
                                    keybinds.map("norg", "n", "[s", "<cmd>Neorg keybind norg core.integrations.treesitter.previous.heading<cr>")

                                    -- Keybinds for the journal module
                                    keybinds.map("norg", "n", "<leader>njc", '<cmd>Neorg journal custom<cr>')
                                    keybinds.map("norg", "n", "<leader>njf", '<cmd>Neorg journal tomorrow<cr>')
                                    keybinds.map("norg", "n", "<leader>njy", '<cmd>Neorg journal yesterday<cr>')

                                    -- Keybinds for the calendar module
[...]

PR #950, opened 2023-06-16
fix(calendar): give calendar enough space to render

TJ-Adams:
This commit changes the logic for creating the calendar window to always request at least the minimum size necessary to render the full calendar. Previously, a window size could be created even though it was too small to render the calendar.


TJ-Adams:
This is in response to this [issue](https://github.com/nvim-neorg/neorg/issues/941). Please feel free to give feedback or request for changes. I'll try to get to them this weekend if it should be done differently!

vhyrro:
Thank you! :D

PR #958, opened 2023-06-21
feat: add blank lines between tangled blocks

### LINKED ISSUE #955: insert blank lines when tangling
### Issues

### Feature description

By default, when emacs tangles org-mode files, it inserts blank lines between code which originates in different blocks. This behavior makes for more readable output files---not the biggest deal with literate programming, but definitely useful. It's also probably good for the defaults to line up unless there's a reason for them not to.

Thanks for the great software--I'm currently trying to convert my ~2000 line config to norg so I'll probably have more issues to come. Happy to write a PR for this, doesn't seem too difficult.

### Help

Yes

### Implementation help

_No response_

### LINKED ISSUE #957: tangle entire directory of files
### Issues

### Feature description

I would like to run tangles for every .norg file in a directory(namely my project root directory).
I see that there was an event for this that is commented out in the code ":Neorg tangle directory"
I was wondering if there was more work done on this as I would love to just have my whole project which is written in .norg files be updated all with one command.

### Help

No

### Implementation help

_No response_

rileyshahar:
Resolves #955.

### DIFF
--- a/lua/neorg/modules/core/tangle/module.lua
+++ b/lua/neorg/modules/core/tangle/module.lua
@@ -276,7 +276,13 @@ module.public = {
                     end
 
                     if file_to_tangle_to then
-                        tangles[file_to_tangle_to] = tangles[file_to_tangle_to] or {}
+                        if tangles[file_to_tangle_to] then
+                            -- insert a blank line between blocks
+                            table.insert(content, 1, "")
+                        else
+                            tangles[file_to_tangle_to] = {}
+                        end
+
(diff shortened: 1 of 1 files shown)


nihilistkitten:
Sorry for the force push; I copied the wrong issue number.

nihilistkitten:
Happy to close in favor of #981, if that gets merged. Will leave it open in case we prefer the lower-complexity solution.

vhyrro:
As we discussed in the other thread, at some point we would also like this to be configurable. In the meantime though this is perfectly fine, thank you! :purple_heart:

PR #975, opened 2023-07-05
fix(ui): remove possible ui noise caused by user's opts

RayJameson:
Hello everyone! Thanks for great plugin. I tried to fix this in config by using autocmds, but couldn't make it work, so I decided to make this PR, I believe this is sensible default for UI window

**Before:**
image

**After:**
image


NTBBloodbath (approved):
Good catch, thank you so much!

PR #981, opened 2023-07-06
feat: delimit tangled code blocks with headings

jonathf:
An extention to the idea discussed in #955 .

Add heading text as delimiter as a comment if:

1. There is in-fact a current heading. Headings are optional.
2. Current heading hasn't been used as a delimiter before.
3. `commentstring` is available for the filetype in question.

Otherwise just use a newline.

That last one (3) requires making a temporary scratch buffer to trigger ftplugin, which retrieves `commentstring` for all supported filetypes in vim.

### DIFF
--- a/lua/neorg/modules/core/tangle/module.lua
+++ b/lua/neorg/modules/core/tangle/module.lua
@@ -245,6 +245,8 @@ module.public = {
         })
 
         local query = utils.ts_parse_query("norg", query_str)
+        local previous_headings = {}
+        local commentstrings = {}
 
         for id, node in query:iter_captures(document_root, buffer, 0, -1) do
             local capture = query.captures[id]
@@ -253,7 +255,8 @@ module.public = {
                 local parsed_tag = module.required["core.integrations.treesitter"].get_tag_info(node)
 
                 if parsed_tag then
-                    local file_to_tangle_to = options.languages[parsed_tag.parameters[1]]
+                    local language = parsed_tag.parameters[1]
+                    local file_to_tangle_to = options.languages[language]
                     local content = parsed_tag.content
 
                     if parsed_tag.parameters[1] == "norg" then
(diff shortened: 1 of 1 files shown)


nihilistkitten:
It would be nice if this supported immediately nested headings, i.e.
```norg
* Foo
** Bar
    @code py
    print("hi")
    @end
```
should output comments with both foo and bar, which I don't think the current impl does.

Also, I think we want to insert the heading even if nothing's been tangled to the current file, whereas we only want to insert a newline when something has been tangled to the current file. If I'm reading it right, the current PR doesn't do that.

jonathf:
Heading to the beginning of the file is trivial and has been added now.

Immediat nested heading is doable, but I think I want the reviewers input before making the PR more complex than it already is.

vhyrro:
I like the idea of this! But yeah this should definitely be put behind some feature flag as to not confuse users. It could be interesting as a fancy default, but there also should be a way to turn it off if wanted. I guess we want to read data from the document metadata where we could have some info regarding this. Maybe `tangle-delimiter: none|newline|headings` or something of the like? :D

vhyrro:
Additionally I see that there are conflicts here. When I am again free to work solely on Neorg I'll gladly help out with this PR. It might take some time for me to revisit this though! Not ignoring the PR, just busy with priorities :)

rileyshahar:
@vhyrro any chance we can merge the extremely simple #958 in the interim as a minimal impl of this?

vhyrro:
sounds good to me!

jonathf:
@vhyrro, I am glad you like the idea!

I've done as you suggested and added a configuration option, and moved the branch to the tip of main.

In addition I've done two more things:
1. Added content to the filetype detection. This mostly means that shebangs know are included in the filetype detection.
2. Added a generic `_` filetype that allows for catch-all file output when also setting other parameters. E.g. `tangle: output` is equivalent to `tangle: {languages: {_: output}}`.

I appologies if this is considered feature creep, but I feel it adds to the completness of the feature I am proposing.

lua/neorg/modules/core/tangle/module.lua
```diff
@@ -181,38 +191,38 @@ module.load = function()
...
-                    options.languages[vim.filetype.match({ filename = file })] = file
-                end
-            elseif parsed_document_metadata.tangle.languages then
-                for language, file in pairs(parsed_document_metadata.tangle.languages) do
-                    options.languages[language] = file
-                end
-            end
-        elseif type(parsed_document_metadata.tangle) == "string" then
-            options.languages[vim.filetype.match({ filename = parsed_document_metadata.tangle })] =
-                parsed_document_metadata.tangle
```
vhyrro, line 213: I realized that one feature stopped working as a result of this being removed - you should also be able to specify e.g. `tangle: ./init.lua` in the metadata, which now errors with the new version. It always assumes that `languages` is a thing, which it may not be :p

jonathf: Good catch. I have a handler for it, on line 220, but it was implemented incorrectly. Fixed now.

vhyrro:
Tried stress testing it and it seems to work with the cases I threw at it, thank you for this PR! :purple_heart:

PR #982, opened 2023-07-07
fix: replace `get_filetype` with `vim.filetype.match`

jonathf:
`utils.get_filetype` doesn't seem to work in autocmd like `BufWritePre` and `BufWritePost`, which is a real problem if you like to set up auto-tangle.

Fortunatly, neovim 0.8 introduce `vim.filetype.match` that does the same natively without issues.

Since the complexity of `get_filetype` has become trivial, I've deprecated the function.

### DIFF
--- a/lua/neorg/external/helpers.lua
+++ b/lua/neorg/external/helpers.lua
@@ -142,22 +142,6 @@ neorg.utils = {
 
         return ret
     end,
-    get_filetype = function(file, force_filetype)
-        local filetype = force_filetype
-
-        -- Getting a filetype properly is... difficult
-        -- This is why we leverage Neovim instead.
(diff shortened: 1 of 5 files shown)


vhyrro:
Awesome, thank you! Could you please run stylua on the code to keep things consistent? There's a makefile providing a `make format` for this :D

Be sure to have stylua installed with lua5.2 features: `cargo install stylua --features lua52`

jonathf:
Of course. 

Looks like the linter is complaining now. I ran luacheck locally, I don't have the same issue.  In fact, I don't think the error it is seeing there is really there. Might this be a caching issue perhaps?

vhyrro:
I'll take a look at the linter later on, it could just be confused haha. Thank you again!

PR #1001, opened 2023-07-30
refactor(codebase)!: make the `neorg` object local to a `core` module

Groctel:
This is the first PR aimed to address the refactor discussed at #999. The first step in refactoring the code base is moving the `neorg` object to the `neorg.core` module, which other modules can `require` from. This one **is** a big PR, since it touches most if not all files in the project. However, it is necessary to remove the sequencing necessity for loading the `neorg` object when it is defined as a global variable.

Please take your time to review that everything in this first change set is correct. After this, I will address the events system as I had already done on January to separate it from the module loader. I have left this task outside of this PR temporarily to strictly focus on the relocation of the `neorg` object.


vhyrro:
You're a beast for this, thank you. I did an all round check of everything (took me a while lol) but there's probably some stuff I might've missed. I hope the community will tell us soon enough whether anything's broken or not haha

After all of your PRs are done I think it's prime time to implement test suites :D

PR #1010, opened 2023-08-04
fix(concealer,indent): "require'neorg'" missing in v:lua call

taeruh:
I noticed that the indentation wasn't anymore after 5706f1efdcf55f273de8f52deeb35375a303be72. The problem is that the `v:lua` call for the `indentexpr` tries to access `modules` which is not global. It has to be accessed through `v:lua.require'neorg'`. I grepped for other `v:lua` calls and found the same problem for the `foldtext` option.


vhyrro:
Thank you! :D

PR #1014, opened 2023-08-06
feat: delimit tangle code blocks with file content

jonathf:
Same as with heading, but uses the entire Neorg file as a delimiter. In other words, the output file is identical as the source, but "inverted" in that blocks are the content and the rest of the document are comments.

Code is configure to work as expected with mulitple targets.

For estethical reason I chose to keep empty lines empty.

### DIFF
--- a/.luarc.json
+++ b/.luarc.json
@@ -4,5 +4,6 @@
         "_neorgcmd_generate_completions",
         "_neorg_module_autocommand_triggered",
         "vim"
-    ]
+    ],
+    "workspace.checkThirdParty": false
 }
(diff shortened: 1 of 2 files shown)


vhyrro:
This is a pretty fire feature! I like it a lot. Could you switch the name of `filecontent` to `file-content` instead to keep the naming scheme consistent across Neorg? I would do it myself but I probably messed something up with git because I can't push to this PR lol.

vhyrro:
Thank you! Tangling is a feature I haven't had time to extend so I'm really happy someone else is taking the wheel :D

PR #1048, opened 2023-09-04
feat(concealer): set foldlevel to 99 when it's zero

champignoom:
It is otherwise [confusing](https://discord.com/channels/834325286664929280/834326555160608769/1148243431345369159) to have everything folded by default.

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -633,8 +633,8 @@ module.config.public = {
     -- - "varied" - use a mix of round and diamond shapes for headings; no cute flower icons though :(
     icon_preset = "basic",
 
-    -- If true, Neorg will enable folding by default for `.norg` documents.
-    -- You may use the inbuilt Neovim folding options like `foldnestmax`,
+    -- If true, Neorg will enable folding and set `foldlevel` to 99 by default for `.norg` documents.
+    -- You may use the inbuilt Neovim folding options like `foldlevel`, `foldnestmax`,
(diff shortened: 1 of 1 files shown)


max397574:
I disagree with that change. I don't think user options should be changed when you use default options. It's wouldn't be a big deal imo if this was opt-in.
But in the end it's up to @vhyrro

champignoom:
I partially agree but `foldmethod` and `foldexpr` are changed by default already, and `foldlevel=0` is an unfortunate default from vim imo in the neorg context.

champignoom:
> this isn't directly related to the foldlevel

It is related here because `foldlevel` is taking observable effect only because `foldmethod` is set to `expr`.

One way to justify this change is to interpret the default `foldlevel=0` as calling for a context dependent default that is [least astonishing](https://en.wikipedia.org/wiki/Principle_of_least_astonishment.

max397574:
> It is related here since foldlevel is taking observable effect only because foldmethod is set to expr.

no
you could also e.g. set it to indent

max397574:
I have to disagree
reading a untouched default is completely different from reading an uninitialized variable. (I know what you mean with this example but I disagree with it)

I think this is quite a difficult thing to do since I really don't like changing setting when the user easily can do it, it's not required, and not directly related to the plugin setting (enable TS folding in this case)

in this case it's debatable though because this is such a bad default
as I said before I think we should also leave this up to @vhyrro

champignoom:
Closed in favor of https://github.com/nvim-neorg/neorg/pull/1049

PR #1049, opened 2023-09-04
feat(concealer): add option for opening all folds by default

champignoom:
It is otherwise [confusing](https://discord.com/channels/834325286664929280/834326555160608769/1148243431345369159) to have everything folded by default.


champignoom:
wait `:e` shouldn't reopen all folds but I don't know how to prevent that

champignoom:
never mind `:e` does open all folds when `foldlevel=99`

PR #1056, opened 2023-09-06
fix(concealer): more precise anticonceal feature detection

champignoom:
The previous version comparison logic seems flawed: (0,10,0) should be greater than (0,7,1), not less.

To ensure that anticonceal is available, it's [not enough](https://github.com/nvim-neorg/neorg/issues/292#issuecomment-1646543885) to only check version being 0.10.0. We must also ensure that the subversion (the number after '-') is greater than 575 or something iirc.

### DIFF
--- a/lua/neorg/core/utils.lua
+++ b/lua/neorg/core/utils.lua
@@ -104,7 +104,10 @@ end
 ---@param patch number #The patch number (in case you need it)
 ---@return boolean #Whether Neovim is running at the same or a higher version than the one given
 function utils.is_minimum_version(major, minor, patch)
-    return major <= version.major and minor <= version.minor and patch <= version.patch
+    if major ~= version.major then return major < version.major end
+    if minor ~= version.minor then return minor < version.minor end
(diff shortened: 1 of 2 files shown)


vhyrro:
I feel like this could be done much cleaner with the modern `vim.version()` and `vim.version.parse()` APIs, would be much more performant too :D 
It allows for checking if it's a development (nightly) build alongside a specific revision where we could test for anticonceal specifically. Thoughts?

champignoom:
That would be great but I in the result of `vim.version()` I can't find the `1031` part of `NVIM v0.10.0-dev-1031+gc431d820e`.

max397574:
it has the `__lt` field in the metatable
I'll take a look at the source to see if we can use that

vhyrro:
That kinda sucks - given that we check for the current nvim version several times in the source code altogether I would prefer we do not parse strings on every iteration. Neorg is slow enough as it is :laughing:

Any alternate ideas perhaps?

champignoom:
I think It will only be parsed once at startup.

PR #1058, opened 2023-09-06
fix(calendar): call `os.date` twice to generate correct weekday

champignoom:
EDIT: fixes https://github.com/nvim-neorg/neorg/issues/1059


max397574:
potentially fixes https://github.com/nvim-neorg/neorg/issues/1059

Issue #1060, opened 2023-09-06
Neorg does not automatically create workspace directories


linguini1:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.9.1

### Neorg setup

```lua
require("neorg").setup({
    load = {
        ["core.defaults"] = {},
        ["core.completion"] = { config = { engine = "nvim-cmp" } },
        ["core.concealer"] = {},
        ["core.dirman"] = {
            config = {
                workspaces = {
                    school = "~/notes/school",
                    personal = "~/notes/personal",
                    projects = "~/notes/projects",
                },
                default_workspace = "personal",
            },
        },
    },
})
```

### Actual behavior

The behaviour that I experience when running `Neorg workspace school` is the opening of a new workspace
named school, placing my cursor into `index.norg`.

When I write some text in the index file and save/write the file, I get the following error:
```
E212: Can't open file for writing: no such file or directory
```

https://github.com/nvim-neorg/neorg/assets/62850091/20d32061-902f-43bf-9e60-8442ad9c5526

### Expected behavior

The behaviour I expected was for the file to be written/saved normally. In the tutorial videos on YouTube, it's
mentioned that Neorg will automatically create the necessary directories for you.

[Said tutorial](https://youtu.be/NnmRVY22Lq8?si=ImMW-YEIXJk2y4d1&t=463)

Interestingly, if I change my setup function to this:
```lua
require("neorg").setup({
    load = {
        ["core.defaults"] = {},
        ["core.completion"] = { config = { engine = "nvim-cmp" } },
        ["core.concealer"] = {},
        ["core.dirman"] = {
            config = {
                workspaces = {
                    notes = "~/notes",
                    school = "~/notes/school",
                    personal = "~/notes/personal",
                    projects = "~/notes/projects",
                },
                default_workspace = "notes",
            },
        },
    },
})
```

Then the directories are created as expected. This leads me to believe that the issue is that Neorg has trouble creating
nested directories unless the parents are specifically listed as work spaces.

### Steps to reproduce

- Use Neovim 0.9.1 on Windows 11
- Install Neorg using Lazy with the following setup:

```lua
--- /lua/plugins/neorg.lua (within neovim config)
return {
    "nvim-neorg/neorg",
    build = ":Neorg sync-parsers",
    dependencies = { "nvim-lua/plenary.nvim" },
    ft = "norg",
[...]

PR #1061, opened 2023-09-06
fix(dirman): correctly create nested directory

champignoom:
fixes https://github.com/nvim-neorg/neorg/issues/1060

### DIFF
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -150,7 +150,7 @@ module.public = {
         end
 
         -- Create the workspace directory if not already present
-        vim.loop.fs_mkdir(workspace, 16877)
+        vim.fn.mkdir(workspace, "p")
 
         -- Cache the current workspace
         local current_ws = vim.deepcopy(module.private.current_workspace)


lua/neorg/modules/core/dirman/module.lua
```diff
@@ -149,8 +149,12 @@ module.public = {
             return false
         end
 
-        -- Create the workspace directory if not already present
-        vim.loop.fs_mkdir(workspace, 16877)
+        -- Recursively create the workspace directory if not already present
+        local slash_pos = -1
+        repeat
+            slash_pos = workspace:find("/", slash_pos+1, true)
```
max397574, line 155: won't work with windows paths

max397574: just use `vim.fn.mkdir(...,"p")`

vhyrro: For future reference we can also use `vim.fs.joinpath`, not in this case but thought it would be nice to point out in case y'all ever need to deal with paths again :)

linguini1:
Thanks for the super quick fix!

PR #1063, opened 2023-09-06
use function callbacks for keybinds and commands

max397574:


max397574:
superseeded by https://github.com/nvim-neorg/neorg/pull/1076

PR #1064, opened 2023-09-06
doc(neorg): fix min reverse indentation count from 2 to 3

Avishayy:
Hey, new neorg user here, thank you for the very cool plugin!

I couldn't figure out why reverse indentation didn't work until I realized the docs were not up to date :) Hope this saves the headache for anyone else.

### DIFF
--- a/doc/neorg.norg
+++ b/doc/neorg.norg
@@ -122,7 +122,7 @@ version: 0.1
       We can move this text to the fifth level again! 🤯
       ---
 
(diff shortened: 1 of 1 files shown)


vhyrro:
Oh! Ironically it's *the other way around*, the docs are more up to date than the parser is haha. A real fix for this would be #1020 but that's temporarily on hold as I work on GTD. Thank you for attempting the fix either way, I'll simply revert the documentation once the new parser gets merged :)

PR #1067, opened 2023-09-07
fix: don't use deprecated `query.get_node_text()` call

linguini1:
When I entered presenter mode in Neorg, I would receive a deprecation warning about `vim.treesitter.query.get_node_text`,
along with a suggestion to replace it with `vim.treesitter.get_node_text`. This PR does exactly that and only that.

### DIFF
--- a/lua/neorg/modules/core/queries/native/module.lua
+++ b/lua/neorg/modules/core/queries/native/module.lua
@@ -133,7 +133,7 @@ module.public = {
 
         for _, node in ipairs(nodes) do
             local temp_buf = module.public.get_temp_buf(node[2])
-            local extracted = vim.split(vim.treesitter.query.get_node_text(node[1], temp_buf), "\n")
+            local extracted = vim.split(vim.treesitter.get_node_text(node[1], temp_buf), "\n")
 
             if opts.all_lines then
                 table.insert(res, extracted)


linguini1:
Seems like the failing lint check is not due to my changes. Unless I need to fix it to be merged, I will leave it be.

linguini1:
> Whoa, this is odd. What Neovim version are you running? The nvim-treesitter crew literally just deprecated `get_node_text` in favour of `query.get_node_text`, did they re-deprecate `query.get_node_text`? I'd be surprised :p

That is odd, I didn't know about this. I'm running the latest nightly (as the contribution guideline prescribes), and below is a screenshot of the error message I receive entering presentation mode. It seems to say that they are in fact deprecating `query.get_node_text` in favour of `vim.treesitter.get_node_text()`.

deprecation error

max397574:
that's really odd
this isn't a really recent change so I don't remember it unfortunately
but in general in all the discussion they try to move function into modules and not "up" like this deprecation would do it

vhyrro:
Update: I've had no time to actually test this out on my end yet which is why it's taking so long to commit. I should have some spare time tomorrow!

vhyrro:
Wow they deprecated quite a few things actually, I'll make patches for the other deprecated functions too. Thank you for the PR and sorry for the wait!

linguini1:
> Wow they deprecated quite a few things actually, I'll make patches for the other deprecated functions too. Thank you for the PR and sorry for the wait!

No need to apologize! Happy to help.

PR #1068, opened 2023-09-08
fix(concealer): don't rerender at `conceallevel` change when disabled

champignoom:
fixes https://github.com/nvim-neorg/neorg/discussions/1066


vhyrro:
Good stuff :D

PR #1076, opened 2023-09-10
feat: (wip) Redesigned event publishing system

Groctel:
As part of the #999 project, this is a WIP PR just so you can see how I'm working on the events system and offer your suggestions and criticism. Check `neorg.core.events` for an in-depth explanation on how the events system is designed.


lua/neorg/core/events.lua
```diff
@@ -0,0 +1,101 @@
...
+
+        -- Metadata members defining Vim's state when the event was published
+        -- TODO: Deprecate? These could be added to the payload on demand.
+        cursor_position = vim.api.nvim_win_get_cursor(0),
+        filename = fn.expand("%:t"),
+        filehead = fn.expand("%:p:h"),
+        line_content = api.nvim_get_current_line(),
+        buffer = api.nvim_get_current_buf(),
+        window = api.nvim_get_current_win(),
+        mode = api.nvim_get_mode().mode,
```
vhyrro, line 92: Am happy to deprecate these! These used to be required when events were sent via vim.schedule, in which case you could get outdated information about the current buffer ID etc.

lua/neorg/core/topics.lua
```diff
@@ -0,0 +1,13 @@
...
+---@alias neorg.event.module_loaded string
+---Informs that a new module has been loaded and added to Neorg's environment.
+---Since only the module's name is published, a module that would like to gather
+---more information about the newly loaded module should retrieve it by calling
+---`neorg.modules.get()`.
+---TODO: Make this a reality!
+
+
+---@alias neorg.event.neorg_started nil
+---Informs that Neorg has finished loaded. Its payload is empty.
```
vhyrro, line 13: Tiny typo - "has finished loaded" -> "has finished loading" :)

lua/neorg.lua
```diff
@@ -5,7 +5,11 @@
 
 -- Require the most important modules
 local neorg = require("neorg.core")
-local config, log, modules = neorg.config, neorg.log, neorg.modules
+local config, events, log, modules = neorg.config, neorg.events, neorg.log, neorg.modules
+
+---@class event.core.neorg_started
```
max397574, line 10: why a class?

lua/neorg/core/topics.lua
Groctel: That was tense...

Groctel (commented):
--Erroneous comment--

lua/neorg.lua
Groctel, line 10: I got a mail from @max397574 asking why this is a class, but don't see the comment in the web interface. It shouldn't! I've now aliased this type to nil to send an empty payload, which is perfectly valid.

max397574: that's because I deleted the comment because I continued reading and saw why it was a class

PR #1079, opened 2023-09-12
feat: allow falling back to the default metagen template functions

mattpolzin:
First time really trying to do anything proper in Lua so no hard feelings if you've got style critiques or I've done something obviously silly!

I was overriding the template for mutagen earlier today and thought it would be really nice if I could specify that I wanted to override some metadata keys & functions but in other cases I wanted to include a key and fallback to the default builtin function.

For example, the following configuration for the template would include all the default keys but only override the values for the authors and categories entries:
```lua
template = {
  { "title" },
  { "description" },
  { "authors", "Steve M. & Martin S." },
  { "categories", "[ Misc ]" },
  { "created" },
  { "updated" },
  { "version" }
}
```


vhyrro:
The code is very clean for a first contribution! Thanks a lot :D

PR #1080, opened 2023-09-13
fix(concealer): revert a wrong fix

champignoom:


lua/neorg/modules/core/concealer/module.lua
```diff
@@ -1278,7 +1278,8 @@ local function handle_init_event(event)
 
         if init_open_folds == "always" then
             open_folds()
-        -- elseif init_open_folds == "never" then
+        elseif init_open_folds == "never" then
+            local luacheck_please_dont_panic
```
max397574, line 1282: uhm please no
just use lsp code actions to insert the comment to disable diagnostics on next line for this

max397574: as I said
you can let code-actions do it for you

max397574: why don't just put a return statement there?
or change the order of conditions and just use `elseif init_open_fold=="auto"` and just nothing else so nothing will be done if it's never or invalid

champignoom: The fallback behavior was (and should be imo) "auto" instead of "never".

Besides, I believe it would be a further step on the wrong way to change the program logic just for silencing a false alarm from luacheck.

max397574: no it definitely wouldn't be a step in the wrong direction imo
because right now the thing was just broken
and if you don't want never as fallback why not just `return`?

champignoom: >because right now the thing was just broken
> and if you don't want never as fallback why not just return?

Because the only logical thing to do within that block is some kind of no-op. Hacking it around by injecting return which just happens to be equivalent to a no-op where there's temporarily no following statements is a starter for bringing back all the quirks in the future.

champignoom: > no it definitely wouldn't be a step in the wrong direction imo

Paying more attention to the false alarm or trying to sacrifice the control flow integrity is the wrong direction.

The right thing to do is to silence it (if that's ever intented) ASAP with minimum side effect and forget about it.

max397574: I disagree
you want to do nothing in the block and return from the function
that's exactly what `return` does

champignoom: > you want to do nothing in the block and return from the function

No I don't. I want to do nothing and go out of the if statement. That's not what `return` means.

champignoom: > no it isn't. It's a bad thing to just have an empty if statement imo

Empty branch in a (de-facto) switch statement is a standard practice and far from being bad.

max397574: > That's not what return means.

Well but that's what it does in this case

mattpolzin: Does Lua accept `break` in this context? Honest question, I'm very new to the language 😃 and could imagine`break` being disallowed outside of loops even though it feels normal in any block to me.

max397574: well if you don't want to add an if statement then just move the thing to the end imo
you won't change my opinion on this. I think that you should either add a return or move it to the end
If someone spends hours debugging why he gets problems when he adds more code to the end he should perhaps consider reading the code at the beginning of the function

max397574: perhaps the best thing would be to just extract this thing into a function `handle_folds` and then a return is absolutely fine and also makes sense because if it's "never" you "stop" handling the folds and return from the function
that function which contains this thing is huuuge anyway

champignoom: That's not about opinion. Adding `return` in this case is just logically plain wrong.

And the question remains: why bother all these? Why refactoring this and that even if they are lucky enough to be correct? For what benefit?

max397574: ofc it's about opinion whether it's okay to have an empty if branch or not
I think not and you say yes
and a `return` isn't logically plain wrong. It does exactly the same
you can discuss about whether it's the cleanest and most future proof solution but it does it's job logically correctly

and if you move the thing into a function you 1. can get rid of the imo ugly empty if branch 2. improve a big function by extracting parts of it and 3. the return makes "logically" sense

champignoom: It is wrong regardless of opinion in the strict sense that adding `return` into an if branch makes the program no longer equivalent to the original version independently from the context.

So no,
> It does exactly the same

It's not because in software development context does matter. It does the same in this specific commit but software developement is not about this specific version and I am quite confident that the previous versions of this module used to be unpleasant to maintain.

If the if-branch deals with something that would cause the init function to malfunction, then `return` would make sense. But obviously here it's not the case.

> and if you move the thing into a function you 1. can get rid of the imo ugly empty if branch 2. improve a big function by extracting parts of it and 3. the return makes "logically" sense

This part I mostly agree, but again:
1. why bother?
2. empty if branch is ugly only in langauges where a proper switch/match construct is offered. Here, luacheck doesn't know what it's complaining about.

champignoom: > ~Does Lua accept `break` in this context? Honest question, I'm very new to the language 😃 and could imagine`break` being disallowed outside of loops even though it feels normal in any block to me.~ [EDIT] Answered my own question. Lua does not accept `break` outside of loops according to the docs I found.

lua does offer `goto` but adding a `goto` to a place where it makes no difference is just insane.

max397574: well do whatever you want I don't want to block this pr with a discussion when we won't ever agree on this anyway
I'll just open a pr afterwards where I extract this into a function
"why bother?"
why not?
when you're working on this code anyway and it would take you like 10 seconds to do it at most

champignoom: > why not?

Because the reason to refactor a piece of code should be the increment of productivity. Not a false alarm from luacheck which, let's be honest, triggered the whole discussion.

In another word, if there should be a PR that changes the control flow, it should be for luacheck, not for neorg.

max397574: does it matter what triggered the whole discussion?
that it makes sense that this goes into a function and that the function in which it is now is way to big is completely independent of luacheck

champignoom: It makes the motivation biased.

You want to silence luacheck by adding `return` -- but that brings risk, ok so you wrap it in a function.

Should other parts of the init function go to their own wrapping functions? Don't know.

Is the introduction of the new function improves reusibility of anything? Unlikely.

A few lines are added and moved around just for pleasing luacheck in a somewhat disguised and "polite" way.

That's the difference of the discussion trigger makes.

max397574: I disagree with almost everything in that message but as I said
just do whatever you want and merge this fix

champignoom: Unless I horribly misunderstood your intention, please be honest and at least admit this part:

> You want to silence luacheck "politely" by adding a seemingly useful return instead of a no-op -- but that brings risk, ok so you wrap it in a function.

max397574: yes

linguini1: Would you accept a slightly different solution of:
```lua
if init_open_folds == "always" then
    open_folds()
elseif init_open_folds ~= "never" then
    if init_open_folds ~= "auto" then
        log.warn('"init_open_folds" must be "auto", "always", or "never"')
    end

    local foldlevel = vim.api.nvim_get_option_value("foldlevel", opts)
    if foldlevel == 0 then
        open_folds()
    end
end
```
No unused branch this way, and no return statement either.

linguini1: > > No unused branch this way,
> 
> I insist that the only thing to be fixed is this arbitrary objective from luacheck, not the switch-case structure where empty arm is nothing uncommon.

🤷‍♂️ just providing an alternate option to consider since the first two have been contested. I don't think luacheck is changing anytime soon.

max397574: you know why the first fix was wrong?
because one would think that an empty if branch is useless and can just be commented out 🤷‍♂️

champignoom: Yeah and that's my point -- the fix was wrong, and even more wrong is the biased impression **coming from luacheck** that empty branch is always useless.

It's not like luacheck is the holy bible and the scapegoat must be found somewhere else.

champignoom: And hence this PR: it silences luacheck, and at the same time it documents luacheck's mistake, instead of covering it up.

vhyrro:
Sorry for the delays in reviewing this.

I believe this was mentioned already, but would we prefer to silence the warning with a luacheck directive in a lua comment moreso than an empty local? All in all the new alternative is fine, but I believe adding an informative comment alongside a luacheck ignore directive is a better decision than tricking luacheck itself.

If you're not sure how to do so, I can quickly send a code block of a patch here, just let me know!

champignoom:
@vhyrro luacheck directive would certainly be the best choice. This PR is more of a first-aid and placeholder of the issue. You could just commit with the luacheck directive directive and then close this PR.

vhyrro:
Fixed in the latest commit with a luacheck directive! :D

Issue #1091, opened 2023-09-21
Link in `:help neorg` points to wrong line in neorg specs


qstickmurph:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.9.2

### Neorg setup

require("neorg").setup {
            load = {
                ["core.defaults"] = {},
                ["core.concealer"] = {},
                ["core.dirman"] = {
                    config = {
                        workspaces = {
                            personal = "~/Documents/notes/personal",
                            work = "~/Documents/notes/work",
                        }
                    }
                },
                ["core.completion"] = {
                    config = {
                        engine = "nvim-cmp",
                        name = "[Neorg]",
                    }
                },
            }
        }

### Actual behavior

The link on line 148 in doc/neorg.neorg points to the wrong line of the neorg specs. It points to line 1336.

### Expected behavior

It should point to line 1340.

### Steps to reproduce

Type `:help neorg` and go to line 148.

### Help

Yes

### Implementation help

I know this is a tiny change, but I'm making a pull request that fixes it.

PR #1092, opened 2023-09-21
fix(`:h neorg`): make link point to correct line in specs

qstickmurph:
Just a quick documentation change to make the link on line 148 of the `:help neorg` doc point to the correct line.

Side note: I made an issue (#1091) before this PR. For future small changes like this one, should I make both the issue and PR or just the PR? Thanks!

### DIFF
--- a/doc/neorg.norg
+++ b/doc/neorg.norg
@@ -145,7 +145,7 @@ version: 0.1
 ** Links
 
    For more info on links check the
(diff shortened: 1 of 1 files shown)


vhyrro:
Awesome, thank you :D 
For small things it makes no difference if you make an accompanying issue or not. A direct PR is also fine!

PR #1105, opened 2023-10-08
feat: support dotrepeat for `promo` and `todo_items`

champignoom:
A dotrepeat wrapper for event handlers is implemented in neorg.utils.
As showcases, promo and task event handlers are wrapped, so for example `>>` and `<C-Space>` can now be dot-repeated.


vhyrro:
API looks good to me. Thanks a lot!

PR #1128, opened 2023-10-25
feat: option to inject specific metadata instead of defaults

### LINKED ISSUE #1126: Create note with differing filename and metadata title
### Issues

### Feature description

# Goal

Create an API that makes it simple to extend Neorg file creation.

# Why

My motivation stems from a desire to create my files with a timestamp template for the filename (e.g. YYMMDD-HHMMSS.norg), however I do not want the file to get auto generated the same timestamp into the title - because then I always need to edit the metadata.

I believe this feature would make file creation in Neorg better suited to extension and make it easier for anyone to modify it to their own needs.

# Proposed Solution

I am thinking about something like:
```lua
require("neorg").get_module("core.dirman").new_note(
    "notes",
    "path/filename",
    { title="awesome note" }
)
```

The above code would then create a file `path/filename.norg` in the workspace `notes`, prefilled with metadata with the title `awesome note`.
The table i am thinking about not just adding a title but potentially also other metadata fields.

### Help

Yes

### Implementation help

I would like to help implement a lua function to call in order to accomplish this.

I am thinking the proper place for this would by in the dirman module, but i may be wrong :)

If you can help me with where best to place this functionality (assuming it aligns with the vision for the project) - that would be great :)

Dynge:
This PR closes #1126. I did not want to deprecate any existing functionality so I've made new functions to `write_metadata` with potential new values instead of defaults.

## Challenges
I am unsure if there is still is any need to use `fs_open()` to create the file, now that I instead create a new tab if `no_open` is true. My reasoning for creating a new tab is that in order to `inject_metadata` i need to have a buffer number, so I always open the new file in a new buffer. With tabs i can easily close the new tab without needing to worry about what else was open at the time.

## Additional notes
I noticed that there does not seem to be an easy way to add categories into the metadata as it seems the metadata fields need to return a single string.
I thought about added functionality to allow tables of strings as well such that you could have a method return a collection of categories (e.g. `{ "[", "    cat1", "    cat2", "]" }`)

### DIFF
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -560,7 +560,7 @@ module.events.defined = {
     workspace_changed = modules.define_event(module, "workspace_changed"),
     workspace_added = modules.define_event(module, "workspace_added"),
     workspace_cache_empty = modules.define_event(module, "workspace_cache_empty"),
-	file_created = modules.define_event(module, "file_created")
+    file_created = modules.define_event(module, "file_created"),
 }
 
 module.events.subscribed = {
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -279,15 +279,15 @@ module.public = {
         end)
     end,
 
-    ---@class create_file_opts
+    ---@class core.dirman.create_file_opts
     ---@field no_open? boolean do not open the file after creation?
     ---@field force? boolean overwrite file if it already exists?
-    ---@field metadata? metadata metadata fields, if provided inserts metadata - an empty table uses default values
+    ---@field metadata? core.esupports.metagen.metadata metadata fields, if provided inserts metadata - an empty table uses default values
 
     --- Takes in a path (can include directories) and creates a .norg file from that path
     ---@param path string a path to place the .norg file in
     ---@param workspace? string workspace name
-    ---@param opts? create_file_opts additional options
+    ---@param opts? core.dirman.create_file_opts additional options
     create_file = function(path, workspace, opts)
         opts = opts or {}
 
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -34,6 +34,19 @@ To query the current workspace, run `:Neorg workspace`. To set the workspace, ru
 ### Changing the Current Working Directory
 After a recent update `core.dirman` will no longer change the current working directory after switching
 workspace. To get the best experience it's recommended to set the `autochdir` Neovim option.
+
+
+### Create a new note
+You can use dirman to create new notes in your workspaces.
+
+```lua
+local dirman = require('neorg').modules.get_module("core.dirman")
+dirman.create_file("my_file", "my_ws", {
+    no_open  = false,  -- open file after creation?
+    force    = false,  -- overwrite file if exists
+    metadata = {}      -- key-value table for metadata fields
+})
+```
 --]]
 
 local neorg = require("neorg.core")
@@ -272,6 +285,8 @@ module.public = {
     ---  - opts.no_open (bool) if true, will not open the file in neovim after creating it
     ---  - opts.force (bool) if true, will overwrite existing file content
     ---  - opts.metadata (table) Table of metadata data, overrides defaults if present
+    ---    - the table is a key-value table where the key correspondes to a metadata field,
+    ---      and the value is either a string or a function returning a string.
     create_file = function(path, workspace, opts)
         opts = opts or {}
 
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -279,10 +279,10 @@ module.public = {
         end)
     end,
 
-	---@class create_file_opts
-	---@field no_open? boolean do not open the file after creation?
-	---@field force? boolean overwrite file if it already exists?
-	---@field metadata? metadata metadata fields, if provided inserts metadata - an empty table uses default values
+    ---@class create_file_opts
+    ---@field no_open? boolean do not open the file after creation?
+    ---@field force? boolean overwrite file if it already exists?
+    ---@field metadata? metadata metadata fields, if provided inserts metadata - an empty table uses default values
 
     --- Takes in a path (can include directories) and creates a .norg file from that path
     ---@param path string a path to place the .norg file in
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -311,28 +311,19 @@ module.public = {
 
         -- Create the file
         local fd = vim.loop.fs_open(fname, opts.force and "w" or "a", 438)
-
         if fd then
             vim.loop.fs_close(fd)
         end
 
-        if opts.no_open then
-            -- new tab for easy closing again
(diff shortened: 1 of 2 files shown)


max397574:
Tbh I don't like the way you do this. The whole things with the vim commands (`:e`, `:tabnew` etc)
Metagen just needs a bufnr and doesn't require the file to be open.
There are api functions to load a file to get a usable bufnr. I could send you those later.

max397574:
Here is an example where this was done
https://github.com/nvim-neorg/neorg-telescope/commit/34a41e809c427d2d27724358ce40c3a413b99e34#diff-38d33560722aa6a06ec8b1cd99066184484042c5302b0fdf0718d0afad1c7029R72-R93
not sure if you need to load the buffer or not for metagen. If not you can just use the function from dirman which's used there
or see how it's done there and just inline it

Dynge:
Ah perfect. Thanks for the pointers @max397574 

Ill edit the file to use that instead.

Any other thoughts on the function names and usage?

Also is there somewhere else i should write dokumentation for the usage?

max397574:
you can just write docs right above the function and inside the comment at the top of the file

Dynge:
I've changed the code to use `dirman.get_file_bufnr` instead - making it much cleaner.

Also tried to add some more documentation to describe the change.

Seems to work great - the only thing I am a little unhappy about is the implementation of categories that does not work so great in this.

It would be nice if categories could be a table and then it would insert them correctly on separate lines - like this:
```lua
dirman.create_file("note", "notes", { metadata = { categories = { "work", "programming" } } })
```
However that does not work in the current implementation because metadata is expected to be simply a string or a function that returns a string - *not* a table.
Its a special case because categories have a special syntax rule for multiline.
This may expand the scope of this pull request - unless there exists a very easy fix for that, that i am not aware of :)

lua/neorg/modules/core/dirman/module.lua
```diff
@@ -34,6 +34,19 @@ To query the current workspace, run `:Neorg workspace`. To set the workspace, ru
...
+
+### Create a new note
+You can use dirman to create new notes in your workspaces.
+
+```lua
+local dirman = require('neorg').modules.get_module("core.dirman")
+dirman.create_file("my_file", "my_ws", {
+    no_open  = false,  -- open file after creation?
+    force    = false,  -- overwrite file if exists
+    metadata = {}      -- key-value table for metadata fields
```
max397574, line 47: what happens when you pass an empty table there? will just default metadata be generated?

lua/neorg/modules/core/dirman/module.lua
```diff
@@ -271,6 +284,9 @@ module.public = {
     ---@param opts? table additional options
```
max397574, line 284: normally people these days use classes for annotations in cases like this
e.g.
```lua
---@class neorg.dirman.new_note_opts
---@field no_open bool
```

max397574 (approved):

max397574:
haven't tested the functionality but code lgtm

lua/neorg/modules/core/dirman/module.lua
Dynge: Yes. 
It overrides default if there's a match in the key names. Otherwise it does nothing.

lua/neorg/modules/core/dirman/module.lua
Dynge: Not aware of how to do that. 
Can you point me to an example?

Dynge: I see this
https://github.com/LuaLS/lua-language-server/wiki/Annotations#class

Should i just create an empty unused table in the top of the file and annotate that?

Is that the way to do it? 😄

Im quite new to lua 😅

Dynge:
Just realize there may be a bug.

I do not think its possible to create a file *without* metadata if metagen is installed anymore.

We need to check if theres auto metagen setting enabled as well. However om not sure how it should work if auto metagen is off and you provide a non-empty table...

Any thoughts?

lua/neorg/modules/core/dirman/module.lua
max397574: you can just write the "class comments" as in my example above somewhere above the function. It would just be a comment block without any "real" code where you define the class with the fields (see this example https://github.com/LuaLS/lua-language-server/wiki/Annotations#field). And then you just use that class as the type of the parameter opts.

max397574:
I think if there is a table it should add metadata. Just set the default to `nil` and I think it should work as expected.

Dynge:
Alright. Ill test it out and later today 👍

lua/neorg/modules/core/dirman/module.lua
Dynge: @max397574  I added @class documentation now :)

Dynge:
> I think if there is a table it should add metadata. Just set the default to `nil` and I think it should work as expected.

Tested it with `core.esupports.type = "none"` and it works (it does not add metadata unless table is not nil) :)

lua/neorg/modules/core/dirman/module.lua
```diff
@@ -265,12 +278,16 @@ module.public = {
             })
         end)
     end,
+
+    ---@class create_file_opts
```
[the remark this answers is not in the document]
Dynge: Should it go even deeper then and name it `neorg.dirman.create_file_opts` then?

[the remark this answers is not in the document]
max397574: up to you
you could search through neorg rq (just telescope live grep for `---@class` (perhaps with space after ---)) and see how it's done in other parts of the codebase

lua/neorg/modules/core/dirman/module.lua
```diff
@@ -308,20 +325,24 @@ module.public = {
...
+        if fd then
+            vim.loop.fs_close(fd)
+        end
 
-            return
+        local metagen = neorg.modules.get_module("core.esupports.metagen")
+        if opts.metadata and metagen then
+            local bufnr = module.public.get_file_bufnr(fname)
+            metagen.write_metadata(bufnr, true, opts.metadata)
         end
```
vhyrro, line 338: All of the code looks good to me apart from this bit. This intertwines the dirman and metagen modules more than it should and doesn't extend to other modules which might also want to do something special when a file is created. Maybe it would be best to create a `core.dirman.file_created` event whose `content` is all of the parameters (`{ metadata, force, no_open }`)? That way any module (including metagen) can listen for this event and do the metadata generation on its own without the two modules ever knowing about each other :)

If you're unsure about how to go around implementing it, feel free to ask!

Dynge: Thanks for the detailed reply. That sounds like a good solution. I am a little unsure how to implement an event.

I would appreciate any help you can give. 😄

Dynge:
Ive moved metagen creation into something that is executed on an event upon file creation instead.
This way dirman does not know what metagen decides to do with the data and it is properly separated. (Thanks for the tip @vhyrro)

I've also removed duplicated functions that I made previously in this pull request because i did not know that you can call functions just with without all input values :) 

I've tested that i can create notes with metagen creation set to "auto" and "never" and it works as expected.

vhyrro:
Ah I'm super sorry for not providing any help. Always ping me a few times so you can be sure I got the message haha.

Now the code looks good to me. Thank you for all the work put into this PR! :purple_heart:

PR #1132, opened 2023-10-29
fix(calendar): display weekdays based on `nvim_strwidth`

pysan3:
Living in a non-latin country, life is sometimes hard with neovim.

Unfortunately in lua 5.1, passing Japanese characters `string.len("あいうえお")` does not count the _letters_ but return the byte length which does not fit accurately when the letters are multi-width. The above should be 5 _letters_ but returns 15.

The following screenshot is the result in the calendar ui of `weekday:sub(1, 2)` displaying corrupted weekdays.

image

As a reference, here's a code you can copy paste to reproduce.
```lua
local weekdays = {
  monday = "月曜日",
  tuesday = "火曜日",
  wednesday = "水曜日",
  thursday = "木曜日",
  friday = "金曜日",
  saturday = "土曜日",
  sunday = "日曜日",
}

for en, ja in pairs(weekdays) do
  vim.print(en)
  vim.print(ja)
  for i = 1, string.len(ja) - 1 do
    vim.print(ja:sub(i, i + 1)) ---> prints corrupted bytecodes
  end
  vim.print("======")
end
```

In this PR, I have implemented a util function that truncates string by the display length to solve this problem. The following image is the result with this PR.

image

---

But here comes the ~~fun~~ part.

Well in Japanese, the first letters of the weekday names are used to distinguish (as *月* 曜日, *火* 曜日, ...) so we have no problem with basically a modified version of `weekday:sub(1, 2)`, but in Chinese, the names are like the image below where the difference is in the last character (星期 _一_, 星期 _二_, ...) which makes the matter worse.

Chinese weekday names

Maybe just using English everywhere with a hard coded lookup table might be a better solution rather than trying to support many languages (especially in the UI space) since I believe _intelligent_ people using vim will have at least some knowledge of English.

---

I searched through the code base and I think this is the only place right now that might call `string.sub` on a non-latin string. Currently, it is mostly used to parse neovim commands and so on.

However, if we are going to parse the filename, metadata, heading or any other things where user can input anything, we should be careful not to use `string.len` or `string.sub` which might cause some issues that are not easy to test.
I'm quite suspicious that there might be a bug where we call `vim.api.nvim_buf_set_extmark` abd calculate the length of the string, as some point out in this issue: https://github.com/neovim/neovim/issues/14010.

### DIFF
--- a/lua/neorg/core/utils.lua
+++ b/lua/neorg/core/utils.lua
@@ -206,4 +206,21 @@ function utils.wrap_dotrepeat(event_handler)
     end
 end
 
+--- Truncate input str to fit inside the col_limit when displayed. Takes non-ascii chars into account.
+---@param str string
+---@param col_limit integer #str will be cut so that when displayed, the display length does not exceed limit
+---@return string #substring of input str
+function utils.truncate_by_cell(str, col_limit)
+    if str and str:len() == vim.api.nvim_strwidth(str) then
+        return vim.fn.strcharpart(str, 0, col_limit)
+    end
+    local short = vim.fn.strcharpart(str, 0, col_limit)
+    if vim.api.nvim_strwidth(short) > col_limit then
+        while vim.api.nvim_strwidth(short) > col_limit do
+            short = vim.fn.strcharpart(short, 0, vim.fn.strchars(short) - 1)
+        end
+    end
+    return short
+end
+
 return utils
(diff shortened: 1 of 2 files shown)


vhyrro:
This is a great fix, thank you for this! As for Chinese I can see that being quite tricky yeah. Perhaps there's some inbuilt lua date utilities to get the shortened weekday name which we could display in the calendar instead. Will have to wait and see :D

PR #1133, opened 2023-10-29
feat: add LaTex rendering for inline equations

nolbap:
Working over the implementation of the `latex-rendering` branch, I added LaTex rendering using the `image.nvim` plugin.

Added the Neorg command `render-latex` which upon being called it gets all the inline math snippets using treesitter, parses them into a temp LaTex file, and compiles the snippet into a png, it then renders the png, concealing the equation if `conceal` option is true (which it is by default) or it renders them in a virtual line under the inline equation. I tested these changes and everything seems to work fine :)

A preview of the module in action!

https://github.com/nvim-neorg/neorg/assets/61060256/8445470c-3478-4444-8720-fd32d052dac7

Note: The concealer uses the `nvim-extmark` option `virt-text-pos = "inline"` which only works for Neovim 0.10.0+, which I added an assertion for :)

P.S. Its my first pull request ever so if anything is wrong let me know 😅


max397574:
just a few points
1. just starting off this discussion: should this be a core module
    - gives errors or warnings to users which aren't on nightly
    - has an external dependency
2. we have to look into the performance of this. Haven't checked but:
    - Check if it's done only for the visible area of the buffer or
    - Don't clear on every cursor movement

I haven't really looked into the code yet. So perhaps some of those things are already ok

vhyrro:
tysm for this! I won't likely have time today but I should be able to review this tomorrow :)

lua/neorg/modules/core/latex/renderer/module.lua
```diff
@@ -0,0 +1,217 @@
...
+        requires = {
+            "core.integrations.treesitter",
+            "core.autocommands",
+            "core.neorgcmd",
+        },
+    }
+end
+
+module.load = function()
+    Image = neorg.modules.get_module(module.config.public.renderer)
```
vhyrro, line 18: Instead of creating a global `Image` variable here I would rather you do the same thing like in `core.integrations.image` - the `module.private` contains the `image` variable and a `pcall` is executed in the `load` function. The rest looks good to me!

nolbap: Sorry for the wait, college has kept me really busy these past few weeks, but I finally got some time and I pushed the commit :)

vhyrro: All good, thanks for all the work!

pysan3:
Hi, nice addition @nolbap but it would be nice if you could take a look at this action and fix where there are errors.

https://github.com/nvim-neorg/neorg/actions/runs/6906276341/job/18790965878

italovieira:
Is this already functional?
I think it's missing in the documentation/wiki, how can I use configure Neorg to render Latex like this?

max397574:
yes it should be
haven't set it up myself yet
ig you can just load the modules

Vaisakhkm2625:
:*) Thanks @nolbap for your hard work.... 

i actually have been working on this (https://github.com/Vaisakhkm2625/hologram-math-preview.nvim) few months ago, but life get the way and was really paining me that i didn't able to move forward with this and hoping for someone to do it...

but based on what i have done, i like to give few suggestion here... using [tectonic](https://github.com/tectonic-typesetting/tectonic) instead of  normal textlive-latex package give much nicer experience, and it's a single library which will automatically manage dependencies by itself, and actually made for applications like embedded previewing.. and instread of dvipng, we can use `pdftocairo -transp -singlefile " .. document_name .. ".pdf -png " .. png_result` to convert to image, which i find much more faster....

i know it's late as you already done all the work, but if you are looking for any improvement, you can take a look at this...

Issue #1108, opened 2023-10-08
Module `core.summary` ignores metadata


austinbutler:
### Prerequisites

- [ ] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

v0.10.0-dev-6823fdb

### Neorg setup

	require("neorg").setup({
		load = {
			["core.defaults"] = {},
			["core.concealer"] = {},
			["core.dirman"] = {
				config = {
					workspaces = {
						notes = "~/notes",
					},
				},
			},
			["core.presenter"] = { config = { zen_mode = "truezen" } },
			["core.summary"] = {},
			["core.integrations.telescope"] = {},
		},
	})

### Actual behavior

`Neorg generate-workspace-summary` generates a summary where everything is under `Uncategorized` and it seems to ignore the title field, too.

### Expected behavior

`Neorg generate-workspace-summary` uses the metadata.

### Steps to reproduce

Add metadata to a note using `inject-metadata`.

```
@document.meta
title: My title
description: My description
authors: austin
categories: Tickets
created: 2023-10-08T13:52:16-0800
updated: 2023-10-08T13:52:16-0800
version: 1.1.1
@end
```

Run `generate-workspace-summary`.

### Other information

Possibly a different issue with the same module: https://github.com/nvim-neorg/neorg/discussions/1085.

### Help

None

snomics:
To add some information, categories are taken into account once a file has been opened. For this file only.

caprilesport:
Is the behaviour commented by @snomics supposed to be the default?

I'm also suffering a bit with it, as sometimes new files are added in a workspace and when i re-generate the summary I have to open all files to have the correct indexing.

aswa2ds:
I met the same situation. And I solved by modified the `get_document_metadata` function in "neorg/lua/neorg/modules/core/integrations/treesitter" like this, and sovled  your question at the same time @caprilesport :
```
    get_document_metadata = function(buf, no_trim)
        buf = buf or 0

        local languagetree = vim.treesitter.get_parser(buf, "norg_meta")

        if not languagetree then
            return
        end

        local result = {}

        -- for _, tree in ipairs(languagetree:children()) do
            -- if tree:lang() ~= "norg_meta" then
            --     return
            -- end
            local tree = languagetree

            ...

                    result[key_content] = (
                        node:next_named_sibling() and parse_data(node:next_named_sibling()) or vim.NIL
                    )
                end
            end
        -- end

[...]

PR #1139, opened 2023-11-03
fix: workspace summary ignore closed files and title field of metadata tag

aswa2ds:
While using `Neorg generate-workspace-summary`. I met some problems similar with issue #1108 , and I solved them in this PR
1. `Uncategorized`: Files not opened cannot be categorized. Because `get_document_metadata` function use parser:children() function to get `norg_meta` parser. But closed file's parser contains no children. So I directly use `norg_meta` parser to parse the metadata, if `norg` parser found the metadata tag.
2. `title field ignored`: while parsing metadata with `bufnr`, the first line, like `@ducument.meta` will confuse the treesitter. Then `@document.meta \n title: <some title>` while considered as a whole part. Then title be ignored.
3. `title like '02' be considered as number type`: if title field is like '02', the title node's type will be number, then the value parsed would be "2".

### DIFF
--- a/lua/neorg/modules/core/integrations/treesitter/module.lua
+++ b/lua/neorg/modules/core/integrations/treesitter/module.lua
@@ -677,6 +677,10 @@ module.public = {
             end
         end
 
+        if not meta_node then
+          return result
+        end
+
         local meta_source = module.public.get_node_text(meta_node, buf)
 
         local norg_meta_parser = vim.treesitter.get_string_parser(meta_source, "norg_meta")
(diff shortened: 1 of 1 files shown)


vhyrro:
Hah you beat me to it by literally just a few minutes as I also sat down to fix this issue :sweat_smile: 

Thanks a lot for the fix!

aswa2ds:
> Hah you beat me to it by literally just a few minutes as I also sat down to fix this issue 😅
> 
> Thanks a lot for the fix!

Oh, I'm sorry about making you a waste of time on this small problem. And I'm so proud of contributing this amazing project.

vhyrro:
You're all good don't worry about it. Thanks again :D

Issue #241, opened 2021-12-14
dirman functions don't support windows


elmarsto:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

PS C:\Users\1atti> nvim --version
NVIM v0.6.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compiled by runneradmin@fv-az152-703

Features: -acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM\sysinit.vim"
  fall-back for $VIM: "C:/Program Files/nvim/share/nvim"

Run :checkhealth for more info
PS C:\Users\1atti>
### Loaded modules
**Check out https://github.com/elmarsto/nvim-config/ for the full config**
core.norg.completion
core.ui.selection_popup
core.norg.dirman
core.mode
core.norg.qol.todo_items
core.norg.esupports
core.norg.esupports.metagen
core.norg.esupports.hop
core.ui
core.norg.manoeuvre
core.gtd.queries.helpers
core.gtd.queries
core.queries.native
core.neorgcmd.commands.module.load
core.neorgcmd.commands.return
core.scanner
core.gtd.ui.selection_popups
core.highlights
core.ui.text_popup
core.integrations.telescope
core.gtd.ui.displayers
core.gtd.ui.views_popup_helpers
core.gtd.ui.edit_popup_helpers
core.gtd.queries.modifiers
core.gtd.queries.creators
core.gtd.ui
core.norg.concealer.preset_varied
core.gtd.queries.retrievers
core.norg.concealer.preset_basic
core.autocommands
core.neorgcmd
core.norg.concealer.preset_diamond
core.integrations.treesitter
core.storage
core.defaults
core.norg.concealer
core.gtd.base
core.integrations.nvim-cmp
core.neorgcmd.commands.module.list
core.keybinds
core.keybinds.default_keybinds
core.norg.news

### Actual behavior

The error is:

```
:Neorg gtd capture
:messages
[neorg] [ERROR 18:41:42] ...rg/lua/neorg/modules/core/gtd/ui/views_popup_helpers.lua:449:  Inbox file is not from gtd workspace.
[neorg]                             Please verify if the file exists in your gtd workspace.
[neorg]                             Type :messages to show the full error report
[...]

PR #1143, opened 2023-11-05
fix(dirman): add `raw_path` option to work with arbitrary filetype

pysan3:
`core.dirman.utils.expand_path` is the function to resolve filepath with workspace notation (`$<workspace>/`).

This should work with arbitrary file type as it will be used to resolve files in for example `.image $/relative/path/to/image.png`, but currently, it is hardcoded to append `.norg` after the file path.

I've added an argument `raw_path` which, when is true, does not append the norg extension.

The previous calls to this function does not specify this arg, (hence nil) so we should have backwards compatibility.

### Note

https://github.com/nvim-neorg/neorg/blob/98f390d8b0c42e40196ea4d831b42b3908137bc3/lua/neorg/modules/core/dirman/utils/module.lua#L45-L47

Not in the scope of this PR but I don't think this works on Windows. I don't have an environment to test Windows so I'll leave it to someone else for a fix.

Possibly related: https://github.com/nvim-neorg/neorg/issues/241

### DIFF
--- a/lua/neorg/modules/core/dirman/utils/module.lua
+++ b/lua/neorg/modules/core/dirman/utils/module.lua
@@ -14,7 +14,11 @@ local log, modules = neorg.log, neorg.modules
 local module = neorg.modules.create("core.dirman.utils")
 
 module.public = {
-    expand_path = function(path)
+    ---Resolve `$<workspace>/path/to/file` and return the real path
+    ---@param path string # path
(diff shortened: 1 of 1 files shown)


pysan3:
Required for this PR.

- https://github.com/3rd/image.nvim/pull/71

vhyrro:
Looks good to me! That path logic should be ideally fixable with some `vim.fs` functions.

Issue #973, opened 2023-07-04
Nested headings for summary


SteveWolligandt:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

I think a nice feature would be to get nested headings if in the meta data you have a category setup like this:

``` neorg
@document.meta
title: My Title
description: My description
authors: Steve
categories: [
  Issue.Bug
  Closed
]
...
@end
```

What it would produce is something like that:

``` neorg
* Issue
** Bug
    - {My Title}[<link>] - My Description

* Closed
  - {My Title}[<link>] - My Description
```

### Help

No

Issue #953, opened 2023-06-16
Option to ignore paths and categories in core.summary


theherk:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

It would be excellent if `core.summary` to configuration for `exclude_categories` and `exclude_paths`. For example:

```lua
          ["core.summary"] = {
            config = {
              exclude_categories = { "no index", "private" },
              exclude_paths = { "journal" },
            }
          },
```

### Help

Yes

### Implementation help

Time is my main issue with contribution to this project. Though I'm absolutely loving it, so I will try to find some. Forgive me for opening so many issues, but I'm trying to transition from a large org, org-roam setup.

UtkarshVerma:
This would be nice to have. I've noticed that `index.norg` is added in the summary, which I'd like to avoid.

PR #1144, opened 2023-11-06
feat: Nested Workspace Summary Categories

aswa2ds:
enabled nested workspace summary categories. Categories with dots can be split to nested categories. And it's allowed to show the category title and file link in the same level category, like school.science --a file link-- and school.test --a category-- in the demo below . #973 
image

And I found that some users wanna ignore some files in workspace summary. So do I. #953
Actually we can use the current ability to realize it. While the categories field was filled with empty list like: []. Like my index.norg file in the demo above. Because of the judge condition below. It hits no branch, and be ignored in the loop below the if-else. And I think it's more flexible than just configuring.
image
image

### DIFF
--- a/lua/neorg/modules/core/journal/module.lua
+++ b/lua/neorg/modules/core/journal/module.lua
@@ -46,6 +46,8 @@ module.examples = {
                                 -- Don't print the year if it hasn't changed
                                 if not current_year or current_year < entry[1] then
                                     current_year = entry[1]
+                                    current_month = nil
+                                    current_quarter = nil
                                     table.insert(output, "* " .. current_year)
                                 end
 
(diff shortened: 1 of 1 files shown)


vhyrro:
Generally the code looks good to me! Just one thing - it might make sense to precompute the value of `lib.title(path)` once in a variable (e.g. `local path_as_title = lib.title(path)`) and use it throughout the code instead of invoking it several times every time we need it. The `lib.title()` function is a bit expensive to run so it would be best if we could run it as few times as possible. If you could fix that up that would be great!

aswa2ds:
😞Had some stupid behaviors. It has been changed as you say. Sorry for ignoring the cost of repeat method call.

vhyrro:
Huh, github seems to be messing things up. Didn't mean to close this, give me a moment :p

vhyrro:
Okay it seems like I messed something up. I merged the changes manually with the main branch because github apparently decided to autoclose this PR. I gave attribution in the commit message. Thank you for the pull request! :purple_heart:

PR #1146, opened 2023-11-06
docs: update README.md

trimclain:
Since the previous [link](https://github.com/nvim-neorg/neorg/wiki/Installation#the-concept-of-modules) doesn't exists anymore, I updated it to the [link](https://github.com/nvim-neorg/neorg/wiki/#module-naming-convention) I thought made sense the most. But of course there might be a better link I didn't find.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,7 @@ require('neorg').setup {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


vhyrro:
Thank you!

Issue #1147, opened 2023-11-07
FEATURE: Timestamp in UTC to Hide Local


pysan3:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

Introduced by [this PR](https://github.com/nvim-neorg/neorg/pull/1052), now neorg updates metadata with time and timezone in the `updated` field.

However, when writing document to be published online ([example](https://github.com/pysan3/minorg)), I do not want to reveal my local due to privacy reasons.

It would be nice if I could set an option somewhere so that the timestamp uses UTC instead of printing the timezone. It would be awesome if I can configure this on per-document basis (I do want local timezone on my journals etc).

Could you give some insights @champignoom, @vhyrro about this issue?

Thanks in advance.

### Help

Yes, but I don't know how to start. I would need guidance

champignoom:
I added a few options [here](https://github.com/nvim-neorg/neorg/pull/1152), but changing them on-the-fly or only for specific files would be orthogonal to the options and thus should be implemented in another PR and/or with your own customization.

PR #1152, opened 2023-11-09
feat(metagen): customize timezone and its format

champignoom:
Resolves https://github.com/nvim-neorg/neorg/issues/1147

Three options are added:
- "utc": like "2023-11-09T04:18:49+0000"
- "local" (default): like "2023-11-09T06:18:49+0200"
- "implicit-local": like "2023-11-09T06:18:49"

### DIFF
--- a/lua/neorg/modules/core/esupports/metagen/module.lua
+++ b/lua/neorg/modules/core/esupports/metagen/module.lua
@@ -29,9 +29,18 @@ end
 local function get_timestamp()
     -- generate a ISO-8601 timestamp
     -- example: 2023-09-05T09:09:11-0500
-    local tz_offset = get_timezone_offset()
-    local h, m = math.modf(tz_offset / 3600)
-    return os.date("%Y-%m-%dT%H:%M:%S") .. string.format("%+.4d", h * 100 + m * 60)
+    --
+    timezone_config = module.config.public.timezone
+    if timezone_config == "utc" then
+        return os.date("!%Y-%m-%dT%H:%M:%S+0000")
+    elseif timezone_config == "implicit-local" then
+        return os.date("%Y-%m-%dT%H:%M:%S")
+    else
+        -- assert(timezone_config == "local")
+        local tz_offset = get_timezone_offset()
+        local h, m = math.modf(tz_offset / 3600)
+        return os.date("%Y-%m-%dT%H:%M:%S") .. string.format("%+.4d", h * 100 + m * 60)
+    end
 end
 
 -- The default template found in the config for this module.
@@ -126,6 +135,12 @@ module.config.public = {
 
     -- Custom template to use for generating content inside `@document.meta` tag
     template = default_template,
+
+    -- Timezone information in the timestamps
+    -- - "utc" the timestamp is in UTC+0
+    -- - "local" the timestmap is in the local timezone
+    -- - "implicit-local" like "local", but the timezone information is omitted from the timestamp
+    timezone = "local",
 }
 
 module.private = {


pysan3:
Amazing, thank you!

Would it be possible to change on per-document basis by any chance?

max397574:
two things 1. @champignoom not sure but is this the time format which we'll also use e.g. for tasks (can be seen in norg-specs)

> Would it be possible to change on per-document basis by any chance?

How would the configuration of that look like?

champignoom:
> two things 1. @champignoom not sure but is this the time format which we'll also use e.g. for tasks (can be seen in norg-specs)

No, because it's an (ISO8601 conformant) extension of the original YYYY-MM-DD format, which was already incompatible with the one used for tasks.

pysan3:
> > Would it be possible to change on per-document basis by any chance?
> 
> 
> 
> How would the configuration of that look like?

1. Look at a different key in metadata. For example we can introduce a `public = true` key to control it (with config option `timezone = metadata` or something like that. 

2. `timezone = git` and check if the file is added to git. 

How do you think?

max397574:
I personally think that it's quite a big effort for a really small niche usecase
but I'm not the one to make decisions here

pysan3:
I think this should be solved in a different PR as a feature request and I'm happy with this PR as-is. 

I will look into it and try to implement it myself (I'm leaning more towards 1; would you agree?)

Thanks for the discussion. 

> niche usecase 

But you do write README.norg in your repo right?

max397574:
> I started thinking this should be solved in a different PR as a feature request and I'm happy with this PR as-is.
> I will look into it and try to implement it myself (I'm leaning more towards 1; would you agree?)

yes sounds good

> But you do write README.norg in your repo right?

I do

pysan3:
Thanks ;)

> > But you do write README.norg in your repo right?
> 
> 
> 
> I do

So how do you hide your timezone there?

max397574:
I don't
but I also don't care about my timezone being revealed
because millions of people live there

vhyrro:
More options don't hurt especially since there are sane defaults :)

Thanks for this! I'll merge it when I get on my PC.

lua/neorg/modules/core/esupports/metagen/module.lua
```diff
@@ -29,9 +29,18 @@ end
 local function get_timestamp()
     -- generate a ISO-8601 timestamp
     -- example: 2023-09-05T09:09:11-0500
-    local tz_offset = get_timezone_offset()
-    local h, m = math.modf(tz_offset / 3600)
-    return os.date("%Y-%m-%dT%H:%M:%S") .. string.format("%+.4d", h * 100 + m * 60)
+    --
+    timezone_config = module.config.public.timezone
```
vhyrro, line 33:
```suggestion
    local timezone_config = module.config.public.timezone
```

Be sure to make this a local variable here! :)

vhyrro:
Also @pysan3 about it being configurable per-document I think this should be a workspace-wide option instead of a per-file option. I find it difficult to see how such a configuration would look like without the metadata self-referencing itself (which I would rather not do for now :p)

pysan3:
Ok got it @vhyrro 

I think if we make it an option for workspace maybe it'd be more useful if we could select between datetime or just date (no time after the T altogether). 

Anyways I found that not many people were interested in this idea unfortunately so I think being able to configure globally satisfies it. Again, thanks for implementing it @champignoom 

You can close this issue https://github.com/nvim-neorg/neorg/issues/1147 after you merge this PR as well.

vhyrro:
Looks good to me now, thanks! :purple_heart:

PR #1153, opened 2023-11-09
feat(hop): feed wslview with decoded link

champignoom:
Temporarily mitigates https://github.com/wslutilities/wslu/issues/295 until it's fixed.


vhyrro:
could you add a small comment in the code explaining that the code is a temporary workaround? After that we can merge :)

champignoom:
> could you add a small comment in the code explaining that the code is a temporary workaround? After that we can merge :)

Done

PR #1157, opened 2023-11-11
feat: register Neorg filetype in plenary.filetype

kashelkin:
It lets Telescope detect Neorg files and attach Treesitter highlighting to preview buffers.

### DIFF
--- a/lua/neorg.lua
+++ b/lua/neorg.lua
@@ -24,6 +24,16 @@ function neorg.setup(cfg)
         },
     })
 
+    -- Make the Neorg filetype detectable through `plenary.filetype`
+    local plenary_filetype = require("plenary.filetype")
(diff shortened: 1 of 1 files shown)


max397574:
I'm not sure if we want this because it would be like a duplicate after 0.10 release

max397574 (approved):

vhyrro:
Hey!

This is a good idea on the surface but Neorg is slowly depending less and less on plenary and I don't fancy making it a hard dependency once again. Unsure why telescope uses a filetype detection system that isn't the native one but a wrapper/extension on top of it :sweat_smile:.

I think the best course of action is to wait for Neovim `0.10.0` where norg becomes an officially recognized filetype and where more than one filetype hook is not required any longer.

Thank you for the PR but I don't see this being the proper solution :)

PR #1162, opened 2023-11-14
refactor(luacheck): workaround lua type check

pysan3:
- TODO: This should be dealt with later and eliminate the TODOs <2023-11-15, pysan3 pysan3@gmail.com>

### DIFF
--- a/docgen/docgen.lua
+++ b/docgen/docgen.lua
@@ -184,7 +184,7 @@ docgen.map_config = function(buffer, start_node, callback, parents)
     local comments = {}
     local index = 1
 
-    for node in start_node:iter_children() do
+    for node in start_node:iter_children() do ---@diagnostic disable-line -- TODO: type error workaround <pysan3>
         if node:type() == "comment" then
             table.insert(comments, ts.get_node_text(node, buffer))
         elseif node:type() == "field" then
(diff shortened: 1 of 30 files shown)


pysan3:
Alright we are good to go @vhyrro 

Dang it, there's a lot of places to fix.

And on top of that, this is not all of them since we have a lot of places where the type is simply `any` where the langserver lost track of the type.

Type errors for `any` are not raised now, but once we start resolving these, there will be more diagnostic errors.

max397574:
I think the standard todo comment would be `TODO(pysan3): ...` without the `<pysan3>` at the end

pysan3:
Hmm, that's interesting. My snippet does this and I blindly accepted it. 

I'll search where it is defined.

pysan3:
https://github.com/L3MON4D3/LuaSnip/wiki/Cool-Snippets#todo-comments-snippets

BTW how'd you also add the date?

```lua
-- TODO: foo bar <2023-11-15, pysan3>
```

max397574:
why would you do that?
that's just rly ugly and no need for it to be there

pysan3:
Besides, your way of doing things will not be detected by https://github.com/folke/todo-comments.nvim.

Where can I get info or specs about that standard?

max397574:
you could also just write no person at all
because in the end you're not the only person that can resolve the todos

vhyrro:
> Besides, your way of writing the todo will not be detected by [folke/todo-comments.nvim](https://github.com/folke/todo-comments.nvim).
> 
> Where can I get info or specs about that standard you proposed?

That's odd hah. The treesitter comment parser will parse the `TODO(vhyrro)` comments in a different colour so I believe that's why everyone rolled with it.

This looks good to me! Doesn't matter what syntax it's using as long as it's informative and gets its point across. We'll be removing these anyway :)

PR #1165, opened 2023-11-15
fix(journal): toc reset month & add link indent

alfiehub:
Fixes a bug in the generated journal ToC where months can only be incremented, resulting in journal entries over multiple years missing months

Also adds some simple indentation to the links so dont they look bad


vhyrro:
Thanks for this! :)

PR #1167, opened 2023-11-16
feat: allow upward paths in tangle

max397574:


max397574:
there might be one `:h` to few appended if the path points to a file instead of a directory (which will mostly be the case)
I'll look into that later today

vhyrro:
Thx!

PR #1170, opened 2023-11-18
feat: add `traverse-link` Neorg mode

shawaman:
A New mode for quick link traversal using "j" and "k" in that mode.


vhyrro:
Thank you! :purple_heart:

PR #1173, opened 2023-11-19
fix(luacheck): fix type errors caused by autoformat

pysan3:
As title.

It seems that stylua automatically reformats the code and the comment are moved so that it doesn't get aligned with the actual problem.

Would you add type checking action to run after the stylua actions as well?
I guess you need to create a separate `GITHUB_TOKEN` for that as I mentioned here: https://github.com/vhyrro/sample-luarocks-plugin/issues/1.

Anyways, this PR changes the places of `---@diagnostic disable-line` so that the langserver is happy again :)

### DIFF
--- a/lua/neorg/modules/core/export/module.lua
+++ b/lua/neorg/modules/core/export/module.lua
@@ -224,17 +224,17 @@ module.on_event = function(event)
         local exported = module.public.export(event.buffer, filetype) ---@diagnostic disable-line -- TODO: type error workaround <pysan3>
 
         vim.loop.fs_open(
-            filepath,
+            filepath, ---@diagnostic disable-line -- TODO: type error workaround <pysan3>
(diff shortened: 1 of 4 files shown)


pysan3:
Just a reminder @vhyrro .

Is there something preventing this PR from being merged?

vhyrro:
Nothing is blocking, thanks for the PR :)

I hope stylua doesn't "autoformat" it back tho :p

pysan3:
Well, it should be the other way around. The main branch should've been _perfect_ and no style issue. 

So stylua should be activated against all PRs and if no one force pushed to main, it should be working. 

I'm not sure why this is not the case now so I'll investigate (as well as ur comment on the other thread).

pysan3:
At least we have a ✅ back up again ☺️

PR #1175, opened 2023-11-19
fix(config): fix luajit version detection

pysan3:
image

LuaJIT version in command `:version` starts with `LuaJIT` and not `LUAJIT` at least on my machine (Manjaro KDE).

But the actual intention of this PR is to bring up the debate on whether we need this function in the first place.

NVIM will only detect the patch version and only when using nightly (1572 in the example above), and in stable (or other tag release versions), it fails to detect.

On top of that, I don't see it being used throughout the codebase.
It is possible that some other external module might be using it, but it is rather advised to use `vim.fn.has("nvim-0.9.0") == 1` if you want a specific neovim version, and I don't think the lua version will change (except for possibly very minor changes) and stick to LuaJIT compatible with `lua v5.1`.

Do you think it is worth keeping this function in the codebase?

I'm asking this question since `vim.api.nvim_exec`, which is used in this function, will be deprecated in `v0.10.0` to `vim.api.nvim_exec2` with different parameters, and if we want to keep the compatibility, we need to write a compatibility layer by ourselves. It'd be easier to simply delete the entire function lol.

### DIFF
--- a/lua/neorg/core/config.lua
+++ b/lua/neorg/core/config.lua
@@ -12,7 +12,7 @@ local function neovim_version()
         end
 
         if not key then
-            key, value = line:match("(LUAJIT)%s+(.+)")
+            key, value = line:match("(LuaJIT)%s+(.+)")
         end
 
         if key then


vhyrro:
Yeah honestly I can't remember why I made that function - I think it had to do with better anticonceal feature detection but it's not that useful anymore. If you want you can remove the whole function in this PR :)

pysan3:
Ok, done @vhyrro !

BTW `vim.api.nvim_exec` will be deprecated in `v0.10.0` and is used elsewhere as well.

The current type check will fail when we run it against the nightly version.

I'm currently thinking of making a new file `core.compatable` with the content like this.

```lua
--[[
    Neorg Wrappers to Keep Compatibility of Deprecated Functions in Neovim Core API.
--]]

local compat_utils = {}

---Wrapper function for `vim.api.nvim_exec2`
---https://github.com/neovim/neovim/pull/19032
---*nvim_exec()*			Use |nvim_exec2()| instead.
---@param src string #Vimscript code
---@param output boolean #Whether to capture and return all (non-error, non-shell |:!|) output. (false)
---@return string #Result of output if `output` is true.
function compat_utils.nvim_exec(src, output)
    if vim.fn.has("nvim-0.9.0") == 1 then
        return vim.api.nvim_exec2(src, { output = output }).output or "" ---@diagnostic disable-line
    else
        return vim.api.nvim_exec(src, output) or "" ---@diagnostic disable-line
    end
end

return compat_utils
```

What do u think about this approach? This file could be used to track on future change and clarify which neovim version are supported OOTB (`vim.fn.has("nvim-0.9.0") == 1` => in the future delete this function and make neorg only support `>= 0.9.0`)

pysan3:
FYI, we need https://github.com/nvim-neorg/neorg/pull/1173 fix as well to work correctly.

vhyrro:
> Ok, done @vhyrro !
> 
> BTW `vim.api.nvim_exec` will be deprecated in `v0.10.0` and is used elsewhere as well.
> 
> The current type check will fail when we run it against the nightly version.
> 
> I'm currently thinking of making a new file `core.compatable` with the content like this.
> 
> ```lua
> --[[
>     Neorg Wrappers to Keep Compatibility of Deprecated Functions in Neovim Core API.
> --]]
> 
> local compat_utils = {}
> 
> ---Wrapper function for `vim.api.nvim_exec2`
> ---https://github.com/neovim/neovim/pull/19032
> ---*nvim_exec()*			Use |nvim_exec2()| instead.
> ---@param src string #Vimscript code
> ---@param output { output: boolean } #Whether to capture and return all (non-error, non-shell |:!|) output. (false)
> ---@return string #Result of output if `output` is true.
> function compat_utils.nvim_exec(src, output)
>     if vim.fn.has("nvim-0.9.0") == 1 then
>         return vim.api.nvim_exec2(src, output).output or "" ---@diagnostic disable-line
>     else
>         return vim.api.nvim_exec(src, output.output) or "" ---@diagnostic disable-line
>     end
> end
> 
> return compat_utils
> ```
> 
> What do u think about this approach? This file could be used to track on future change and clarify which neovim version are supported OOTB (`vim.fn.has("nvim-0.9.0") == 1` => in the future delete this function and make neorg only support `>= 0.9.0`)

It might make sense for us to have a compatibility module but I'd try not to create one if we absolutely don't have to. Because Neorg is versioned properly we could make a specific Neorg version tied to a specific Neovim version (e.g. `6.3.0` is Neovim `0.10+`). That way we don't have old code building up with no incentive from anyone to deprecated it :p

If there is a case though where there's a large breakage then we might want to consider a set of compatibility utilities. So far the vast majority of Neovim users update their Neovims regularly so it's not like we have some people running Neovim 0.3 and others running 0.10 :smile:

vhyrro:
aaand the commits look good to me! Thanks a lot for the cleanup :)

PR #1176, opened 2023-11-19
fix(luacheck): setting non-standard global variables in latex renderer module

nolbap:
As requested by @pysan3 in https://github.com/nvim-neorg/neorg/pull/1133#issuecomment-1817793348 this PR fixes the `luacheck` errors by creating and calling `module.private` variables instead of global variables.


vhyrro:
Thank you! :)

pysan3:
Thanks @nolbap !!

PR #1178, opened 2023-11-20
add shortcut to traverse link

shawaman:
added shortcut for link-traverse mode and updated syntax.

### DIFF
--- a/lua/neorg/modules/core/keybinds/keybinds.lua
+++ b/lua/neorg/modules/core/keybinds/keybinds.lua
@@ -162,18 +162,18 @@ module.config.public = {
             -- Apply the below keys to all modes
             keybinds.map_to_mode("all", {
                 n = {
-                    { leader .. "mn", ":Neorg mode norg<CR>", opts = { desc = "Enter Norg Mode" } },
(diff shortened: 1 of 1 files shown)


shawaman:
@vhyrro Should I remove the link-traversal mode. Because Tab and Shift Tab is more natural to traverse through link and was used in vimwiki.

max397574:
I'm quite sure that in the meantime there already was a mapping for the mode added
and imo we should keep it like this instead of mapping tab

shawaman:
@max397574 
Previously added shortcut was for entering a separate mode and then traverse through link and you have to switch back to follow a link, which is to many steps.

Tabs can be used without entering any mode just like you follow a link using enter. This way is very simple and flows well with linking notes. Using Backspace to navigate backwards will also allow seamless notes reading like a wiki.

I Suggest to keep this simple way and remove the separate mode just to traverse a link and switch back to follow the link.

max397574:
I disagree
I'm against mapping tab because it's such common key
just use the builtin <c-o> to jump back

What imo could be considered though is allowing to follow links in every mode
anyway this should be discussed in a discussion or an issue and not this pr which can be closed

Issue #569, opened 2022-08-31
Nested list icons


arensonzz:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

When I use lists that have mixed nested parts it becomes hard to know which level I am looking at even with the indentation.
I think using slightly different symbols for each list level would improve the look.

### Help

No

max397574:
do you have suggestions for icons?
if we find icons it's easily doable

arensonzz:
•◇▪-→⇒
These are the icons Cherrytree note taking application uses. What are the constraints for icons in Neorg? Can we use any Unicode symbol?

vhyrro:
Fonts are always a difficult thing. Too dynamic to make something look good everywhere :/

Some good news - these "weird" characters will stop being used as soon as anticonceal gets merged in Neovim. One day :smiling_face_with_tear:.

In the meantime I quite like the cherrytree icons to be honest! The diamond does overlap with the heading icon in the "diamonds" preset though, so some extra thought would be required in that department. Keep the suggestions for icons coming!

vhyrro:
Ooh my bad, when I read this i thought we were talking about ordered lists specifically, but you're talking about lists in general. I'm open to the idea for sure.

PR #1179, opened 2023-11-22
feat(concealer)!: simpler config for ordered list icon & multichar icon for unordered list

champignoom:
One can now configure the icon of ordered lists simply with a sample string that represents the iconf for index 1. The icons for the rest of indices will be calculated automatically.

For example:
```lua
["core.concealer"] = {
    config = {
        icons = {
            ordered = {
                icons = { "item 1.", "[0]", "A.", "a.", "§ 1)", "I.", "i.", "⒈", "⑴", "Ⓐ", "ⓐ" },
            },
        },
    },
},
```
the icons will be
- `item 1.`, `item 2.`, ... for level 1
- `[0]`, `[1]`, ... for level 2
- `A.`, `B.`, ..., `Z.`, `AA.`, `AB.`, ... for level 3
- `a.`, `b.`, ..., `z.`, `aa.`, `ab.`, ... for level 4
- `§ 1)`, `§ 2)`, ... for level 5
- `I.`, `II.`, `III.`, `IV.`, `V.`, ..., `MMMCMXCIX.` for level 6
- `i.`, `ii.`, `iii.`, `iv.`, `v.`, ..., `mmmcmxcix.` for level 7
- `⒈`, `⒉`, `⒊`, ..., `⒛` for level 8
- etc

===EDIT===

Now unorderd list also supports multichar inline icon just as ordered list, for example:
```
["core.concealer"] = {
    config = {
        icons = {
            list= {
                icons = { "*", " +", "  -", "   >" },
            },
        },
    },
},
```
This emulates indentation=2 for the first four levels of unordered list:
```
* level-1
  + level-2 (indent=2)
    - level-3 (indent=4)
      > level-4 (indent=6)
       > level-5 (indent=7)
        > level-6 (indent=8)
```

So this PR should partially resolve https://github.com/nvim-neorg/neorg/issues/569.


vhyrro:
Are we sure doing calculations like these aren't performance intensive? If it's still fast then I think we can go with this configuration approach, sure :+1: 

~~Also if you could fix the conflict please that would be great :P~~

champignoom:
> Are we sure doing calculations like these aren't performance intensive?

No, it calculates and memoizes the format string only once.

> Also if you could fix the conflict please that would be great :P

Done.

Issue #1184, opened 2023-11-26
concealer disabled in TOC upon saving


champignoom:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.9.4

### Neorg setup

as in TL;DR

### Actual behavior

Concealer is disabled, `{◉ a}` becomes `{* a}`

### Expected behavior

Nothing

### Steps to reproduce

open test.norg, write
```
* a
** b
*** c
```
then `:Neorg toc`, then `:w`

### Help

Yes

PR #1186, opened 2023-11-26
fix(concealer): get winid of bufid when rendering

champignoom:
Fixes https://github.com/nvim-neorg/neorg/issues/1184.

The problem is that, when the toc buffer is created, its filetype is not yet set, so the `BufNew` event is ignored by concealer, and the init function who registered the line change listeners is not called.

To solve this problem, `BufNew` event is replaced by `FileType` event.

Hopefully this won't break anything.


vhyrro:
Works with all the cases I threw at it... let's hope there are no nasty breakages in weird edge cases haha. Thanks!

Issue #1189, opened 2023-11-27
conceal numeric footnote reference to superscript


champignoom:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

Conceal for example `{^ 123}` to `¹²³`.

### Help

Yes

Anrock:
That's very nice. At least brings some reason to use footnotes instead of plain headers. Maybe that behaviour should be mentioned in documentation?

PR #1190, opened 2023-11-27
feat(concealer): conceal numeric footnote title to superscript 

champignoom:
Resolves https://github.com/nvim-neorg/neorg/issues/1189.

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -388,6 +388,19 @@ local function get_ordered_index(bufid, prefix_node)
     return count, (sibling or header_node:parent())
 end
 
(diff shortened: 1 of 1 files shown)


max397574:
just wondering
why are some autocommands commented out?
if they should be removed remove them
if this is a wip mark it as draft

champignoom:
See https://github.com/nvim-neorg/neorg/pull/1186, on which this PR happens to be based.

I'll rebase it to main if that solves the problem.

max397574:
much nicer diff like this

vhyrro:
This is beyond an awesome idea. The concealer is easily the coolest part of the Neorg TUI experience right now :rocket:

PR #1191, opened 2023-11-27
fix(modules): Check the right config key in module.wants

mdietrich16:
If a module `wants` another module, the loading process checks `config.user_config` for that module configuration. However, module configs are located in `config.user_config.load`.


vhyrro:
Oh, this is such a simple mistake. Thanks for the fix! :D

PR #1192, opened 2023-11-27
fix(latex): Make latex module want image integration, (partially) fixes #1183

### LINKED ISSUE #1183: Latex-rendering error
### Prerequisites

### Neovim Version

0.10.0

### Neorg setup

opts = {
        load = {
            ["core.defaults"] = {},
            ["core.concealer"] = {
                config = {
                    folds = true,
                    init_open_folds = "always",
                    icon_preset = "diamond",
                },
            },
            ["core.dirman"] = {
                config = {
                    default_workspace = "notes",
                    workspaces = {
                        notes = "~/Documents/Sync/neorg/",
                    },
                },
            },
            ["core.completion"] = {
                config = {
                    engine = "nvim-cmp",
                },
            },
            ["core.summary"] = {},
            ["core.integrations.telescope"] = {},
            ["core.integrations.image"] = {},
            ["core.latex.renderer"] = {},
        },
    }

### Actual behavior

With `image.nvim` installed, the intagrations enabled and rendering another image,
upon running :Neorg render-latex I get the following message
```
Error executing Lua callback: ...y/neorg/lua/neorg/modules/core/latex/renderer/module.lua:172: attempt to index field 'image' (a nil value)
stack traceback:
    ...y/neorg/lua/neorg/modules/core/latex/renderer/module.lua:172: in function 'on_event'
    .../.local/share/nvim/lazy/neorg/lua/neorg/core/modules.lua:709: in function 'broadcast_event'
    ...im/lazy/neorg/lua/neorg/modules/core/neorgcmd/module.lua:286: in function <...im/lazy/neorg/lua/neorg/modules/core/neorgcmd/module.lua:193>
```

After deleting the `.image /some/path.png` I now get (weird) latex rendering but also 
```
Error executing Lua callback: ...y/neorg/lua/neorg/modules/core/latex/renderer/module.lua:153: bad argument #1 to 'rep' (number expected, got nil)
stack traceback:
    [C]: in function 'rep'
    ...y/neorg/lua/neorg/modules/core/latex/renderer/module.lua:153: in function 'render_inline_math'
    ...y/neorg/lua/neorg/modules/core/latex/renderer/module.lua:174: in function 'on_event'
    .../.local/share/nvim/lazy/neorg/lua/neorg/core/modules.lua:709: in function 'broadcast_event'
    ...im/lazy/neorg/lua/neorg/modules/core/neorgcmd/module.lua:286: in function <...im/lazy/neorg/lua/neorg/modules/core/neorgcmd/module.lua:193>
```

Weird latex rendering means multiple instances of the same image after scrolling, the neorg src for latex still gets rendered and the locations are not in sync.

### Expected behavior

Render latex instead of neorg src i guess?

### Steps to reproduce

Well, try to render latex _with_ another image rendered and also try to render latex on its own.

_No response_

### Other information

_No response_

### Help

None

### Implementation help

_No response_

mdietrich16:
When opening any `norg` file (not necessarily with math inside) directly in neovim with both the `integration.image` and `latex.renderer` configured, every event registered by the latter throws an error.
This is because the latex renderer tries using `module.private.image`, which is the image module and gets required on latex module load. If the image module is however not yet loaded, this return nil and therefore throws an error later.

There are a couple of possible fixes for this.
- In `neorg.module.get_modules` load the module and not just return `nil`.
- Do a better check after invoking `get_module`, since it never errors but only return nil. This does not fix anything, however.
- Use this pull request to "want" the image integration, so it is loaded on latex module load. However, this might not be desired, if multiple possible renderers are to be allowed in the future.
- Do a lazy-load of the image module. I do not know if this is possible right now.
- Make latex renderer lazy-load. Right now, it registers all autocmds it could need. This still requires a forced loading of the image integration later, however.

### DIFF
--- a/lua/neorg/core/modules.lua
+++ b/lua/neorg/core/modules.lua
@@ -279,7 +279,7 @@ function modules.load_module_from_table(module)
 
             -- This would've always returned false had we not added the current module to the loaded module list earlier above
             if not modules.is_module_loaded(required_module) then
-                if config.user_config[required_module] then
+                if config.user_config.load[required_module] then
                     log.trace(
                         "Wanted module",
                         required_module,
--- a/lua/neorg/modules/core/latex/renderer/module.lua
+++ b/lua/neorg/modules/core/latex/renderer/module.lua
@@ -6,6 +6,9 @@ assert(vim.re ~= nil, "Neovim 0.10.0+ is required to run the `core.renderer.late
 
 module.setup = function()
     return {
+        wants = {
+            "core.integrations.image",
+        },
         requires = {
             "core.integrations.treesitter",
             "core.autocommands",


vhyrro:
I think for now `want`ing the module is good enough on its own. It will probably take a while before we create any other renderers, so this is just fine for the time being. Whenever we need more we can then start thinking of adding extra loading logic. Thanks!

PR #1193, opened 2023-11-27
feat(concealer): add configuration for hrule start and end position

champignoom:
Horizontal rules can now start from the very left of window (`left="window"`) or from the node position (`left="here"`), and end at the very right of window (`right="window"`) or at the column of `textwidth` defaulted to 79 just like vim (`right="textwidth"`).

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -533,13 +533,17 @@ module.public = {
             end
         end,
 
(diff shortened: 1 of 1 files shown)


pysan3:
@champignoom Lua check fails.

https://github.com/nvim-neorg/neorg/actions/runs/7009500418/job/19068081326

vhyrro:
Fixed it :)

pysan3:
Amazing. Thanks!

PR #1194, opened 2023-11-27
feat(toc): auto shrink toc vsplit width upon creation

champignoom:
The ToC vsplit window width is 1/2 of original window width by default.

With this PR, the vsplit width will be auto shrinked to the max (virtual) line length + 1 (for margin) if this is smaller than default.


vhyrro:
Super neat, I agree this should be the default :rocket:

PR #1197, opened 2023-11-27
feat(toc): add config for enabling synchronized cursorline in toc window

champignoom:
New boolean option: `sync_cursorline`.

When enabled:
- `cursorline` is enabled (highlighted) in the ToC window
- when the cursor of the content window changes, the cursor of the ToC window changes accordingly


champignoom:
I did but the force-push is ignored in the PR. See the new PR from the same branche but with a different diff.

I'm too tired to open a issue for GitHub itself.

PR #1198, opened 2023-11-27
feat(toc): add config for enabling synchronized cursorline in toc window

champignoom:
New boolean option: `sync_cursorline`.

When enabled:
- `cursorline` is enabled (highlighted) in the ToC window
- when the cursor of the content window changes, the cursor of the ToC window changes accordingly

So that one can track which section they're reading in the ToC window.

### DIFF
--- a/lua/neorg/modules/core/qol/toc/module.lua
+++ b/lua/neorg/modules/core/qol/toc/module.lua
@@ -304,6 +304,16 @@ local function get_max_virtcol()
     return result
 end
 
+local function unlisten_if_closed(ui_buffer, listener)
+    return function(ev)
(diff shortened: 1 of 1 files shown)


rodhash:
Wow nice feat! Thank you!

Only one thing I was thinking about

When we open the toc pane the cursor automatically goes all the way to the top.. it goes to  the first line "Table of Contents"

Would be nice if the cursor kept synced too .. let's say I'm in the middle of the file under some heading named "foo" .. would be nice if opening the TOC pane the cursor goes to the heading "foo" instead of going to the first line..

this is how Markdown TOC behaves and I find this kind of behavior very useful in large MD files.. after opening the TOC I don't need to scroll down in order to find where I was

Thanks again for this

pysan3:
Typecheck fails @champignoom @vhyrro. Please fix it.

https://github.com/nvim-neorg/neorg/actions/runs/7059860778/job/19218279880

PR #1222, opened 2023-12-13
fix(concealer): use vim.treesitter.foldexpr for stabler folding

champignoom:
`vim.treesitter.foldexpr()` behaves more stably than `nvim_treesitter#foldexpr()`


vhyrro:
How recent is `vim.treesitter.foldexpr`? Should we maybe have a fallback to the old function in case the user is running a version that doesn't have this API (`vim.treesitter.foldexpr and "foldexpr" or "fallback"`)? Neorg should work for 0.8 and above :)

Issue #1212, opened 2023-12-06
Add hop-back keybind


mbpowers:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

I think it would be good to add a hop-back keybinding to the hop module. 
I imagine it could be as easy as mapping a key (default as backspace imo) to `<cmd>bun<cr>` or `<cmd>bp<cr>`.

### Help

Yes

### Implementation help

I don't use multiple buffers often so I expect my suggested implementation might have ways it could break, in which case I would need suggestions on a the best way to implement.

max397574:
just use `<c-o>`

mbpowers:
If you navigate around the file in ways that affect the jumplist then you will have to <c-o> many times to go back to the previous file and it will affect the cursor position which I would ideally be preserved.

mbpowers:
`<c-^>` works fine once but then it cycles so if you go `A -> B -> C` pressing `<c-^>` will go `C -> B -> C -> B`.

I personally see it as giving users a way to easily achieve what they might want without having to know the intricacies of neovim bindings and buffers etc. or even if they know it will be repeated work as I think many will want this.

Thoughts on opt-in (could accept a key to bind to)? Or if the concern is it's not worth it to undergo this to add just a single binding, I think it would be nice to add as a note to the hop wiki.

champignoom:
I lose track of buffers as well, not just in neorg but vim in general, so my solution was to customized the hop module so that `<CR>` opens links in new tabs, instead of trying to navigate through the invisible list of buffers.

max397574:
is that already a builtin option? to open the links in tabs? Otherwise I'd be nice if you could open a pr

mbpowers:
`:bp` and `:bun` have unwanted side-effect as expected, I think the best way would be to store the path of the note that you are coming from in the destination note's buffer variables. This could be done when a hop is executed.

vhyrro:
I would have to think about this one, more specifically how it would scale with complex jump trees (when you jump between files and links your path isn't always guaranteed to be linear).

It's true that the jumplist isn't informative enough as is. I think what would be *more* interesting is building up a tree of jumps which one could navigate. Then a simple `hop-back` would move up the tree, but we could provide more interesting navigation methods to go along with it.

[...]

PR #1225, opened 2023-12-14
feat(hop): keybinding for opening links in new tab and with external apps

champignoom:
Resolves comment https://github.com/nvim-neorg/neorg/issues/1212#issuecomment-1855507662

One can now:
- use `\gt` (t for tab) to open links in new tab
- use `\ge` (e for external) to open links with external apps

### DIFF
--- a/lua/neorg/modules/core/keybinds/keybinds.lua
+++ b/lua/neorg/modules/core/keybinds/keybinds.lua
@@ -77,6 +77,14 @@ module.config.public = {
                         opts = { desc = "Jump to Link (new tab)" },
                     },
 
+                    -- Same as `<CR>`, except opens the destination with external applications
+                    {
+                        leader .. "ge",
+                        "core.esupports.hop.hop-link",
+                        "external",
+                        opts = { desc = "Jump to Link (external)" },
+                    },
+
                     -- Same as `<CR>`, except opens the destination in a vertical split
                     {
                         "<M-CR>",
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/keybinds/keybinds.lua
```diff
@@ -65,9 +65,25 @@ module.config.public = {
 
                     -- Hop to the destination of the link under the cursor
                     { "<CR>", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
-                    { "gd", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
-                    { "gf", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
-                    { "gF", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
+                    { leader .. "gd", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
+                    { leader .. "gf", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
+                    { leader .. "gF", "core.esupports.hop.hop-link", opts = { desc = "Jump to Link" } },
```
vhyrro, line 70: The idea of the PR sounds good (having more keybinds for more specific ways of opening a file), however the purpose of these keybinds were to emulate the vim-specific `gd`, `gf` keybinds. Adding them to the leader just clutters up the keybind space a lot. We can't use something like `ge` as that corresponds to going backwards to the end of the word within vim.

I feel like we shouldn't try to have every possible way of opening a file as a key, especially not by default, as this will just mess with the brains of users (KISS).

vhyrro:
Closing this one as "overcomplicating things" :)

Issue #1228, opened 2023-12-15
Conceal comments


champignoom:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

Conceal

`%inline comments%`

and

```norg
#comment
block comments
that may span across lines
```

### Help

Yes

PR #1229, opened 2023-12-15
feat(concealer): hide comments

champignoom:
Resolves https://github.com/nvim-neorg/neorg/issues/1228


max397574:
fyi you can also just create PRs without issues describing the problem first

PR #1230, opened 2023-12-17
fix(neorg.norg): clarify horizontal line syntax

mikemc:
Fix error regarding horizontal lines.


vhyrro:
Thanks! The new syntax parser will allow two or more but currently it's three or more so it makes more sense for the docs to reflect it this way :)

PR #1234, opened 2023-12-19
fix: `update-metadata` would fail to work with several parse trees in the document

### LINKED ISSUE #1233: update-metadata doesn't work with code cells in buffer
### Prerequisites

### Neovim Version

NVIM v0.9.4 Build type: Release 
LuaJIT 2.1.1693350652

    system vimrc file: "$VIM/sysinit.vim"
   fall-back for $VIM: " /nix/store/xnfdsm9dgfng2spl17m7b7wgfc0yyaa4-neovim-unwrapped-0.9.4/share/nvim "

### Neorg setup

```lua
{
    load = {
      ["core.defaults"] = {},
    }
}
```

### Actual behavior

`:Neorg update-metadata` doesn't work when there's a code cell in the file

### Expected behavior

it should work

### Steps to reproduce

Open a .norg file
`:Neorg inject-metadata`
add a code block:
```
@code python
print('hi') # need to load the python TS parser
@end
```

none

### Other information

This is the issue: https://github.com/nvim-neorg/neorg/blob/baaf13a3145534144b795ad37db22bfffd2ad343/lua/neorg/modules/core/esupports/metagen/module.lua#L282-L284

We're early returning if the first lang isn't correct. But we should probably check all of them.

### Help

Yes

### Implementation help

I'm fairly sure I know what's going on here. But I'll update this issue if I get smacked by reality.

benlubas:
fixes #1233

no way I got #1234

### DIFF
--- a/lua/neorg/modules/core/esupports/metagen/module.lua
+++ b/lua/neorg/modules/core/esupports/metagen/module.lua
@@ -280,16 +280,17 @@ module.public = {
 
         for _, tree in pairs(languagetree:children()) do
             if tree:lang() ~= "norg_meta" or meta_root then
-                return
+                goto continue
             end
 
             local meta_tree = tree:parse()[1]
 
             if not meta_tree then
-                return
+                goto continue
             end
 
             meta_root = meta_tree:root()
+            ::continue::
         end
 
         if not meta_root then


vhyrro:
Ah, I should have seen this issue coming while writing the code but apparently I didn't, lol. Thank you for the fix!

PR #1236, opened 2023-12-20
fix(hop): fix range check across lines

champignoom:
Fixes an error when hopping a link across lines:
```norg
{/ /link/target}[description
across
lines]
```


vhyrro:
Sweet! I've been wanting to fix this for so long but never found the time to bother with it, awesome.

PR #1239, opened 2023-12-21
Add a new strategy `by_path` for summary generation

johanneshorner:
I like to organize my .norg files in subfolders and I'd like to use the names of these subfolders as the category for the files contained in them. With this change the command `generate-workspace-summary` can make use of this information.


vhyrro:
This is a great idea. Thank you for the PR! :purple_heart:

PR #1248, opened 2023-12-25
docs(README): improve lazy.nvim installation instructions

refaelsh:
I've added:
1. Lazy loading on file type.
2. The "tag" thingy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,8 +252,6 @@ You can install it through your favorite plugin manager:
     {
       "nvim-neorg/neorg",
       build = ":Neorg sync-parsers",
-      -- lazy-load on filetype
-      ft = "norg",
       -- tag = "*",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -252,6 +252,9 @@ You can install it through your favorite plugin manager:
     {
       "nvim-neorg/neorg",
       build = ":Neorg sync-parsers",
+      -- lazy-load on filetype
+      ft = "norg",
```
vhyrro, line 256:
```suggestion
```

Adding the optional `tag = "*"` sounds great, however it's generally not a good idea of lazy loading Neorg in general, since it exposes commands that can be used from outside of `norg` files. We used to have things like this in the readme but had to remove them because they were just confusing users. If you could remove these two lines then that would be great :)

vhyrro:
thanks!

PR #1250, opened 2023-12-26
feat(tangle): add `report_on_empty` option in `core.tangle`

pysan3:
> Do we need this warning? https://github.com/nvim-neorg/neorg/blob/aa8e66dd40c07a4de58f9ed93f27ab4dac9a241c/lua/neorg/modules/core/tangle/module.lua#L424
> I run an autocmd for all norg files to automatically try tangling, and I don't want to get a warning everytime.
>
> @pysan3

## Changes

- Add `report_on_empty` option.

```lua
module.config.public = {
    -- Notify when there is nothing to tangle (INFO) or when the content is empty (WARN).
    report_on_empty = true,
}
```

- Fixes some type errors from https://github.com/nvim-neorg/neorg/pull/1162

## Notes

- I'm not sure how to update the wiki: https://github.com/nvim-neorg/neorg/wiki/Tangling
  - I'd be happy if you could guide me how.
- Possibly conflicts with https://github.com/nvim-neorg/neorg/pull/1167

### DIFF
--- a/lua/neorg/modules/core/tangle/module.lua
+++ b/lua/neorg/modules/core/tangle/module.lua
@@ -436,53 +436,40 @@ module.on_event = function(event)
         local tangled_count = 0
 
         for file, content in pairs(tangles) do
-            local upward_count = 0
-      
-            for _ in string.gmatch(file, "%.%.[\\/]") do
-                upward_count = upward_count + 1
-            end
-      
+            -- resolve upward relative path like `../../`
+            local relative_file, upward_count = string.gsub(file, "%.%.[\\/]", "")
             if upward_count > 0 then
-                -- adding one because the filename also has to be removed
-                local base = vim.fn.fnamemodify(vim.fn.expand("%"), ":p" .. string.rep(":h", upward_count + 1))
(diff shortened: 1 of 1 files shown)


vhyrro:
I sure love github's default commit messages. Either way the conflict should be resolved now properly. Could you double check if this is the case @pysan3? Afterwards we can merge it.

pysan3:
@vhyrro As you can see, now the ci result is a ❌ due to these errors caused by the PR I mentioned.

image

Are they gonna be my fault now? Should I fix them in this PR?

pysan3:
During some experiments, does the current v1 parser parse relative tags correctly in the first place?

I don't think this exports to the correct path (or doesn't do tangling in the first place?). 

```txt
#tangle ../init.lua
@code lua
print("Hello world")
@end
```

Upward relative path does work when specified in the `@document.meta` tho.

pysan3:
The above commit resolves the conflict and uses better mechanism to generate the output filepath without some complicated lua code like counting gmatch.

vhyrro:
Fantastic, thanks a lot for all the work.

PR #1251, opened 2023-12-26
fix(export): Fix metadata missing values when exported to markdown

### LINKED ISSUE #1247: Markdown exported metadata missing values
### Prerequisites

### Neovim Version

0.9.4

### Neorg setup

```
{
    load = {
        ["core.defaults"] = {},
        ["core.export"] = {},
        ["core.export.markdown"] = {
            config = { extensions = "all" },
        },
    },
}
```

### Actual behavior

When exporting the metadata block of a file to markdown, the value of each key-value pair appears empty

https://github.com/nvim-neorg/neorg/assets/43814863/25395667-ff5e-4cc1-8cc5-a7afdb17d2dc

### Expected behavior

The values of the metadata key-value pairs should be copied over to the markdown YAML header (converting their format as appropriate)

### Steps to reproduce

1. Create a `.norg` file
2. Add a header with at least 1 key-value pair
3. Convert the `.norg` file to markdown
4. See the values of the pairs be missing

_No response_

### Other information

When the export module descends over the `pair` nodes, it finds a `key` node and a node of type `number`, `string`, `array` or `object`. The `key` node is correctly copied over because the `functions` table in `core.export.markdown` has a `key` entry set to `true` (copy the value over). However, the other node types don't have an entry in that `functions` table, which causes the exporter to try to descend into those nodes. `string` and `number` don't have any child nodes, so the exporter doesn't return anything and thus they are ignored. For `object` and `array` nodes, this recursion into the values always eventually ends at `string` or `number` nodes, which are ignored due to the previous issue (with keys of `object` nodes found along the way being exported)

The `functions` table in `core.export.markdown` does have a `value` entry, which from what I understand is a generic node representing a node of type `number`, `string`, `array` or `object`. However, when the exporter searches for the node type in the `functions` table, it searches for the specialized node type, which causes it to miss this entry

https://github.com/nvim-neorg/neorg/blob/0ccc3bba3f67f6f8740b86a50aa5e1428327a741/lua/neorg/modules/core/export/markdown/module.lua#L556-L562

I think the solution would be to use this function for `string` and `number` nodes, which also handles `array` nodes. `object` nodes should be handled separately and are a bit trickier due to the first key having a lower indentation level and the prefix `- ` when inside of an `array` node. Additionally, indentation should be added to key-value pairs to properly format the content of `object` nodes

### Help

Yes

### Implementation help

I'm not sure how to handle the indentation of key-value pairs for `object` nodes when their parent is an `array` node, since the first pair would need a different indentation than the others

ALVAROPING1:
fixes #1247

https://github.com/nvim-neorg/neorg/assets/43814863/b4bb48da-f7f0-4948-af8b-aec62a0d801a

### DIFF
--- a/lua/neorg/modules/core/export/markdown/module.lua
+++ b/lua/neorg/modules/core/export/markdown/module.lua
@@ -76,12 +76,13 @@ local function todo_item_extended(replace_text)
     end
 end
 
-local function handle_metadata_literal(text, node, state)
-    if node:parent():type() == "array" then
-        return "\n" .. string.rep(" ", state.indent) .. "- " .. text
-    end
+local function get_metadata_array_prefix(node, state)
+    return node:parent():type() == "array" and string.rep(" ", state.indent) .. "- " or ""
+end
 
-    return text
+local function handle_metadata_literal(text, node, state)
+    -- If the parent is an array, we need to indent it and add the `- ` prefix. Otherwise, there will be a key right before which will take care of indentation
+    return get_metadata_array_prefix(node, state) .. text .. "\n"
 end
 
 local function update_indent(value)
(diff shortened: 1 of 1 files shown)


vhyrro:
Seems to work well on my end, thank you for the fixes!

Issue #1255, opened 2023-12-29
Add `opts` to `remap`, `remap_event` (core.keybinds)


aorith:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

Hi,

I just noticed that:

```lua
hook = function(k)
  k.map("norg", "n", "<LocalLeader>T", "<cmd>Neorg toc<cr>", { desc = "[neorg] ToC" })
end,
```

Shows the description but:

```lua
hook = function(k)
  k.remap("norg", "n", "<LocalLeader>T", "<cmd>Neorg toc<cr>", { desc = "[neorg] ToC" })
end,
```

Does not, since the later does not accept `opts` as argument ([ref](https://github.com/nvim-neorg/neorg/blob/a489e7c4f9d7edb3caa04250d07bb6c6a5b9b890/lua/neorg/modules/core/keybinds/module.lua#L231)).
I think it would be a good idea if all the map/remap functions accept `opts`.

### Help

Yes

aorith:
BTW, it seems like the keybind hooks are being triggered multiple times. For instance, in the following configuration:

```lua
load = {
  ["core.defaults"] = {},

  ["core.keybinds"] = {
    config = {
      default_keybinds = true,
      hook = function(k)
        k.map("norg", "n", "<LocalLeader>T", "<cmd>Neorg toc<cr>", { desc = "Table of Content" })
      end,
    },
  },
}
```

The `map` hook is executed six times for the same keybind whenever a `norg` file is opened.

And in this configuration:

```lua
load = {
  --["core.defaults"] = {},

  ["core.keybinds"] = {
    config = {
      default_keybinds = true,
      hook = function(k)
        k.map("norg", "n", "<LocalLeader>T", "<cmd>Neorg toc<cr>", { desc = "Table of Content" })
      end,
    },
  },
}
```

The `map` hook is called four times.
Is it normal or there is a weird loop going on here?

PR #1256, opened 2023-12-30
feat(keybinds): add `opts` arg to `remap(_event)`

aorith:
For consistency with the `map` keybind hook. (#1255)

The new `opts` table has priority over the old one. Feel free to reject the PR if not allowing to override `opts` is intentional.


vhyrro:
Code looks good to me! Thanks :)

PR #1263, opened 2024-01-06
Feat!: Keybind Refactor

max397574:
### Todo:
Function to detect conflicts which will automatically cleanup the keybinds table (if that's wanted).
Not yet done because it's not clear whether a conflicting keybind should still be set or not. A warning would be omitted in any case.
Perhaps that could also be something displayed in a healthcheck. -> would have to get merged table before setting keymaps and `vim.keymap.get` all of them to see if they are set to something


max397574:
closed in favor of https://github.com/nvim-neorg/neorg/pull/1509

PR #1265, opened 2024-01-06
[Lua]: Add annotations to the core codebase.

vhyrro:
This PR adds LuaCATS annotations (i.e. the annotation standard used by the lua language server) to the core of Neorg (not to all modules, but just to the core).
LuaCATS is a more modern spinoff of EmmyLua.

This gives proper type checking, better autocompletion and verifying the contents of tables provided to the `setup()` function. It's even useful to users as they can have autocompletion for their config :D

### DIFF
--- a/lua/neorg.lua
+++ b/lua/neorg.lua
@@ -5,6 +5,8 @@
 local neorg = require("neorg.core")
 local config, log, modules = neorg.config, neorg.log, neorg.modules
 
+--- @module "neorg.core.config"
+
 --- Initializes Neorg. Parses the supplied user configuration, initializes all selected modules and adds filetype checking for `.norg`.
 --- @param cfg neorg.configuration.user A table that reflects the structure of `config.user_config`.
 --- @see config.user_config
--- a/lua/neorg.lua
+++ b/lua/neorg.lua
@@ -1,35 +1,33 @@
---[[
---    ROOT NEORG FILE
---    This file is the beginning of the entire plugin. It's here that everything fires up and starts pumping.
---]]
+--- @brief [[
+--- This file marks the beginning of the entire plugin. It's here that everything fires up and starts pumping.
+--- @brief ]]
 
--- Require the most important modules
 local neorg = require("neorg.core")
 local config, log, modules = neorg.config, neorg.log, neorg.modules
 
---- This function takes in a user config, parses it, initializes everything and launches neorg if inside a .norg or .org file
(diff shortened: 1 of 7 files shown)


vhyrro:
It seems to me like the CI is using some older version of LuaLS which does not use LuaCATS. @pysan3 do we pin the luals-typecheck language server to a specific version?

vhyrro:
It seems that there are general issues with the typechecking CI, but lua-ls itself is not having any issues when parsing the codebase for me locally. I'll merge the PR as-is and I hope the other issues will fix themselves later on.

pysan3:
The type checking is using this repo, which I have no control over.

https://github.com/mrcjkb/lua-typecheck-action

I'm not very sure, but as far as I understand, it uses nix script? to setup the environment, and as nix has to be run against a lock file (`flake.lock`), I think the lang server version is locked.

It seems that the repo auto generates a new release when a new version of luals is released, so we would need to update the `@v0.2.*` always.

GitHub is not cleaver enough to resolve `@v0.2` as the latest release of `v0.2.*`, and we would need to ask the repo owner to make a branch named `v0.2` and update it recordingly to match the latest `@v0.2.*` release.

pysan3:
Oh, you seem to have pinned it to the latest version: https://github.com/nvim-neorg/neorg/commit/fb23d2e78bf6ee601ed1de2a9ded23d6201f7506

vhyrro:
Yeah I tried to fix it by pinning to the latest version but it's still failing for some reason :thinking: 

It's unable to find the neodev files, despite them being there (I've been debugging the CI for the best part of half an hour now :/)

pysan3:
I think upstream v0.2 fails to load [`.luarc.json`](https://github.com/nvim-neorg/neorg/blob/main/.github/workflows/.luarc.json) and is running the LS without any config file, hence does not recognize neodev.

pysan3:
LuaLS seems to accept the command line arg `--config-path` as stated here https://luals.github.io/wiki/configuration/, but the action passes `--configpath` (without `-` before path), and ignores that arg i think.

I don't know when it changed, but I think this is the reason.

pysan3:
@vhyrro Found the reason!

Abs path to the work dir has changed from `v0.1.*` to `v0.2.*` in [mrcjkb/lua-typecheck-action](https://github.com/mrcjkb/lua-typecheck-action) since it moved from Dockerfile to nix action style.

So we had to change this in `.github/workflows/.luarc.json`.

image

PR #1267, opened 2024-01-06
fix(ci): fix abs path to libs in luarc

pysan3:
> Yeah I tried to fix it by pinning to the latest version but it's still failing for some reason :thinking: 
>
> It's unable to find the neodev files, despite them being there (I've been debugging the CI for the best part of half an hour now :/)
>
> _Originally posted by @vhyrro in https://github.com/nvim-neorg/neorg/issues/1265#issuecomment-1879760949_

### DIFF
--- a/.github/workflows/.luarc.json
+++ b/.github/workflows/.luarc.json
@@ -6,7 +6,7 @@
     "lua/?/init.lua"
   ],
   "workspace.library": [
-    "/github/workspace/deps/neodev.nvim/types/stable"
+    "/home/runner/work/neorg/neorg/deps/neodev.nvim/types/stable"
   ],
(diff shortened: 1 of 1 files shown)


pysan3:
And now the CI is failing ❌ for _real_ issues, which should have been fixed before force merging the type annotation PR as-is...

pysan3:
In order to resolve the `undefined-field` errors, I think we need to add

```lua
---@type core.xxx.xxx
local foo = module.required["xxx.xxx"]
```

since they cannot be resolved automatically.

vhyrro:
Let's go! Thanks a lot. I'll spend the rest of today trying to fix as much as I can then.

Issue #1112, opened 2023-10-12
Do not change undo history when auto updating metadata


champignoom:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### Feature description

Normally, `u` followed by `<ctrl-r>` should be a no-op.

But if I save the file in between, that is `u` then `:w` then `<ctrl-r>`, the saving action modifies the timestamp and creates a new undo branch, which prevents redo.

It would be nice if the undo status can somehow remain unaffected upon saving.

### Help

Yes

max397574:
if  we'd just set `vim.o.ul=vim.o.ul` that'll create a new "undo break" or however you want to call that. Then you can undo the change and switch to a different undo branch (idk how but somehow it's possible in vanilla, otherwise use e.g. undotree.vim) and  you're able to redo the change.

champignoom:
Never mind. We just have to replace `<ctrl-r>` with `g-`, so `u` then `:w` then `g-` does the job.

kflak:
Not sure if this is even the same issue, but using metadata makes undo unusable for me... It only happens when the `updated` field exists, otherwise no problem.

https://github.com/nvim-neorg/neorg/assets/13571232/fd339c74-448d-403a-a961-85e69f535706

System:
Arch linux latest,
❯ nvim -v
NVIM v0.10.0-dev-741+g996dd36c77
Build type: RelWithDebInfo
LuaJIT 2.1.1702233742
Run "nvim -V1 -v" for more info

Neorg commit:  5432be2

Best, and thanks for an awesom plugin!

benlubas:
@kflak this is b/c you have an autosave plugin. I use one too, just spent a little time solving it. but the way that I went about doing it seems like a hack, and requires that you remap your `u`/`<c-r>` keys for the best experience. see the PR that just got linked.

kflak:
aaaah... makes sense! Had similar issues with autosave and Oil.nvim now that you mention it. Thanks for the PR!

PR #1272, opened 2024-01-11
feat: undojoin timestamp updates

benlubas:
addresses issues brought up in #1112 (currently closed)

## Problem
Updating the time stamp can cause issues with undo-ing things. This is particularly bad when an autosave plugin is used b/c the user will make a change > autosave > timestamp update > autosave > undo (undoes timestamp update) > autosave > timestamp update > loop. The scenario's described assume there is an autosave plugin saving on text change/insert leave.

Making it impossible to undo changes in the norg file with autosave (while the `updated:` field is present in the metadata).

## Proposed Solution
This PR makes two changes that make it possible to bind keys to mitigate this issue:
- `["core.esupports.metagen"].config.undojoin_updates = true` - When true, we call `:undojoin` just before updating the `updated:` field in the metadata. With this set to true you can make a change, write the file and press `u` to undo your change _and_ the timestamp update. **But**, you will still get another time stamp update after the undo. Making multiple undos in a row impossible b/c of branches. Enter: the second addition..
- `require("neorg.modules.core.esupports.metagen.module").public.skip_next_update()` - a function which will disable exactly one metadata 'updated at' update. This function can prevent the issue above, and makes it possible to remap the `u` and `<c-r>` keys so that they behave mostly normally in a norg doc with autosave.

This solution is maybe not the most elegant. I'm open to feedback on how to make it better, I threw this together quickly and without much thought after I realized that it would mostly work.

I'd like to test drive this for a few more days as well before it potentially gets merged.

## Example Usage

```lua
--- config..
["core.esupports.metagen"] = {
  config = {
    undojoin_updates = true,
  },
},
--- config..
["core.keybinds"] = {
  config = {
    hook = function(keybinds)
      keybinds.map("norg", "n", "u", function()
        require("neorg.modules.core.esupports.metagen.module").public.skip_next_update()
        local k = vim.api.nvim_replace_termcodes("u<c-o>", true, false, true)
        vim.api.nvim_feedkeys(k, 'n', false)
      end)
      keybinds.map("norg", "n", "<c-r>", function()
        require("neorg.modules.core.esupports.metagen.module").public.skip_next_update()
        local k = vim.api.nvim_replace_termcodes("<c-r><c-o>", true, false, true)
        vim.api.nvim_feedkeys(k, 'n', false)
      end)
    end,
  },
},
--- config..
```

### DIFF
--- a/lua/neorg/modules/core/esupports/metagen/module.lua
+++ b/lua/neorg/modules/core/esupports/metagen/module.lua
@@ -141,14 +141,20 @@ module.config.public = {
 
     -- Timezone information in the timestamps
     -- - "utc" the timestamp is in UTC+0
-    -- - "local" the timestmap is in the local timezone
+    -- - "local" the timestamp is in the local timezone
     -- - "implicit-local" like "local", but the timezone information is omitted from the timestamp
     timezone = "local",
+
+    -- Whether or not to call :h :undojoin just before changing the timestamp in `update_metadata`
+    -- This will make your undo key undo the last change before writing the file in addition to the
+    -- timestamp change. This will move your cursor to the top of the file.
+    undojoin_updates = false,
 }
 
 module.private = {
     buffers = {},
     listen_event = "none",
+    skip_next_update = false,
 }
 
 ---@class core.esupports.metagen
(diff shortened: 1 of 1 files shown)


benlubas:
I've used this for about a week and haven't had major issues with it.

This definitely feels like a hack. But I can't think of a better way to do this (other than, just don't use autosave in norg files).

vhyrro:
Sounds like a good idea for the time being. I can't really think of any alternative solution either. Thank you!

PR #1274, opened 2024-01-11
feat(summary): reimplement nested categories

smtucker:
### Why:

Looks like nested categories were added to the summary module but then later disabled.

### What:

This adds it back in with some subtle changes so it works with the current codebase. It also plays nicely with selecting the desired (sub)categories by argument given. Did a little refactoring to shorten the default strategy function and make it easier to read.

This is my first time submitting a PR. I'm just too excited about Neorg to hold back. Feedback appreciated.


vhyrro:
I can confirm it works on my end! Thanks a lot for the PR :D

PR #1285, opened 2024-01-23
fix(concealer): footnote pattern should be matched against full string

champignoom:
Numeric footnote links are superscripted, but the ones mixed with non-digits (like  `{^ abc123456abc}`) were mistakenly superscripted (to a mess) as well. This PR fixes the issue.

Dash (`-` ) is now also supported.

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -456,7 +456,6 @@ module.public = {
             local link_title_node = node:next_named_sibling()
             local link_title = vim.treesitter.get_node_text(link_title_node, bufid)
             if config.numeric_superscript and link_title:match("^[-0-9]+$") then
-                vim.print(link_title)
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/concealer/module.lua
```diff
@@ -454,11 +455,12 @@ module.public = {
         footnote_concealed = function(config, bufid, node)
             local link_title_node = node:next_named_sibling()
             local link_title = vim.treesitter.get_node_text(link_title_node, bufid)
-            if config.numeric_superscript and link_title:match("%d+") then
+            if config.numeric_superscript and link_title:match("^[-0-9]+$") then
+                vim.print(link_title)
```
vhyrro, line 459: Seems like some debugging code was left over? :p

champignoom: I was drunk, sorry.

PR #1286, opened 2024-01-23
fix(highlights)!: updated default groups to match names in treesitter

ALVAROPING1:
Matches the new names in https://github.com/nvim-treesitter/nvim-treesitter/commit/1ae9b0e4558fe7868f8cda2db65239cfb14836d0

This is a breaking change for users whose colorscheme hasn't been updated yet to follow this change

### DIFF
--- a/lua/neorg/modules/core/highlights/module.lua
+++ b/lua/neorg/modules/core/highlights/module.lua
@@ -178,7 +178,7 @@ module.config.public = {
 
         -- Highlights for all the possible levels of ordered and unordered lists.
         lists = {
-            unordered = { prefix = "+@punctuation.delimiter" },
+            unordered = { prefix = "+@markup.list" },
 
             ordered = { prefix = "+@keyword.repeat" },
         },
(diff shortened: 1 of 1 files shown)


ALVAROPING1:
https://github.com/nvim-neorg/neorg/blob/28cbafbc0cde316d8e2618ff26377200225e9393/lua/neorg/modules/core/highlights/module.lua#L312-L319
Is there any reason for these to not be linked to `@markup.underline` and `@markup.strikethrough` respectively?

https://github.com/nvim-neorg/neorg/blob/28cbafbc0cde316d8e2618ff26377200225e9393/lua/neorg/modules/core/highlights/module.lua#L181
Should we link this to `@markup.list`?

vhyrro:
Enough time should have passed for people to update now properly.

@ALVAROPING1 with the new hl groups we will want to remap to `markup.strikethrough` and `markup.underline` yep! The `@markup.list` change also sounds good to me :+1:

ALVAROPING1:
should be good to merge now then, I have updated the remaining groups

vhyrro:
Fantastic, thank you! :)

PR #1296, opened 2024-02-07
docs(README.md) clarify lazy load in lazy.nvim snippet

dangkhoipro:
Because some lazy.nvim distributions set lazy=true by default, I add some config to make the plugin load correctly. (and add a lazy load config when I am in there too)


pysan3:
Thanks ;)

vhyrro:
Looks good to me, thanks!

PR #1313, opened 2024-02-21
doc: pin version to v7.0.0 in tldr

champignoom:
Otherwise the tl;dr instruction fails due to https://github.com/nvim-neorg/neorg/commit/56ad8056b6180dba60ddbd5bca2f29de12f3bd1d.


champignoom:
Btw @vhyrro any recommended colorscheme that supports the latest treesitter?

PR #1315, opened 2024-02-23
feat: refactor module

benlubas:
_This is really messy commit history, please squash merge_

Implements a very basic refactoring module in lua. This is a temporary solution while LSP is stalled.
Adds a link tools module that lets you parse links from a buffer in a convenient way. I made this its own module b/c it seemed like it might be useful

I honestly don't know if it's worth while to merge this, or just wait for everything else to be setup. I implemented this mostly just for fun

## Features

`:Neorg refactor rename file`
`:Neorg refactor rename heading`

Two commands that will prompt for input, and ask the user for a new name.

- Headings can change levels, and file can change directories.
- uses buffer contents when they're available
- uses TS to find links
- links that use `#` keep their `#` even if the heading level changes

### limitations

- doesn't handle links in the current file that have a filepath that points to their own file (ie. `{:path/to/foo:}` in `foo.norg` is not handled)


benlubas:
This is ready for review and serious consideration now

lua/neorg/modules/core/dirman/utils/module.lua
```diff
@@ -76,7 +84,7 @@ module.public = {
     ---@param raw_path boolean? # If true, returns resolved path, otherwise, returns resolved path and append ".norg"
     ---@return string? # Resolved path. If path does not start with `$` or not absolute, adds relative from current file.
     expand_path = function(path, raw_path)
-        local res = module.public.expand_pathlib(path, raw_path)
+        local res, _ = module.public.expand_pathlib(path, raw_path)
```
pysan3, line 87: FYI substitution to a variable will only pick up the first return value, so no need for this change.

```lua
local function foo()
  return 1, 2, 3
end

print(foo()) -- 1, 2, 3

local a = foo()
print(a) -- 1
```

benlubas:
I'm going to replace this with a new PR in the near future

PR #1316, opened 2024-02-23
feat: move to luarocks

vhyrro:
This PR will be a massive paradigm shift in how we do things - get ready!

# Neorg + Luarocks = :purple_heart: 

This PR shifts Neorg from being a git-first plugin to being a luarocks-first plugin. Luarocks is the standard Lua package manager.

This is what it means for everyone:
- Developers (us!) can easily push breaking changes and dependency changes to a plugin without any user intervention.
- Complex build/installation/compilation scripts are handled automatically, and compilation scripts can be cached so users don't actually have to do any compilation on their machine(s).
- Dependencies are automatically installed on users' machines, versioning is enforced.
- Access to the entire luarocks code repositories, maximizing code reuse.
- Norgopolis will be very easy to embed as part of a future version (once again with no user intervention required!)

Luarocks genuinely makes plugin management within Neovim much, much easier.

Here are the things that will be changing for users post-`8.0.0`:
- The recommended installation for method will be [rocks.nvim](https://github.com/nvim-neorocks/rocks.nvim). This is not shilling for the project in any real way, it's just the only active plugin manager that exists with builtin luarocks support (packer is discontinued and lazy [doesn't have it](https://github.com/folke/lazy.nvim/issues/37)).
 While it's much easier to install something from luarocks when the plugin manager itself supports luarocks, lazy users can still install Neorg just fine with some tweaks.
- Lazy users will need to install a plugin that will support luarocks installation within lazy. This will likely be an improved version of https://github.com/camspiers/luarocks once I start sending some PRs to it. Neorg will then ship with a `build.lua` which will attempt to interface with the luarocks plugin to install Neorg as a rock - thanks to this the user won't have to do any installation themselves. This should make the experience streamlined, generally hassle-free and should hopefully encourage more people to start adopting luarocks in their workflows :)

After this is released I'll also spend a bunch of time going through the Neorg codebase and cleaning it up (including fleshing out existing features and improving them where possible). Now that we'll have access to libraries whose developers spent ages perfecting, we won't have to worry about odd bugs popping up in our code. More developer time will be aimed at feature development instead of writing code that has already been written for us :laughing:


vhyrro:
Everything should now be working! The implementation can be tested with the following snippet:
```lua
    {
        "vhyrro/luarocks.nvim",
        branch = "more-fixes",
        config = function()
            require("luarocks").setup({})
        end,
    },
    {
        "nvim-neorg/neorg",
        branch = "luarocks",
        dependencies = { "luarocks.nvim" },
        config = function()
            require("neorg").setup({
                load = {
                    ["core.defaults"] = {},
                }
            })
        end
    }
```

If you get any errors in the meantime then please do let me know!

vhyrro:
We are nearing the release of Neorg 8.0.0! Just a few errands left to run :)

vhyrro:
Moment of truth!

PR #1319, opened 2024-02-25
fix(export.markdown): export `authors` metadata field key as `author`

ALVAROPING1:
When pandoc reads the metadata block of a markdown file, it uses the `author` field to set the authors' names. Neorg, however, uses the `authors` field for this. This patch converts the key name during export so the data can be properly recognized on both cases


vhyrro:
Thx!

Issue #1325, opened 2024-02-29
Jumping to line adds unwanted jump history


champignoom:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.10.0-dev-2480+g0eaae1bc0

### Neorg setup

tl;dr

### Actual behavior

Jump to first line of `tmp.txt`

### Expected behavior

Jump back to `tmp.norg`

### Steps to reproduce

- `$ seq 10 > tmp.txt`
- `$ echo '{/ ./tmp.txt:3}' >tmp.norg`
- `$ nvim tmp.norg`
- press `<enter>`
- press `<ctrl-o>`

### Help

Yes

PR #1326, opened 2024-03-01
fix(hop): remove extra jump to line

champignoom:
fixes https://github.com/nvim-neorg/neorg/issues/1325


vhyrro:
Managed to find a simpler fix for it that does not break anything (tested with all working link types). See 918f2a39f96e1447c00871eb611bed2018a047b5. Thanks for the PR either way :)

PR #1329, opened 2024-03-03
feat: integrations.otter

benlubas:
Integrates with [otter.nvim](https://github.com/jmbuhr/otter.nvim), a plugin that provides LSP features for embedded code. This includes diagnostics, go to def/typedef/refs, rename, etc. 

This integration will configure otter.nvim correctly (`otter.setup({ handle_leading_whitespace = true })`), so you just need to install it.

There are two commands, `:Neorg otter enable` and `:Neorg otter disable`.


benlubas:
Hmmm. This does call activate every time you enter the buffer. I'm not sure I like that. I think I'll switch to a new auto command

vhyrro:
Tested and it all looks good to me!

PR #1331, opened 2024-03-05
feat: add author field to provide custom author name in config

antigluten:
### Feature description

Add a capability to override system username that is used in generated metadata.

### Usage 

```
{
    ["core.esupports.metagen"] = {
        config = {
            author = "Vladimir Gusev",
        },
    },
}

```

### Expected behaviour

```
authors: Vladimir Gusev
```


Chouhartem:
Very nice, I can remove my ugly hack now <3

```lua
require('neorg').utils.get_username = function()
  return "Chouhartem"
end
```

PR #1332, opened 2024-03-09
feat: inject metadata on file creation if the feature toggled

antigluten:
### Feature description

Add a capability to inject metadata automatically on file creation using the 'dirman.new.note' keybinding.

### Usage 

```
{
    ["core.dirman"] = {
        config = {
            ...,
            inject_metadata = true,
        },
    },
}

```

### Expected behaviour

Metadata is injected when a file created.


vhyrro:
Hey -- metadata generation on new files should already be handled by metagen. If it's not automatically creating metadata on new files then that's a bug. I believe I already saw an issue about metagen not generating files with `new.note`, so I'll be looking into that.

I wouldn't like this approach because it glues together the behaviours of two separate modules, whereas the whole idea of modules is to be as isolated as possible. Thanks for the efforts either way :)

Issue #1121, opened 2023-10-18
Heading + Todo + Folding (+ concealer?) = trouble


j-xella:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.10.0-dev-1361+g0892c1cd9

### Neorg setup

```lua
-- neorg support
-- https://github.com/nvim-neorg/neorg

return {
  {
    "nvim-neorg/neorg",
    build = ":Neorg sync-parsers",
    dependencies = {
      "nvim-lua/plenary.nvim",
      "nvim-treesitter/nvim-treesitter",
      "nvim-tree/nvim-web-devicons",
    },
    cmd = { 'Neorg', },
    ft = { 'norg', },
    config = function(_,_)
      require("neorg").setup {
        load = {
          ["core.defaults"] = {},
          ["core.dirman"] = {
            config = {
              default_workspace = "dev",
              workspaces = {
                personal = "hidden",
                dev      = "hidden",
              },
            },
          },
          ["core.keybinds"] = { -- https://github.com/nvim-neorg/neorg/wiki/User-Keybinds
            config = {
              hook = function(keybinds)
                keybinds.remap_event( "norg", "n", "<C-1>", "core.qol.todo_items.todo.task_cycle" ) -- Like in OneNote

                vim.opt.conceallevel = 2
              end,
            },
          },
          ["core.qol.todo_items"] = {
            config = {
              order = {
                { "undone", " " },
                { "pending", "-" },
                { "done", "x" },
              },
            },
          },
          ["core.concealer"] = {},
          ["core.summary"] = {},
          ["core.ui.calendar"] = {},
        }
      }
    end,
  },
}
```

### Actual behavior

Watch the second-level heading on line 32.  In insert mode, as is:

![image](https://github.com/nvim-neorg/neorg/assets/8959819/196b934a-10e2-4e0d-8fc2-412207562893)

Now in normal mode, concealed. Unfolded - looks OK:

![image](https://github.com/nvim-neorg/neorg/assets/8959819/4799f6e3-ba3f-4de9-9836-93f433c20cbb)

Same heading, folded. **Looks BAD**:

![image](https://github.com/nvim-neorg/neorg/assets/8959819/ce947276-be27-4547-b1cc-297419638e4f)

### Expected behavior

Now removing the TODO marker:

[...]

PR #1335, opened 2024-03-15
Fix for issue #1121

j-xella:
This should fix the problem described in #1121. Maybe with the latest greatest nvim this problem is also bypassed (see 086891d). But the fix is still relevant for older nvim versions.

The problem seems to be that lua `line:sub` method is not Unicode-aware and does not work properly with strings that contain Unicode characters. That is definitely the case when concealer required more than one substitution (one for heading, one for todo marks). First substitution (heading) would work fine, but the second one (todo mark) would become off.


vhyrro:
LGTM, thanks for the fix!

PR #1356, opened 2024-03-28
TODO Introspector version 2 electric boogaloo

vhyrro:
The old introspector code was forgotten over a year ago, so time to revive it!

The introspector is the handy module which displays `[0/1] (0%)` next to things that have subtasks.

TODOs:
- [ ] Make the display super customizable (will be done in a different PR, I would like to release this in a working state beforehand) - currently the introspector will display one format only
- [x] Correctly handle the deletion of lines (somewhat inconsistent right now, difficult to fix)
- [x] Correctly handle nodes which are broken into two (one list becoming two)

Everything else is fully functional already :)

### DIFF
--- a/lua/neorg/modules/core/todo-introspector/module.lua
+++ b/lua/neorg/modules/core/todo-introspector/module.lua
@@ -48,6 +48,19 @@ function module.public.attach_introspector(buffer)
         error(string.format("Could not attach to buffer %d, buffer is not a norg file!", buffer))
     end
 
+    module.required["core.integrations.treesitter"].execute_query(
+        [[
+    (_
+      state: (detached_modifier_extension)) @item
+    ]],
+        function(query, id, node)
+            if query.captures[id] == "item" then
+                module.public.perform_introspection(buffer, node)
+            end
+        end,
+        buffer
+    )
+
     vim.api.nvim_buf_attach(buffer, false, {
         on_lines = vim.schedule_wrap(function(_, buf, _, first)
             ---@type TSNode?
--- a/lua/neorg/modules/core/todo-introspector/module.lua
+++ b/lua/neorg/modules/core/todo-introspector/module.lua
@@ -148,6 +148,7 @@ function module.public.perform_introspection(buffer, node)
     -- as the total also includes things like uncertain tasks.
     vim.api.nvim_buf_set_extmark(buffer, module.private.namespace, line, col, {
         virt_text = { { string.format("[%d/%d]", counts.done, total), "Normal" } },
+        invalidate = true,
     })
 end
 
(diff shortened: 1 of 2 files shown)


lua/neorg/modules/core/todo-introspector/module.lua
```diff
@@ -0,0 +1,155 @@
...
+            if module.private.buffers[buf] then
+                return
+            end
+
+            module.private.buffers[buf] = true
+            module.public.attach_introspector(buf)
+        end,
+    })
+end
+
```
pysan3, line 40: Please add this for type annotation reason.

```lua
---@class core.todo-introspector
module.public = {}
```

vhyrro:
@pysan3 any idea why TSNode is not being picked up? It is in my config, and my config doesn't have any special setup lol

I assume these definitions are from the Neovim source code itself? Could that be added to the CI as well if it isn't already?

lua/neorg/modules/core/todo-introspector/module.lua
```diff
@@ -0,0 +1,155 @@
...
+    vim.api.nvim_buf_clear_namespace(buffer, module.private.namespace, line, line + 1)
+
+    if total == 0 then
+        return
+    end
+
+    -- TODO: Make configurable, make colours customizable, don't display [x/total]
+    -- as the total also includes things like uncertain tasks.
+    vim.api.nvim_buf_set_extmark(buffer, module.private.namespace, line, col, {
+        virt_text = { { string.format("[%d/%d]", counts.done, total), "Normal" } },
```
pysan3, line 151: It might be nice if the string generation is defined as a separate public function so that it may be used in lualine etc.

In neo-tree we have a `get()` function separated from `vim.wo.winbar = `

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/16d1b194376bf1fc2acd89ccb3c29ba8315bfcea/lua/neo-tree/ui/selector.lua#L225-L241

https://github.com/nvim-neo-tree/neo-tree.nvim/blob/16d1b194376bf1fc2acd89ccb3c29ba8315bfcea/lua/neo-tree/ui/selector.lua#L391-L402

which makes it possible to do things like this: https://github.com/nvim-neo-tree/neo-tree.nvim/issues/1368#issuecomment-1970541005 (inject neo-tree's selector into bufferline.nvim (topline))

vhyrro:
Code is almost ready, just some cleanup left :)

lua/neorg/modules/core/todo-introspector/module.lua
vhyrro: I'll definitely do that when I implement the customization options for this! Sounds good.

Issue #1365, opened 2024-03-30
:Neorg journal wipes existing journal files


Ladas552:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

NVIM v0.9.5

### Neorg setup

```
      require("neorg").setup({
        load = {
          ["core.defaults"] = {},
          ["core.concealer"] = {
            config = {
              icon_preset = "diamond",
            },
          },
          ["core.summary"] = {},
          ["core.esupports.metagen"] = { config = { timezone = "implicit-local", type = "empty", undojoin_updates = "false"} },
          ["core.completion"] = { config = { engine = "nvim-cmp", name = "[Norg]" } },
          ["core.keybinds"] = {
            config = {
              default_keybinds = true,
              neorg_leader = "<Leader><Leader>",
            },
          },
          ["core.journal"] = {
            config = {
              workspace = "journal",
            },
          },
          ["core.dirman"] = {
            config = {
              workspaces = {
                general = "~/Documents/Norg/",
                life = "~/Documents/Norg/Life/",
                work = "~/Documents/Norg/Study/",
                journal = "~/Documents/Norg/Journal/",
              },
              default_workspace = "general",
            },
          },
        },
      })
```

### Actual behavior

When entering a journal entry using the :cmd command, it gets wiped. No correlation to template file. Just blank. 
Creating a new file works as expected, but entering it will wipe it. 

Works with every type of file that can be entered with Neorg journalc command. Entries, toc, template.

Also, cannot be reversed with “undo” command even with swap file options enabled in init.lua

https://github.com/nvim-neorg/neorg/assets/94762349/9c905f67-74cb-4329-ae2a-9038be9826fe

### Expected behavior

It would enter a file without modifying its content. And not wiping my 100 lines of enlightening text that can't be returned now.

### Steps to reproduce

Be on the latest Neorg release. 
Use journal module.
Reenter existing journal entry with a command `Neorg journal`
(entering with telescope or in file manager will not reproduce the issue)

### Other information

It bug may be caused by one of the yesterdays or today's commits. Because yesterday I could edit the same file several times using the cmd to enter it. 

### Help

No

pysan3:
I am very sorry. There was a bug in pathlib.nvim.

I'll fix it asap.

PR #1371, opened 2024-03-31
fix(dirman): respect force option in dirman create_file

pysan3:
I noticed that I ignored the `opts.force` flag in `dirman.create_file`.

Now respects `opts.force` to decide whether to create a new empty file, or, append to the file  to just update the timestamp.

I also had a bug in pathlib `self:touch` so I changed to use `fs_open` instead.
This is fixed in `pathlib.nvim ~> 2.2` so bumped the version as well just in case.

Fixes: https://github.com/nvim-neorg/neorg/issues/1365

### DIFF
--- a/.github/workflows/luarocks.yml
+++ b/.github/workflows/luarocks.yml
@@ -30,4 +30,4 @@ jobs:
             lua-utils.nvim == 1.0.2
             plenary.nvim == 0.1.4
             nui.nvim == 0.3.0
-            pathlib.nvim ~> 2.1
+            pathlib.nvim ~> 2.2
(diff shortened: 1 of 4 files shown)


pysan3:
@vhyrro BTW I need to change 3 files to update dependency version atm. 

I'm afraid of missing some place and pointing to a different version on these files. 

Do you have any good idea?

vhyrro:
I have been thinking of this.. perhaps we could use dhall as our source of truth that could then trickle down into all three files.

pysan3:
For build.lua, I think we should rather hardcode the version but make release-please update the version number on the fly. 

```lua
luarocks.ensure({
  "neorg == 0.0.0" -- x-release-please-version
})
```

In case of luarocks-release action, why does it not respect `*.rockspec` but we have to define the dependency in yaml in the first place?

I think the action should either generate `*.rockspec` and commit it or do not generate one on its own and rather fetch all values from existing `*.rockspec` in root.

pysan3:
@vhyrro (just in case you didn't receive a notification)

PR #1381, opened 2024-04-06
docs(README): Fix navigation links

kluen:
The links on the main page are broken.

I edited them so that they again mirror the page's structure. I placed the line break so that the layout looks similar to before. I kept the roadmap link as it was. I suppose it is still intended to be there?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,19 +15,15 @@ Your New Life Organization Tool - All in Lua
 •
 [Tutorial](#-tutorial)
 •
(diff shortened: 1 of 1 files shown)


pysan3:
Maybe `docs(README): ...`?

I'm not a maintainer tho ;)

kluen:
Sounds good and I found other PRs that use this format. Changed it, thanks!

vhyrro:
Thank you :D

Issue #465, opened 2022-05-22
Support relative paths for xdg-opening


mrossinek:
### Discussed in https://github.com/nvim-neorg/neorg/discussions/456

<div type='discussions-op-text'>

<sup>Originally posted by **esquires** May 14, 2022</sup>
From the documentation [here](https://github.com/nvim-neorg/neorg/blob/main/doc/neorg.norg#L166) I can form links using absolute paths to images to be opened with xdg-open but I can't seem to figure out the syntax for opening files relative to the current file. Is that possible?

I'd like to do something like this but neither seem to work:

```
{file:./mypicture.png}[mypicture]
{file:///.mypicture.png}[mypicture]
``` 

Can you point me to the right syntax for doing relative paths so that they open in an external program? It seems that the `@path` syntax works fine (e.g. for text files): `{@ myfile.txt}[myfile]`</div>

The problem appears to be, that the File URI scheme does not (officially) support relative paths. However, we should be able to support some sensible custom solution for this within Neorg.

BoManev:
bump

It's useful if you rely on images and other non-norg file to open using xdg-open.
```
{file:///abosulute/path} 
{file://:path:} # this can use the capabilities of the :path: target 
```

brglng:
Hope to see this feature.

pysan3:
Links to non norg files are written as `{/ path/to/file}`. You see the `/` at the front which means a normal file path.

Please read `{# File Location}` section in the specification for more detials. https://github.com/nvim-neorg/norg-specs/blob/51aab3dfd3bdf1d7ebdd2648d2e418ddd8d9ce2c/1.0-specification.norg#L1343

Absolute paths also works exactly the same with `{/ /etc/passwd}` or `{/ C:/Users/...}`.

brglng:
The `/ path/to/file` only works for text files.

benlubas:
@brglng what are you trying to open b/c this works perfectly fine for me

https://github.com/nvim-neorg/neorg/assets/56943754/4b8ba668-977a-4128-89e7-0febbbcc0a68

benlubas:
So it doesn't work the way you want for text files? 

There's the inbuilt `gx` command if you want to 'force' open a file externally, I think that it makes sense for neorg to open files that nvim can handle inside nvim, and only defer to `xdg-open` when it sees a file that it can't handle.

brglng:
[...]

PR #1382, opened 2024-04-06
hop to relative paths (file:./rel/path)

esquires:
This pull request adds a feature discussed here (https://github.com/nvim-neorg/neorg/discussions/456). Currently a user can open a file in an external program with `{file:///absolute/path/to/file}`. On my end where I put my notes in a repository that may end up on different machines with different paths, it is way more useful to be able to specify a relative path, e.g. `{file:./relative/path}`.

This pull request adds the ability to add `{file:./relative/path}`.

Example:

Put `mypic.png` in a directory somewhere and `foo.norg` in the same directory with the following:

```
{file:./mypic.png}
```

Place the cursor over the link and hit `<cr>`. It should open in an external program.

### DIFF
--- a/doc/neorg.norg
+++ b/doc/neorg.norg
@@ -158,7 +158,8 @@ version: 0.1
     - `:path:` (another norg file at a given path without specific target)
     - `/ path` (a non-norg file at a given path)
     - `https://github.com` (a URL)
-    - `file:///some/path` (any file, opened via `(xdg-)open`)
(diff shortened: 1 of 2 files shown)


esquires:
Looks like an issue (https://github.com/nvim-neorg/neorg/issues/465) was added in addition to the discussion.

benlubas:
@esquires `{/ ./foo/bar.png}` works perfectly fine for me

esquires:
Thanks! I'll go ahead and close this.

PR #1386, opened 2024-04-07
Add missing parameter to string format.

ethanbrews:
Added the missing `key` to `string.format` when logging a warning about attempting to load an unknown module.


vhyrro:
Thanks! Unsure how I missed that :)

PR #1390, opened 2024-04-13
Remove `nvim-treesitter` from dependency list

vhyrro:
This PR serves two purposes:
1. Remove `core.maneouvre` (an already deprecated module in need of a rewrite)
2. Remove `nvim-treesitter` from the dependency list

### Why?

`nvim-treesitter` will soon be dropping their `ts_utils` module in the `main` branch. To accommodate for this ahead of time, Neorg is removing its reliance on nvim-treesitter as a dependency. Not only this, the main branch of nvim-treesitter will no longer ship the `norg` parser due to its reliance on C++ as its scanner (something that is fixed in the V3 parser but not yet shipped within Neorg).

The biggest benefit of this change is loading order - since Neorg needs nothing from nvim-treesitter any longer, nvim-treesitter can load naturally and thus hopefully prevent highlighting errors that we've seen so far.

There is a hack in the codebase right now related to MacOS installation (due to old compilers), but it's one that will soon be removed. The norg tree-sitter parser is already published on luarocks in precompiled form for all major operating systems in a way that complements and does not interfere with nvim-treesitter. This rock can be pulled down as a dependency and can immediately work with nvim-treesitter's builtin highlighting module or the `rocks-treesitter.nvim` module.

### Testers Welcome

I encourage anyone to test this PR and report any errors you may encounter. This is marked as a breaking change but will not affect users, only external modules that may rely on `ts_utils` themselves :)

### TODOs

- [x] Remove `:Neorg sync-parsers` (issue a deprecation warning for users when they try to use it)
- [x] Remove automatic parser pinning logic
- [x] Rename `core.integrations.treesitter` to `core.treesitter` (no longer an integration but a component of the core)
- [x] Remove MacOS hacks by using precompiled parser


vhyrro:
Just one thing left: make `luarocks.nvim` use the rocks-binaries servers and precompile the parsers so that users needn't compile them on their machines. After that we're good to go! :D

vhyrro:
Update: this PR in of itself is complete, it seems like our CI on rocks-binaries is having some issues with compiling C++ code properly, and we haven't found a way to fix it yet. Once that happens, this PR can finally be merged! Should make Neorg more stable :)

Anrock:
@vhyrro is it possible to get this somehow to master without waiting for rocks-binaries CI? If I understand correctly that will lead to local treesitter compiling as the only option, am I right? 

I'm kinda stuck without neorg now, since neorg depends on nvim-treesitter and rocks, and nvim-treesitter is horribly broken with rocks for some reason.

vhyrro:
@Anrock seems like it's related to the actual compilation process itself which super sucks. I'm trying to fix it over here as fast as I can :)

vhyrro:
btw @Anrock in the meantime you can use the git version of treesitter with `:Rocks install rocks-git.nvim` and `:Rocks install nvim-treesitter/nvim-treesitter`. That should at least resolve the treesitter errors.

We're revamping the `luarocks-build-treesitter-parser` dependency to use `tree-sitter build` (the new, recommended way of building TS parsers). It will probably be complete in the next day or two.

cesorious:
Does this adresse usecases where `rocks-treesitter.nvim` is not used and `tree-sitter-norg` is installed directly through rocks?

vhyrro:
> Does this adresse usecases where `rocks-treesitter.nvim` is not used and `tree-sitter-norg` is installed directly through rocks?

Precisely :)

vhyrro:
Temporarily on hold as even more nix + luarocks-related problems are fixed :|

twilightty:
Any updates? I'm facing the same issue on macOS 14.

```
Failed to run `config` for neorg

...ua/neorg/modules/core/integrations/treesitter/module.lua:53: Unable to load nvim-treesitter.ts_utils :(
```

vhyrro:
Closing this in favour of a new PR that I'll make (less merge conflicts to deal with) but with roughly the same content.

A lot of things have improved since then, and it seems that tree-sitter-norg compilation is now fully successful on all platforms, including MacOS thanks to the work of Marc Jakobi :D

PR #1394, opened 2024-04-17
docs: emphasise luarock.nvim extra install steps

benlubas:
Documentation (readme) update to make the install instructions for luarocks.nvim more clear.

- links the repo (b/c it very hard to find by nature of it being a fork)
- make it more clear that `luarocks.nvim` has additional installation steps

_also, I haven't done it in this PR, but I think that we should add Pysan's written tutorial to the README._

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ See [this blog](https://vhyrro.github.io/posts/neorg-and-luarocks/) for more inf
 <details>
 <summary>Click for installation snippet.</summary>
 
-- Ensure you have `vhyrro/luarocks.nvim` installed:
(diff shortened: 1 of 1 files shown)


juniorsundar:
It's a +1 from me on this as well.

I find that for most configs luarocks.nvim is a one-off installation. So having to go to luarocks.nvim GitHub page to read the installation steps there and then Neorg can cause goofheads like me to fumble.

Since we are already showing that luarocks.nvim is a required dependency in Neorg install steps i think it's imperative to either:
- Explain the full install process for luarocks.nvim with it's dependencies as well, OR
- direct them to a expansive luarocks.nvim install procedure with a link.

*For the record I did not RTFM and FU'd the install...*

vhyrro:
Thank you :)

PR #1397, opened 2024-04-19
feat(latex): async image rendering

benlubas:
https://github.com/nvim-neorg/neorg/assets/56943754/d5f8e015-2cb3-4497-985d-ab9dca7f761a

Brings many many improvements to the latex renderer:

- Images are created and show asynchronously using nvim-nio
    - Creating images never blocks nvim, even with excessive amounts of images
- Images correctly position themselves in line (accounting for other virtual text, folds, etc.)
- Enable/Disable commands
- minimum length requirement to render. This allows you to type `$x$` and not have it render a massive disproportionate letter `x` for no good reason
- Better documentation
- Image caching
- Persistent images with auto update--before images would just disappear when you started typing
- Foreground color controlled by a highlight group
    - this updates and rerenders on colorscheme change
    - Default links to `Normal`
- Debounced rendering so it waits for you to finish typing, this is just to avoid an (even more) excessive amount of temp images being created.
    - Debouncing can make things look bad, it's completely configurable, default is 200ms, setting it to 10ms makes it near impossible to notice (I don't recommend setting it to less than this in case you're running a macro or something).

Unfortunately this clobbers all of the changes in #1372, but it improves on most of what's happening there I think. I started this work before that PR went up, and then I got stalled with school and other things.

---

TODO:

- [x] When images move, they flash. I know why that happens, I think I can fix it, but I don't consider it blocking for this PR
- [x] Switching buffers messes with things really heavily. All of the information that we store needs to be on a per-buffer level
- [x] Image.nvim doesn't automatically handle text concealing causing images to move
	- fixed by https://github.com/3rd/image.nvim/pull/156
- [x] Other image.nvim issues/inconcitencies
	- fixed by https://github.com/3rd/image.nvim/pull/159
- [x] image.nvim PRs need to be merged

### DIFF
--- a/lua/neorg/modules/core/latex/renderer/module.lua
+++ b/lua/neorg/modules/core/latex/renderer/module.lua
@@ -154,6 +154,7 @@ module.private.clear_extmark = function(key)
 end
 
 module.public = {
+    ---@async
     async_latex_renderer = function()
         ---node range to image handle
         -- module.private.ranges = {}
@@ -230,6 +231,7 @@ module.public = {
     end,
 
     ---Writes a latex snippet to a file and wraps it with latex headers to it will render nicely
+    ---@async
     ---@param snippet string latex snippet (if it's math it should include the surrounding $$)
     ---@return string temp file path
     async_create_latex_document = function(snippet)
@@ -253,6 +255,7 @@ module.public = {
     end,
 
     ---Returns a filepath where the rendered image sits
+    ---@async
     ---@param snippet string the full latex snippet to convert to an image
     ---@return string | nil
     async_generate_image = function(snippet)
@@ -361,7 +364,7 @@ local function render_latex()
     end
 
     if not render_timer then
-        render_timer = vim.loop.new_timer()
+        render_timer = vim.uv.new_timer()
     end
 
     render_timer:start(module.config.public.debounce_ms, 0, function()
--- a/lua/neorg/modules/core/latex/renderer/module.lua
+++ b/lua/neorg/modules/core/latex/renderer/module.lua
@@ -157,7 +157,8 @@ module.public = {
         local new_limages = {}
         for _, limage in pairs(module.private.latex_images[buf] or {}) do
             if limage.extmark_id then
-                local extmark = nio.api.nvim_buf_get_extmark_by_id(buf, module.private.extmark_ns, limage.extmark_id, {})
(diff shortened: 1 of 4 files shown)


lua/neorg/modules/core/latex/renderer/module.lua
```diff
@@ -108,116 +251,254 @@ module.public = {
...
+                running_proc = nio.run(
+                    function()
+                        module.public.async_latex_renderer(buf)
+                    end,
+                    vim.schedule_wrap(function()
+                        module.public.render_inline_math(module.private.latex_images[buf] or {}, buf)
+                        running_proc = nil
+                    end)
+                )
+            end)
```
vhyrro, line 398: Is there a reason for the doubly-wrapped schedule + async calls here?

I've been able to replicate the same behaviour with the following snippet:
```suggestion
            running_proc = nio.run(
                function()
                    nio.scheduler()
                    module.public.async_latex_renderer(buf)
                end,
                function()
                    module.public.render_inline_math(module.private.latex_images[buf] or {}, buf)
                    running_proc = nil
                end
            )
```

It doesn't seem to hurt performance nor introduce new bugs. The second function from nio.run is never run from api-fast, so there's no need for a schedule there. `nio.scheduler()` is a cleaner way of doing a whole `vim.schedule` wrap.

Since you're running from a timer callback and the operation itself is async I believe this only introduces visual latency with no extra benefits. Do let me know if I overlooked something :)

vhyrro:
Code generally looks great! I've managed to find another bug that may or may not be known: other virtual text seems to break positioning within the buffer (e.g. images generated by `.image`). I assume this is a bug that needs fixing on image.nvim's side, right?

benlubas:
I don't see that behavior, and we definitely account for that in image nvim. Is your image nvim on latest from GitHub?

vhyrro:
I'm on the latest stable luarocks release which I realized happens to be from 144 days ago :joy:, perhaps that's due for an update :sweat_smile:

benlubas:
I asked 3rd to cut a release, I think that triggers a luarocks build too, so you can try the rocks version again

vhyrro:
Thanks for getting that release out, that's fixed now at least :)

wilville:
so wait does this allow for display math where it would be bigger or does this only work for concealing inline math.

PR #1412, opened 2024-04-29
feat: TS autocomplete

benlubas:
Uses Treesitter to find autocompletion suggestions. Has a few impacts:
- we'll no longer incorrectly link to norg in a verbatim or like a code block or something
- now links to line targets like `inline <target>` with `{#|}` completions
- `{#|}` completions will correctly suggest footers and other targets that it wouldn't before

I just pulled the query from `core.esupports.hop`. We could maybe refactor that out, b/c currently completion requires hop, which is slightly weird. Lmk if I should do that and what helper module you'd like me to use/make.


vhyrro (approved):
LGTM! I think that the wisest way of handling `esupports.hop` is to extract the finding logic out into a separate module (e.g. `core.links`), then having `esupports.hop` do the actual hopping. Will you want to tackle that in a separate PR?

benlubas:
Blocked on #1428, but this PR is now updated so it can be merged right after #1428

vhyrro:
One conflict that needs fixing, you know what to do :p

PR #1413, opened 2024-05-03
feat: auto tangle

benlubas:
Adds the `tangle_on_write` option to the tangle module. Defaults to false, when true, call `:Neorg tangle current-file` on `BufWritePost`.

Fixes an annoying "bug" where tangling fails when code is de-dented. This is just a qol thing, I don't see a point in failing with a loud error in this case when we can easily handle the error and correctly export anyway.

### DIFF
--- a/lua/neorg/modules/core/integrations/treesitter/module.lua
+++ b/lua/neorg/modules/core/integrations/treesitter/module.lua
@@ -519,8 +519,9 @@ module.public = {
         return result
     end,
     --- Given a node this function will break down the AST elements and return the corresponding text for certain nodes
-    -- @Param  tag_node (userdata/treesitter node) - a node of type tag/carryover_tag
-    get_tag_info = function(tag_node)
+    --- @param tag_node TSNode - a node of type tag/carryover_tag
+    --- @param throw boolean - when true, throw an error instead of logging and returning on failure
+    get_tag_info = function(tag_node, throw)
         if
             not tag_node
             or not vim.tbl_contains(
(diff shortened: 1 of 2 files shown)


vhyrro:
Sorry for taking a while to get around to this. I generally disagree with the lenient tag checking (allowing arbitrary indentation), because it will cause some confusion amongst users. The reason all this behaviour exists in the first place is because you can't correctly "guess" where the start of the indentation is unless you have a strong rule like we do. Imagine someone does the following:
```
   @code
   -- This is correctly indented
this will get cut off
   @end
   @code 
-- This is incorrectly indented
   this is correctly indented visually, but will not show up as such in the tangle output
   @end
   @code 
  -- Oops, I accidentally dedented this line by one character, and so now everything else will be indented incorrectly.
   this has the correct indentation, but will now appear one character to the right in the tangle output
and this will be cut off
   @end
```

I'd much rather have a *single* case which can fail: "you've dedented too far". This is much less ambiguous and better UX than trying to guess what the user meant and then being unpredictable in our output :)

benlubas:
For most practical purposes tangled code need not be indented perfectly.

Python and haskal are the two exceptions. Could there be a list of languages to allow correction on? Can it be configurable? 

The main reason I changed the intent handling was to avoid annoying/excessive error messages while I was editing code with auto tangle. Because I use an auto save plugin, it was very easy to trigger those errors. 

We could also make it a diagnostic warning instead of error for languages other than Python and haskal. That way it's hard for users to unknowingly have an intent error.

vhyrro:
If the errors are annoying on auto-save, then I see two paths forward:
- enable lenient tag checking only when auto-tangle is enabled, otherwise always perform strict tag checking
- when auto-tangle is enabled, instead of erroring loudly, just print a message in the bottom left saying that tangling didn't succeed

I'm personally more fond of the second approach. I really feel like lenient tag checks are a source of confusion, especially since ill-indented code will produce garbled output. More convenience is not a good reason for invalid output.

benlubas:
@vhyrro I updated this PR with some customizable behavior for the error messages

vhyrro:
ty!

PR #1419, opened 2024-05-11
chore: mark finished clipboard todo

benlubas:
The clipboard TODO on the roadmap is finished as far as I can tell. Copying from a code block via a visual selection does remove the leading white space from the selection.


vhyrro:
I've found that the module is still buggy when used with visual block selection. Once that's fixed I'm happy to merge this :)

PR #1423, opened 2024-05-15
feat: add `integrations.coq_nvim` as a new completion engine

TheLeoP:


vhyrro (approved):
Awesome work, thank you!

PR #1428, opened 2024-05-17
chore: refactor links query into separate module

benlubas:
Pulls out a link query from esupports.hop into a `core.links` module, which only exists to house that one query at the moment.

The query will be used in the completion module in the TS autocomplete PR.

### DIFF
--- a/lua/neorg/modules/core/links/module.lua
+++ b/lua/neorg/modules/core/links/module.lua
@@ -2,6 +2,7 @@
     file: Links
     title: Find links/target in the buffer
     description: Utility module to handle links/link targets in the buffer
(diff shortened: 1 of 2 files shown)


lua/neorg/modules/core/links/module.lua
```diff
@@ -0,0 +1,94 @@
+--[[
+    file: Links
+    title: Find links/target in the buffer
+    description: Utility module to handle links/link targets in the buffer
```
vhyrro, line 4: Might be wise to mark this as `internal: true` :)

vhyrro (approved):

PR #1429, opened 2024-05-17
docs(wiki/kickstart): add alpine requirements

angelacastaneda:
had a lot of issues with neorg on alpine till @vhyrro helped discover an upstream gnu wget dependency. I tested with builds.sr.ht to ensure those dependencies would be enough with the kickstart config:

https://builds.sr.ht/~acsqdotme/job/1223933


vhyrro:
Thanks! :)

PR #1438, opened 2024-05-25
fix(latex-renderer): handle broken latex

benlubas:
Adds another call to yield to the scheduler so that we don't call a vim function when we're not supposed to.

It's only by shear luck that this didn't break on us all the time.

also, this commit was the problem: https://github.com/nvim-neorg/neorg/commit/2a9c3fab1bb6beabc4160264835be7f3b9a579e7

### DIFF
--- a/lua/neorg/modules/core/latex/renderer/module.lua
+++ b/lua/neorg/modules/core/latex/renderer/module.lua
@@ -403,13 +403,16 @@ local function render_latex()
         render_timer = nil
 
         if not running_proc then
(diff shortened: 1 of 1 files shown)


benlubas:
ahh wait this doesn't work either, I'm going back to my double wrapped schedules

benlubas:
ready for review

vhyrro:
Could you rebase on latest main and force push so that the CI stops complaining? Thanks :)

benlubas:
@vhyrro done

Issue #499, opened 2022-07-04
Foldtext bugged


jghauser:
### Prerequisites

- [X] I have read [contribution guidelines](https://github.com/vhyrro/neorg/blob/main/docs/CONTRIBUTING.md)
- [X] I am using the latest version of the plugin
- [X] I am using either 0.6 stable or the latest compiled neovim version

### Neovim Version

NVIM v0.7.2

### Neorg setup

require("neorg").setup({
	load = {
		["core.defaults"] = {}, -- Load all the default modules
		["core.norg.concealer"] = {}, -- Allows for use of icons
	},
})

### Actual behavior

When opening a fold, writing some text and then closing it, sometimes the foldtext becomes garbled.
![20220704_21h50m36s_grim](https://user-images.githubusercontent.com/10319377/177211051-53767424-d243-44f5-b802-f61b3f3517b6.jpeg)

### Expected behavior

Fold text displays normally

### Steps to reproduce

Create a file with folds, open them, add some text in the section, and close the fold again. It doesn't always happen, and I'm not quite sure what triggers it.

### Other information

I think it's an issue with the concealer as the problem doesn't show up when I disable that module.

### Help

Yes, but I don't know how to start. I would need guidance (check question below)

vhyrro:
Hey! This issue was caused by the concealer leaving leftover extmarks in certain scenarios (sometimes even triple the amount of necessary extmarks 👀). Because of this the foldtext algorithm broke in very weird ways when it received duplicates lol

Let me know if it's fixed on your end!

PR #1448, opened 2024-05-30
ref(math-renderer): modularize image generation to make way for typst

benlubas:
More or less what it says on the tin.

Extracts the latex specific image generation code into it's own module, and provides that module as configuration to the (now called) math-renderer. This opens up the possibility of other math renderers, the most notable being typst.

--- 

Down the road, the norg file format will hopefully have some standard way of specifying the math language, and then this module can be reworked again to take that into account (and possibly render typst and latex in the same document). But for now, user configuration that allows rendering either latex or typst seems like a good starting point.


vhyrro:
question: isn't this a breaking change, given the module name is now different?

benlubas:
yes, but the other one hasn't released yet

lua/neorg/modules/core/math/renderer/latex/module.lua
```diff
@@ -0,0 +1,131 @@
+--[[
+    file: Core-Math-Renderer-LaTeX
```
vhyrro, line 2: nitpick: just `Math-Renderer-LaTeX` will do :p

lua/neorg/modules/core/math/renderer/latex/module.lua
```diff
@@ -0,0 +1,131 @@
...
+    title: Convert LaTeX snippets into image files
+    summary: A module that provides images of LaTeX for `core.math.renderer`
+    ---
+
+The module is used by `core.math.renderer` to render math blocks as LaTeX.
+
+Requires:
+- `latex` command in path (you need to install a latex distribution)
+- `dvipng` this will come with the latex distribution
```
vhyrro, line 11: suggestion: incorporate the recommendations provided via #499?

lua/neorg/modules/core/math/renderer/module.lua
```diff
@@ -0,0 +1,483 @@
+--[[
+    file: Core-Math-Renderer
```
vhyrro, line 2: same here, just `Math-Renderer` will do :)

lua/neorg/modules/core/math/renderer/module.lua
```diff
@@ -0,0 +1,483 @@
...
+    ---
+
+This is an experimental module that requires nvim 0.10+. It renders math snippets as images making
+use of the image.nvim plugin. By default, images are only rendered after running the command:
+`:Neorg render-math`. Rendering can be disabled with `:Neorg render-math disable`
+
+Requires:
+- [image.nvim](https://github.com/3rd/image.nvim)
+
+A highlight group that controls the foreground color of the rendered math: `@norg.rendered.math`,
```
vhyrro, line 14: `@norg` -> `@neorg`

lua/neorg/modules/core/math/renderer/module.lua
```diff
@@ -0,0 +1,483 @@
...
+        requires = {
+            "core.integrations.image",
+            "core.integrations.treesitter",
+            "core.autocommands",
+            "core.neorgcmd",
+            "core.highlights",
+        },
+        wants = {
+            "core.math.renderer.latex",
+        },
```
vhyrro, line 35: question: unsure if this is desired. a module shouldn't want a specific integration.

lua/neorg/modules/core/math/renderer/module.lua
```diff
@@ -0,0 +1,483 @@
...
+                goto continue
+            end
+            module.private.image_api.render({ limage })
+            ::continue::
+        end
+    end,
+}
+
+local running_proc = nil
+local render_timer = nil
```
vhyrro, line 316: question: wouldn't these "global" variables cause issues if I open two side-by-side buffers?

vhyrro (commented):
Big fan of the changes! Just some things here and there.

lua/neorg/modules/core/math/renderer/module.lua
benlubas: I think the idea was that in the future it will want both latex and typst, so the user can choose which to load and use.

I'm not sure if I'm understanding the `wants` table correctly.

lua/neorg/modules/core/math/renderer/module.lua
benlubas: I think there were a few fixes that the latex renderer saw that this PR has not adopted yet. I will work on this a bit throughout the day today

lua/neorg/modules/core/math/renderer/module.lua
vhyrro: Wants simply looks if a module is loaded and errors if it doesn't find it. Maybe this module should try to `neorg.modules.load_module` the renderer provided in the public config?

lua/neorg/modules/core/math/renderer/module.lua
vhyrro: Awesome, looking forward to that

vhyrro:
I'll probably keep this for Neorg `10.0.0`. I'm worried that there'll be an overload of breakages in this release. One step at a time, as they say :)

PR #1453, opened 2024-06-04
More completions

benlubas:
Adds:
- Link names completions for links like:
    - `{:$/some/file:# Heading}[|` - suggests file title, file description, and "Heading"
    - `{:$/some/file:}[|` - suggests file title and description
    - and `{** local heading}[|` - suggests "local heading"
- Anchor completions (`[|`) with `[a1]{# something}` in the document will suggest "a1"
- additional detached modifier extensions for: `$`, `^`, `*`, and `~`

Refactors so TS helpers to take more sources, I'm slowly unifying the Neorg TS functions so they can all take buffer numbers, file paths, source strings, or nil for current buffer.

Fixes an off by one in get node text when the source was a string.

---

I think this wraps up all the completions. If there are more that I'm missing, let me know, but  I can't think of any new ones.. ~~Maybe it would be nice to get detached modifier extensions to work with more than just unordererd lists though~~

### DIFF
--- a/lua/neorg/modules/core/completion/module.lua
+++ b/lua/neorg/modules/core/completion/module.lua
@@ -547,7 +547,7 @@ module.public = {
             },
         },
         { -- foreign link name suggestions `{:path:target}[|]`
-            regex = "^(.*){:([^:]*):([^}]*)}%[",
+            regex = "^(.*){:([^:]*):[#$*%^]* ([^}]*)}%[",
 
             complete = module.private.foreign_link_names,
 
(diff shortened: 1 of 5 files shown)


max397574:
> it would be nice to get detached modifier extensions to work with more than just unordererd lists

+1 for that
was recently writing an ordered list with tasks which had to be done in the correct order and really missed the completions there

vhyrro (approved):
Code looks nice, am a fan. Thanks for the PR :)

PR #1456, opened 2024-06-07
fix: supress swap file errors

benlubas:
#1451 caught an error caused by swap files and produced an error message that heavily suggested the error was caused by swap and could be ignored when that was never confirmed. Some concern was raised about this approach, and an additional cases of the same bug was found...

So this PR checks the error message (via pattern match) to confirm that a swap error was the cause of `vim.cmd.edit` failing, and in these cases eats the error message. It's okay (and I'd say preferred) to eat this error b/c a lengthy swap message will pop up on screen for the user, so users will not be confused about what's happening.

In case the error is not caused be swap, it's now re-raised with the exact error message.

#1451 handled the edit call in `:Neorg index`, this pr additionally handles a call in the hop module when following a link that's opened in nvim.

### DIFF
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -53,13 +53,14 @@ local Path = require("pathlib")
 
 local neorg = require("neorg.core")
 local log, modules, utils = neorg.log, neorg.modules, neorg.utils
+local dirman_utils
 
 local module = modules.create("core.dirman")
 
 module.setup = function()
     return {
         success = true,
-        requires = { "core.autocommands", "core.ui", "core.storage" },
+        requires = { "core.autocommands", "core.ui", "core.storage", "core.dirman.utils" },
     }
 end
 
@@ -69,6 +70,8 @@ module.load = function()
         -- module.config.public.workspaces[name] = vim.fn.expand(vim.fn.fnameescape(workspace_location)) ---@diagnostic disable-line -- TODO: type error workaround <pysan3>
         module.config.public.workspaces[name] = Path(workspace_location):resolve():to_absolute()
     end
+
+    dirman_utils = module.required["core.dirman.utils"]
 
     modules.await("core.keybinds", function(keybinds)
         keybinds.register_keybind(module.name, "new.note")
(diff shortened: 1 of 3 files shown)


lua/neorg/modules/core/dirman/utils/module.lua
```diff
@@ -68,6 +68,22 @@ module.public = {
...
+    ---present. Re-raises other errors via log.error
+    ---@param path string
+    edit_file = function(path)
+        local ok, err = pcall(vim.cmd.edit, path)
+        if not ok then
+            -- Vim:E325 is the swap file error, in which case, a lengthy message already shows to
+            -- the user, and we don't have to crash out of this function (which creates a long and
+            -- misleading error message).
+            if err and not err:match("Vim:E325") then
+                log.error("Failed to edit file %s. Error:\n%s"):format(path, err)
```
vhyrro, line 82: Perhaps we may want to do one better than Neovim and issue an informative error message here? Something along the lines of "swap file detected, \<some short explanation of why this error happened>, read `:help swap-file` for more"?

lua/neorg/modules/core/dirman/module.lua
```diff
@@ -53,13 +53,14 @@ local Path = require("pathlib")
 
 local neorg = require("neorg.core")
 local log, modules, utils = neorg.log, neorg.modules, neorg.utils
+local dirman_utils
```
vhyrro, line 56: If this variable is only used once, then is there a point of creating it here? I'd say [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) in this case :p

lua/neorg/modules/core/dirman/utils/module.lua
benlubas: There's already a message like this whenever a swap file is found:

```
Found a swap file by the name ".main.c.swp" ~
	  owned by: mool   dated: Tue May 29 21:09:28 2001 ~
	 file name: ~mool/vim/vim6/src/main.c ~
	  modified: no ~
	 user name: mool   host name: masaka.moolenaar.net ~
	process ID: 12559 (still running) ~
While opening file "main.c" ~
	     dated: Tue May 29 19:46:12 2001 ~
~
(1) Another program may be editing the same file. ~
    If this is the case, be careful not to end up with two ~
    different instances of the same file when making changes. ~
    Quit, or continue with caution. ~
~
(2) An edit session for this file crashed. ~
    If this is the case, use ":recover" or "vim -r main.c" ~
    to recover the changes (see ":help recovery"). ~
    If you did this already, delete the swap file ".main.c.swp" ~
    to avoid this message. ~
```

lua/neorg/modules/core/dirman/module.lua
benlubas: I just think it's so much cleaner this way, and I just do that everywhere.

+ smaller diffs in the future if you are to add something that uses `dirman_utils` and want to pull it into a variable then.

You've also already merged code that does this :P

PR #1468, opened 2024-06-14
fix: insert template at the end of a new note

Dynge:
I noticed this small bug as I've begun using journal entries in my notes.

I believe the problem occurs because I've set metagen.type = "auto".

This fixes a bug where metadata is injected into the newly created file before the template contents. Which results in the template content being placed above the metadata.

By changing the 0 to a $ we always insert the template at the end of the file. I believe this is safe as this only occurs on newly created journal entries, which should therefore always be empty (except for some possible metadata injected by metagen).


vhyrro:
Looks good to me! Can't think of a place where this would break either.

Issue #1471, opened 2024-06-17
Add support for periodic notes


maxkuzn:
### Issues

- [X] I have checked [existing issues](https://github.com/vhyrro/neorg/issues?q=is%3Aissue) and there are no existing ones with the same request.

### tl;dr
Add support for various types of periodic notes. Currently, only daily notes (a.k.a. journal) are supported.

### Types of Periodic Notes
- Daily (currently implemented using the "core.journal" module)
- Weekly
- Monthly
- Quarterly
- Yearly
- Custom

#### Custom Notes
It would be beneficial to allow custom periodic notes, such as 2-week periods. See my use case for further details.

### Implementation Idea
All periodic notes could be implemented as presets of custom periodic notes. This means reorganizing the current support for daily journal notes into a more flexible system that supports custom periods. After this, presets for core periodic notes (such as daily, weekly, monthly, etc.) can be added for ease of use.

### My Use Case
I use notes for both my professional and personal life. Here is how I organize them:

- **Personal Life:**
  - Yearly notes
  - Monthly notes
  - Weekly notes
  - Sometimes daily notes

- **Job:**
  - Quarterly notes
  - Sprint notes (2-week period)
  - Daily notes

This feature is inspired by the ["periodic notes" plugin](https://github.com/liamcain/obsidian-periodic-notes) for Obsidian.

---

By adding support for these additional types of periodic notes and implementing them as presets of custom notes, users will have more flexibility in how they organize and manage their notes, catering to both personal and professional needs. This approach not only simplifies the implementation but also enhances user experience by providing ready-to-use templates for common periodic notes.

--- 

P.s. It would be great to have the ability to generate links between periodic notes. For example, links from a weekly note to all 7 daily notes and from each daily note back to the weekly note. However, this might be better suited as a separate feature request.

### Help

Yes, but I don't know how to start. I would need guidance

benlubas:
I think you could implement this with the current journal + the dynamic templates plugin and a few auto commands, it would just require a decent bit of logic (that could be extracted into its own neorg module). The thing you would need a neorg module is the custom commands.

Becuase this feature should use dynamic templates, I think it should be its own external module that makes use of [neorg-templates](https://github.com/pysan3/neorg-templates) rather than an internal module.

benlubas:
I have time today, I'll take a shot at implementing it

vhyrro:
[...]

PR #1474, opened 2024-06-22
feat!(journal): periodic & multiple journals

benlubas:
implements #1471

There's a few things to note here:
- looking for feedback on:
    - functionality--does this cover all the cases we want? Does this ship the defaults that we want?
    - user experience/configuration--anything else that should be configurable? Should each journal be able to have its own folder? or should they continue to be in the same overarching folder, with configurable subfolders/the option to not be in a subfolder?
- TOC code is untouched and untested, I think forcing each journal into its own subfolder would make the TOC implementation easier, other suggestions on how to handle it would be welcome
- The breaking change is pretty minimal right now. Templates are now configured differently, and templates files will have a new name. Instead of configuring with `template_name = template.norg` it's now `base_template_name = "template"` and templates files will be created on a per-journal basis like: `template-daily.norg`

### DIFF
--- a/lua/neorg/modules/core/journal/module.lua
+++ b/lua/neorg/modules/core/journal/module.lua
@@ -4,87 +4,118 @@
     description: The journal module allows you to take personal notes with zero friction.
     summary: Easily track a journal within Neorg.
     ---
-The journal module exposes a total of six commands.
-The first three, `:Neorg journal today|yesterday|tomorrow`, allow you to access entries
-for a given time relative to today. A file will be opened with the respective date as a `.norg` file.
+Journals are "periodic" in Neorg. You can have a daily journal, a weekly journal, a quarterly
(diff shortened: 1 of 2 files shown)


benlubas:
@maxkuzn if you have feedback on the features here that would be great

vhyrro:
With core.tempus now being the defacto (since 0.10 is required) I can start reviewing this sooner than later. I took a general glance at it and the execution looks solid.

benlubas:
Had a look at this again. I think the best solution for the `journal toc` commands will be to make them pattern match for file names so that you get a toc specific to each journal.

I'm fairly sure it's possible to transform the `path_format_strategy` into a lua pattern iff the `path_format_strategy` is a string. When that value is a function (which is allowed) we would have to also ask the user to provide some type of filter function that takes a path and tells us if it's a part of the journal.

I think this is a fairly reasonable approach. Let me know if you have thoughts @vhyrro

PR #1478, opened 2024-06-24
More ToC QoL features/fixes

### LINKED ISSUE #1475: Table of Content QOL options
### Issues

### Feature description

I'd like a few new (optional) features for the table of contents:
- Auto open TOC on norg buf enter
	- optionally place your cursor in the TOC on open 
- Auto close TOC when its buffer is closed
	- and exit nvim when TOC is the last open window (similar to how help docs work)

These two would be configurable separately, and are not a package deal, you can have one without the other.

### Help

Yes

### Implementation help

I don't need help :P

benlubas:
closes #1475 

### features:
- configurable max width
- automatically open/enter ToC on neorg buf enter
- close ToC when norg buffer is closed
- exit nvim when ToC is the last window (similar to how help windows function)
- set `winfixbuf` on the ToC window

These are all configurable behaviors.

### fixes:
- when a ToC is open and it's norg buffer has been removed, it will no longer throw an error when you try to jump to a heading, instead it opens a new buffer, re-sets the ToC width, and placed your cursor in the right spot.
- opening a ToC from within a ToC window will no longer break the window

---

There's also a random completion documentation fix in here b/c I'm lazy and don't want to open a separate PR.

### DIFF
--- a/lua/neorg/modules/core/completion/module.lua
+++ b/lua/neorg/modules/core/completion/module.lua
@@ -20,6 +20,8 @@ Completions are provided in the following cases (examples in (), `|` represents 
 - file path + header links (`{:path:*|`)
 - file path + fuzzy header links (`{:path:#|`)
 - file path + footnotes (`{:path:^|`)
+- anchor names (`[|`)
+- link names (`{<somelink>}[|`)
 
 Header completions will show only valid headers at the current level in the current or specified file. All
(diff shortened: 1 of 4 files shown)


benlubas:
Found a bug relating to closing/opening windows 'at the same time' I'll take a look tomorrow

vhyrro (approved):

PR #1486, opened 2024-06-26
fix(summary): ordering and indentation of nested entries

nuhakala:
Current state sometimes mixes entries with nested categories. For example, if there is three items:
- C with category cat
- A with category cat.sub
- B with category cat.sub

The produced summary looks like this:
```
* Index
** Cat
*** Sub
   - A
   - B
   - C
```
even though I want C to be directly under `Cat`. With multiple notes this causes nested categories to be rather messy sometimes.

I fixed this by simply introducing few temporary variables and adding nested categories to the summary at the end.

### DIFF
--- a/lua/neorg/modules/core/summary/module.lua
+++ b/lua/neorg/modules/core/summary/module.lua
@@ -187,17 +187,18 @@ module.load = function()
                 local starting_prefix = string.rep("*", heading_level)
 
                 local function add_category(category, data, level)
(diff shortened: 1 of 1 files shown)


vhyrro:
Code looks good! I'll also play around with the implementation later today to make sure it all works :)

vhyrro:
Seems to work well for me, thanks for the PR!

PR #1487, opened 2024-06-26
fix(promo): promote/demote range + keybinds

benlubas:
what it says on the tin. These were broken/removed a while ago. Presumably due to concealer changes but I'm not sure. Someone asked about bringing them back today, so here they are.

The re-indentation code in this module seems to struggle a little though. Not sure what's up with that.

### DIFF
--- a/lua/neorg/modules/core/esupports/indent/module.lua
+++ b/lua/neorg/modules/core/esupports/indent/module.lua
@@ -145,7 +145,7 @@ module.public = {
         end
 
         local leading_whitespace = line:match("^%s*"):len()
-        return vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
+        vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
     end,
 }
 
--- a/lua/neorg/modules/core/esupports/indent/module.lua
+++ b/lua/neorg/modules/core/esupports/indent/module.lua
@@ -122,6 +122,31 @@ module.public = {
 
         return new_indent
     end,
+
+    ---re-evaluate the indent expression for each line in the range, and apply the new indentation
+    ---@param buffer number
+    ---@param row_start number 0 based
+    ---@param row_end number 0 based exclusive
+    reindent_range = function(buffer, row_start, row_end)
+        for i = row_start, row_end - 1 do
+            local indent_level = module.public.indentexpr(buffer, i)
+            module.public.buffer_set_line_indent(buffer, i, indent_level)
+        end
+    end,
+
+    ---Set the indent of the given line to the new value
+    ---@param buffer number
+    ---@param start_row number 0 based
+    ---@param new_indent number
+    buffer_set_line_indent = function(buffer, start_row, new_indent)
+        local line = vim.api.nvim_buf_get_lines(buffer, start_row, start_row + 1, true)[1]
+        if line:match("^%s*$") then
+            return
+        end
+
+        local leading_whitespace = line:match("^%s*"):len()
+        return vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
+    end,
 }
 
 module.config.public = {
--- a/lua/neorg/modules/core/promo/module.lua
+++ b/lua/neorg/modules/core/promo/module.lua
@@ -79,6 +79,34 @@ module.private = {
         "generic_list",
         "quote",
     },
+
+    get_line = function(buffer, target_row)
+        return vim.api.nvim_buf_get_lines(buffer, target_row, target_row + 1, true)[1]
+    end,
+
+    buffer_set_line_indent = function(buffer, start_row, new_indent)
+        local line = module.private.get_line(buffer, start_row)
+
+        if line:match("^%s*$") then
+            return
+        end
+
+        local leading_whitespace = line:match("^%s*"):len()
+
+        return vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
+    end,
+
+    reindent_range = function(buffer, row_start, row_end)
+        local indent_module = modules.get_module("core.esupports.indent")
+        if not indent_module then
+            return
+        end
+
+        for i = row_start, row_end - 1 do
+            local indent_level = indent_module.indentexpr(buffer, i)
+            module.private.buffer_set_line_indent(buffer, i, indent_level)
+        end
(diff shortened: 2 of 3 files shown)


lua/neorg/modules/core/promo/module.lua
```diff
@@ -320,38 +320,14 @@ module.on_event = neorg.utils.wrap_dotrepeat(function(event)
         local end_pos = vim.api.nvim_buf_get_mark(event.buffer, ">")
 
         for i = 0, end_pos[1] - start_pos[1] do
-            module.public.promote_or_demote(event.buffer, "promote", start_pos[1] + i)
-        end
-
-        if modules.loaded_modules["core.concealer"] then
```
mtrajano, line 326: Is this no longer needed?

mtrajano:
Looks to be working fine for me. Tried indenting several nested lists with no problem, what issues did you run into re: the indentation?

lua/neorg/modules/core/promo/module.lua
benlubas: Yeah that event doesn't seem to exist anymore, and the concealer doesn't seem to break when promoting/demoting a range.

vhyrro:
@mtrajano the issue stems from e.g. paragraphs. It seems that the indentation is cumulative: with the following snippet:
```
* One
** Two
     content.
```

Doing a ranged promotion indents the paragraph twice as far:
```
** One
*** Two
              content.
```

This isn't the absolute end of the world, but would be nice to have fixed somehow :thinking:

benlubas:
@mtrajano  @vhyrro

Now this behaves itself a little better. There is still one behavior that might be unexpected. If you promote:
```
* heading
  Normal text
```
Normal text is correctly indented. But if you promote:
```
normal text
- list item
```
Then "normal text" is indented, while "list item" is promoted. However, this case:

```
- list

new paragraph
- new list
```
"new paragraph" is not indented.

---

I think this behavior is fine. It allows visual `>`/`<` to still work like normal in norg documents when you don't highlight anything promote-able. And if you only highlight the things you want to promote without extra paragraphs, there's no problem with it.

lua/neorg/modules/core/promo/module.lua
```diff
@@ -285,22 +287,11 @@ module.public = {
         end
 
         if not reindent_children then
+            print("hi")
```
mtrajano, line 290: hello

mtrajano:
Changes make sense to make, there's just a random print statement there

lua/neorg/modules/core/promo/module.lua
```diff
@@ -79,6 +79,34 @@ module.private = {
...
+        if line:match("^%s*$") then
+            return
+        end
+
+        local leading_whitespace = line:match("^%s*"):len()
+
+        return vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
+    end,
+
+    reindent_range = function(buffer, row_start, row_end)
```
mtrajano, line 99: Probably outside of the scope of this pr but I feel like this functionality probably belongs in the `indent` module. It can be responsible for properly reindenting the line/range instead of just querying it for the indent level

mtrajano (approved):

lua/neorg/modules/core/promo/module.lua
benlubas: I put it in this PR

lua/neorg/modules/core/promo/module.lua
```diff
@@ -189,7 +171,7 @@ module.public = {
             local current_visual_indent = vim.fn.indent(row + 1)
             local new_indent = math.max(0, current_visual_indent + n_space_diff)
```
mtrajano, line 172: This also feels internal to the indent module. Shouldn't have to pass it an indent, just a bufnr and a range/line number and it should be able to figure it out

lua/neorg/modules/core/promo/module.lua
```diff
@@ -28,17 +28,21 @@ local neorg = require("neorg.core")
 local modules = neorg.modules
 
 local module = modules.create("core.promo")
+local indent
```
mtrajano, line 31: can be part of `module.private`, not sue what the convention around this is

lua/neorg/modules/core/promo/module.lua
mtrajano: Nice looks much cleaner, hooray for encapsulation

lua/neorg/modules/core/esupports/indent/module.lua
```diff
@@ -122,6 +122,31 @@ module.public = {
...
+            local indent_level = module.public.indentexpr(buffer, i)
+            module.public.buffer_set_line_indent(buffer, i, indent_level)
+        end
+    end,
+
+    ---Set the indent of the given line to the new value
+    ---@param buffer number
+    ---@param start_row number 0 based
+    ---@param new_indent number
+    buffer_set_line_indent = function(buffer, start_row, new_indent)
```
mtrajano, line 141: I would rename to `reindent_line` just to be consistent with the above, and rename the `start_row` to `line`

mtrajano (approved):

lua/neorg/modules/core/promo/module.lua
benlubas: I like this convention, vhyrro and I have gone back and forth on this in the past as well.

vhyrro:
@benlubas the new behaviour is much better than what we had previously, so that sounds good enough for the time being. Running out of time today, but I'll revisit this PR tomorrow :)

lua/neorg/modules/core/promo/module.lua
benlubas: If we do this, normal `>>`/`<<` as well as visual `>`/`<` will lose their ability to move text around at all. So I don't think we should do that.

lua/neorg/modules/core/esupports/indent/module.lua
```diff
@@ -122,6 +122,31 @@ module.public = {
...
+        local line = vim.api.nvim_buf_get_lines(buffer, start_row, start_row + 1, true)[1]
+        if line:match("^%s*$") then
+            return
+        end
+
+        local leading_whitespace = line:match("^%s*"):len()
+        return vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
```
vhyrro, line 148:
```suggestion
        vim.api.nvim_buf_set_text(buffer, start_row, 0, start_row, leading_whitespace, { (" "):rep(new_indent) })
```

nitpick: return statement looks to be extraneous.

vhyrro (commented):
Just a single nitpick, other than that no other issues :)

PR #1490, opened 2024-06-27
feat(intro-spector): recursively search for TODO items

benlubas:
Before:
image

After:
image

There is now the weird behavior with nested lists that I point out in the example above. I still think this is an improvement, but feedback is of course welcome.

### DIFF
--- a/lua/neorg/modules/core/todo-introspector/module.lua
+++ b/lua/neorg/modules/core/todo-introspector/module.lua
@@ -49,9 +49,9 @@ function module.public.attach_introspector(buffer)
 
     module.required["core.integrations.treesitter"].execute_query(
         [[
-    (_
-      state: (detached_modifier_extension)) @item
-    ]],
+            (_
+              state: (detached_modifier_extension)) @item
+        ]],
         function(query, id, node)
             if query.captures[id] == "item" then
                 module.public.perform_introspection(buffer, node)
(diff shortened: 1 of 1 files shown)


max397574:
do we even want this?
because if you have a normal list (with nested items or not) it will just show you how many of these items are completed
but now afaict it will e.g. for headings show how many nested things (e.g. todo list in the headings) are completed

vhyrro:
I'm personally opposed to this (and hence why the implementation is as it stands today) - there's too much semantic fuzziness when you start mixing all TODO states with each other. This will be even more true once the spec is revised to allow lists to have more complex content inside. For instance, does an undone definition contribute to the state of a list item? If I have several TODO lists in a heading, do I really want to see `[0/928]` at the top?

I find this to make sense for the simple case, but for larger notes you'd start to see a thicket of introspector hints fairly quickly. Let's keep the behaviour as it is currently :)

benlubas:
Can we give it a configurable depth?I like when a heading has `[n/x]` for all its TODO items.

Default to 1 so it behaves the same as it does now. But allow people to set it how they want?

vhyrro:
The depth is essentially always 1 currently, as the introspector looks for immediate children of the same type (e.g. all child headings) to get its information. This happens recursively, breadth-first iirc, so the whole thing becomes an inverted, tree-like fold.

An option could be to have a flag that allows for looking deeper and without these checks, but doing so feels off to me. I'd rather keep the existing behaviour. I feel like we underutilize the module system's capabilities of hotswapping modules. Perhaps it would just be wiser to create an external module with more lenient behaviour that hotswaps this one?

Anrock:
> For instance, does an undone definition contribute to the state of a list item? If I have several TODO lists in a heading, do I really want to see [0/928] at the top?

TBH sounds useful for my workflow. I often write big docs/specifications which are gradually refined and expanded with time and seeing which tree still has todo marks at a glance on top level is useful to see overall state of I-don't-know-ingness of current doc and what next experiment should be about.

max397574:
> TBH sounds useful for my workflow. I often write big docs/specifications which are gradually refined and expanded with time and seeing which tree still has todo marks at a glance on top level is useful to see overall state of I-don't-know-ingness of current doc and what next experiment should be about.

but for that it doesn't need to be recursive
it's enough if it's not 100% imo
and as vhyrro pointed out this could perhaps be implemented externally because imo it doesn't make sense like this for most useres

benlubas:
> Perhaps it would just be wiser to create an external module with more lenient behaviour that hotswaps this one

Could do this but it would just be needlessly duplicating code I think. If that's what you'd prefer I can do that

benlubas:
going to close this b/c it's stale, and honestly I've come to agree that recursively searching here is more trouble than it's worth

PR #1492, opened 2024-06-28
feat: List toggling/inverting for sublists

mtrajano:
First pass. Might still need some cleaning up but toggling / inverting sublists is working. I can also write some tests for this, what is the status on testing? Do we have this set up already?

Demo:

https://github.com/nvim-neorg/neorg/assets/4333381/e37dc621-7765-4399-b734-1d79ced0deb2

### DIFF
--- a/lua/neorg/modules/core/pivot/module.lua
+++ b/lua/neorg/modules/core/pivot/module.lua
@@ -35,47 +35,94 @@ module.load = function()
     vim.keymap.set("", "<Plug>(neorg.pivot.list.invert)", lib.wrap(module.public.change_list, true))
 end
 
+module.private = {
+    --- Return current node we are on, accounting for possible root of list
+    ---@param bufnr integer
+    ---@return TSNode?
+    get_current_node = function(bufnr)
+        local cursor = vim.api.nvim_win_get_cursor(0)
+        local node = module.required["core.integrations.treesitter"].get_first_node_on_line(bufnr, cursor[1] - 1)
+
+        -- if on root of the list we are actually interested in the first list item not the generic_list node
+        if node and node:type() == "generic_list" then
+            node = node:child(0)
+        end
+
+        return node
+    end,
+
+    ---@param node TSNode
+    ---@return TSNode?
+    get_parent_list = function(node)
+        local parent = node:parent()
+
+        return module.required["core.integrations.treesitter"].find_parent(parent, {
+            "generic_list",
+            "unordered_list1",
+            "unordered_list2",
+            "unordered_list3",
+            "unordered_list4",
+            "unordered_list5",
+            "unordered_list6",
+            "ordered_list1",
+            "ordered_list2",
+            "ordered_list3",
+            "ordered_list4",
+            "ordered_list5",
+            "ordered_list6",
+        })
+    end,
+
+    --- Returns the prefix the current list node should be toggled to
+    ---@param node TSNode
+    ---@return string
+    get_target_prefix = function(node)
+        local type = node:type():match("^un") and "~" or "-"
+        local level = tonumber(node:type():match("ordered_list(%d)")) or 0
+
+        return type:rep(level)
+    end,
+}
+
 module.public = {
     ---@param invert boolean
     change_list = neorg.utils.wrap_dotrepeat(function(invert)
         local buffer = vim.api.nvim_get_current_buf()
-        local cursor = vim.api.nvim_win_get_cursor(0)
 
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/integrations/treesitter/module.lua
```diff
@@ -616,22 +616,34 @@ module.public = {
...
-                    return _node
-                end
-            elseif vim.tbl_contains(types, _node:type()) then ---@diagnostic disable-line -- TODO: type error workaround <pysan3>
-                return _node
+    ---@param node TSNode? #The node to start at
+    ---@param types table|string #Will attempt to pattern match any of the types present in the table/string.
+    ---@param match_self boolean #If true can match on self, if false needs to match on ancestor to node
+    ---@return TSNode?
+    -- TODO: if remove match_self param, need to check other functionality that depends on this
+    find_parent = function(node, types, match_self)
```
mtrajano, line 624: Not sure if this was intended functionality but the previous version of this method allowed to match on the node passed so I had to include this param to stricly match on parent. The default is to keep the previous behavior (match_self = true).

lua/neorg/modules/core/integrations/treesitter/module.lua
```diff
@@ -616,22 +616,34 @@ module.public = {
...
+
+        if not node then
+            return
+        end
+
+        if not match_self then
+            node = node:parent()
+        end
+
+        if type(types) == "string" then
```
mtrajano, line 635: Simplified the method so we're only working with one type of param (table) otherwise should behave the same way. For tables it also used to check for equality whereas now it does pattern matching. If a string is passed that is not a pattern it should behave the same way

lua/neorg/modules/core/integrations/treesitter/module.lua
```diff
@@ -616,22 +616,34 @@ module.public = {
         return tree:root()
     end,
     --- Attempts to find a parent of a node recursively
-    ---@param node userdata #The node to start at
-    ---@param types table|string #If `types` is a table, this function will attempt to match any of the types present in the table.
-    -- If the type is a string, the function will attempt to pattern match the `types` value with the node type.
-    find_parent = function(node, types)
-        local _node = node
```
mtrajano, line 623: not really sure what this local _node was doing so I removed it. Also added the TSNode typehint to get rid of the ignore diagnostic comments cc: @pysan3

lua/neorg/modules/core/pivot/module.lua
```diff
@@ -38,43 +82,38 @@ end
...
         if not node then
             log.error("No node found under the cursor! Make sure your cursor is in a list.")
             return
         end
 
-        node = module.required["core.integrations.treesitter"].find_parent(node, "^generic_list$")
+        local parent = module.required["core.integrations.treesitter"].find_parent(node, {
+            "^unordered_list%d$",
+            "^ordered_list%d$",
+            "^generic_list$", -- for level 1 lists the parent is a generic_list
```
mtrajano, line 95: because `find_parent` can now accept patterns we can now change [this](https://github.com/nvim-neorg/neorg/blob/main/lua/neorg/modules/core/esupports/indent/module.lua#L249) to also be a list of pattern instead of an exhaustive list

lua/neorg/modules/core/pivot/module.lua
```diff
@@ -38,43 +82,38 @@ end
...
+        if not parent then
             log.error("No list found under the cursor! `toggle-list-type` works only for lists.")
             return
         end
 
-        local first_child = node:iter_children()()
-
-        if not first_child then
-            return
-        end
```
mtrajano, line 62: not really necessary due to the previous checks on node and it's parent existing

vhyrro:
Great work, unfortunately the refactor of the find_parent function seems to have broken quite a lot of functionality that relied on it. An example is link resolution: pressing <CR> on a simple link will cause the link resolver to fail.

I don't think I have the time to track down the source of the issue for this release, but I can sure come back to this and we can try to figure it out :)

mtrajano:
Hey @vhyrro thank you for the heads up I can look into it and clean this pr up a bit. I'll ping her once again once I get that out.

mtrajano:
@vhyrro updated the previous pr. Essentially kept the scope of changes to just the pivot module to prevent the other breaking changes. Open to any feedback/improvements you may have

vhyrro:
Fantstic, thank you! I've been unable to break it in my testing and it works much better than the current implementation :)

PR #1493, opened 2024-06-29
chore: refactor a few functions to allow for an external LSP module

benlubas:
@vhyrro  this is the PR that I said wouldn't make a lot of sense. But these refactors don't break anything (I checked all the places where `expand_pathlib` is used, they're all fine), and they allow [this](https://github.com/benlubas/neorg-interim-ls) to work.

For context, that module used to be a PR to core, but I've since pivoted to making it it's own external module.

### DIFF
--- a/lua/neorg/modules/core/dirman/utils/module.lua
+++ b/lua/neorg/modules/core/dirman/utils/module.lua
@@ -18,10 +18,18 @@ local module = neorg.modules.create("core.dirman.utils")
 ---@class core.dirman.utils
 module.public = {
     ---Resolve `$<workspace>/path/to/file` and return the real path
-    ---@param path string|PathlibPath # path
-    ---@param raw_path boolean? # If true, returns resolved path, otherwise, returns resolved path and append ".norg"
-    ---@return PathlibPath? # Resolved path. If path does not start with `$` or not absolute, adds relative from current file.
-    expand_pathlib = function(path, raw_path)
+    ---@param path string | PathlibPath # path
+    ---@param raw_path boolean? # If true, returns resolved path, otherwise, returns resolved path
+    ---and append ".norg"
+    ---@param host_file string | PathlibPath | nil file the link resides in, if the link is
+    ---relative, this file is used instead of the current file
+    ---@return PathlibPath?, boolean? # Resolved path. If path does not start with `$` or not absolute, adds
+    ---relative from current file.
+    expand_pathlib = function(path, raw_path, host_file)
+        local relative = false
+        if not host_file then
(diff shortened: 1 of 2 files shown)


lua/neorg/modules/core/dirman/utils/module.lua
```diff
@@ -66,7 +75,7 @@ module.public = {
             end
             filepath = filepath:add_suffix(".norg")
         end
-        return filepath
+        return filepath, relative
```
vhyrro, line 78: question: since a `PathlibPath` is being returned from this function, why not just use `:is_relative()` instead in your own code? It's true that this function can return a raw path as well, but it makes more sense to get back a non-raw path instead, do the checks you want, and then `:tostring()` the object to get the raw path.

Doing it this way adds more clutter to the function definition which isn't really needed.

lua/neorg/modules/core/integrations/treesitter/module.lua
```diff
@@ -231,31 +231,51 @@ module.public = {
...
+
+    search_tree = function(tree, node_type)
+        local result = {}
+        local root = tree:root()
+
+        --- Recursively searches for a node of a given type
+        ---@param node userdata #The starting point for the search
+        local function descend(node)
+            -- Iterate over all children of the node and try to match their type
+            for child, _ in node:iter_children() do ---@diagnostic disable-line -- TODO: type error workaround <pysan3>
```
vhyrro, line 263: suggestion: while we're at it, give this boy an upgrade, then you can remove the type error workaround a few line lower:
```suggestion
        ---@param node TSNode #The starting point for the search
        local function descend(node)
            -- Iterate over all children of the node and try to match their type
            for child, _ in node:iter_children() do
```

vhyrro (commented):
Just some questions and suggestions :)

lua/neorg/modules/core/dirman/utils/module.lua
benlubas: The nature of expand path is to give you a full absolute path back. In this case, `:is_relative()` is always `false`.

vhyrro: Right, I see how this is supposed to flow now. That's fair enough, sounds good.

vhyrro (approved):

lua/neorg/modules/core/dirman/utils/module.lua
benlubas: ah, little bit of testing later. `$/some_file` is reported relative by pathlib, but is not considered relative for neorg's purposes.

PR #1495, opened 2024-06-30
fix(text-objects): node selection inconsistencies

benlubas:
Fixes an issue discussed in discord.

<details>
<summary>Bug Reproduction</summary>

1. Open the lua below as a file
2. source the file

3. `viH`

4. observe that the first line is not highlighted in full, just the newline is highlighted

5. `ggo<esc>dd` (add and remove a line below the first one

6. `viH` again

7. Observe that the first line is highlighted in full

```lua
-- this is some comment text
-- It's going to be highlighted
-- source this file to see the issue in action

-- don't alter that comment text, it's the target of highlights

vim.keymap.set({"o", "x"}, "iH", function()
    vim.api.nvim_buf_set_mark(0, "<", 1, 28, {})
    vim.api.nvim_buf_set_mark(0, ">", 3, 46, {})

    vim.cmd("normal! gv")
end)
```

</details>

**This PR does:**
- Use the method from nvim-treesitter's TS-utils module, which doesn't suffer from the same bug
- trims the ranges to behave a little better. 
	- Will no longer select the newline after the heading for `heading.inner`
	- Will no longer error on trying to select a heading at the end of a document

### DIFF
--- a/lua/neorg/modules/core/text-objects/module.lua
+++ b/lua/neorg/modules/core/text-objects/module.lua
@@ -202,10 +202,26 @@ local function highlight_node(node)
     end
 
     local range = module.required["core.integrations.treesitter"].get_node_range(node)
+    if range.column_end == 0 then
+        range.row_end = range.row_end - 1
+        range.column_end = vim.api.nvim_buf_get_lines(0, range.row_end, range.row_end + 1, true)[1]:len()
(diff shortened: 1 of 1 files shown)


vhyrro (approved):
Works on my end, thanks :)

PR #1500, opened 2024-07-05
Take two on the ToC changes

benlubas:
I fixed the bug that I found in the last PR that still got merged despite the bug :P 

### Also in this PR:
- `fixed_width` option for the ToC window
- `create_vsplit` refactor/update to make use of the new `nvim_open_win` functionality in 0.10.0
- Fixed logic for computing the ToC width
- You can configure the `toc` command to just open but not enter the window


benlubas:
Question, should updating the ToC also update it's width? If I open a document with no headers, it will create an empty ToC that's not too wide, then I type a long heading, the ToC remains at it's earlier width, and is not updated.

I think it would be nice, but could get annoying depending on the other settings you have to see the window keep resizing

vhyrro:
Perhaps it would be nice to have a minimum TOC width as well? I don't think it's too jarring to have the TOC resize personally.

PR #1504, opened 2024-07-07
feat(completion): allow custom completion engines

benlubas:
lmk if you want me to tone down the self promo here. I'm not sure if I should include this example, or just document it in a generic way.

I'm biased, but I think it's good to have an officially recognized way to get completions without a completion plugin.


vhyrro:
Looks good to me :)

PR #1505, opened 2024-07-07
feat(intergrations.otter): update to use otter 2.0

benlubas:
[Otter.nvim](https://github.com/jmbuhr/otter.nvim) got a pretty large update.

TL;DR: it's a language server now, so no need to custom keybinds (normal LSP keybinds are used)

### DIFF
--- a/lua/neorg/modules/core/integrations/otter/module.lua
+++ b/lua/neorg/modules/core/integrations/otter/module.lua
@@ -111,24 +111,13 @@ module.config.public = {
     -- Automatically start Otter when a norg buffer is opened
     auto_start = true,
 
-    -- buffer local mappings set when otter is active
-    keys = {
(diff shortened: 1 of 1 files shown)


benlubas:
Back to draft as there is a bug causing inconsistent completions.

vhyrro:
Happy to see otter being improved. If you could explain the bug I can also give it a look.

benlubas:
https://github.com/jmbuhr/otter.nvim/issues/159

Perhaps you will have better luck finding a consistent reproduction. Though as jmbuhr mentions in that thread, it might be an issue with nvim or pyright.

benlubas:
@vhyrro I'm okay with merging this. I can't reproduce it anymore

vhyrro:
Hah sounds good to me!

PR #1509, opened 2024-07-08
feat!: massive keybinds refactor

vhyrro:
# Modernizing Keybinds

This pull request is a long-awaited refactor of the internal keybind system used by Neorg. In this refactor I seek to:
- Make changing keybinds as convenient as possible
- Introduce keybind presets (~~including a "vim" preset which makes full use of the vim keys for all operations~~ this will require a different pull request to consider all of the vim-available keys)
- Strip out large portions of the core.keybinds code, making startup faster
- Make proper use of `<Plug>` mappings

`<Plug>` may seem like a weird choice, but it is the defacto way of mapping keys. This way, various modules can map their behaviours to a `<Plug>` object without worrying about who binds it and where. Plug mappings also allow us to check for keybind conflicts reliably. When you map your own keybind to a Neorg mapping, Neorg will no longer forcefully create its own keybind, completely eliminating conflicts.

All in all, this system will be much more ergonomic and will only require a vim.keymap.set from the user's end, everything else will just work!

### TODOs

- [x] Refactor all modules to use the new system
- [x] Check all modules to ensure they still function as intended
- [x] Ensure that no other modules require core.keybinds
- [x] Reimplement mode support (like traverse-headings)
- [x] Fix user-overwritten keybinds not rebinding
- [x] ~~Reimplement keybind hook (required for mode interoperability)~~
- [x] Allow external modules to extend the default keybind list, thereby respecting the `default_keybinds` option
- [x] Implement keybind presets
- [x] Verify that dot repeat works
- [x] Update all documentation to no longer reference core.keybinds
- [x] Make docgen work with keybind system for discoverability


lua/neorg/modules/core/itero/module.lua
```diff
@@ -108,22 +109,7 @@ module.on_event = function(event)
...
-
-                vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(fallback, true, true, true), "t", false)
-                return
-            end
-
-            utils.notify(
-                "No object to continue! Make sure you're under an iterable item like a list or heading.",
-                vim.log.levels.WARN
-            )
+            vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<CR>", true, true, true), "t", false)
```
max397574, line 112: from my experience I know that `nvim_feedkeys` respects mappings. So if `<cr>` is mapped to this function bad things will happen.
The only way I could solve this was by creating fallback plug mappings for the keys used which I could create with noremap=true

max397574:
I think modes would be quite similar to what this plugin does https://github.com/shmerl/session-keys

Of course the optimal solution would be namespaced mappings but we don't know when will get those (if ever 😢 )

lua/neorg/modules/core/itero/module.lua
max397574: you may also be able to use [:h i_ctrl-v](https://neovim.io/doc/user/insert.html#i_CTRL-V)

lua/neorg/modules/core/itero/module.lua
```diff
@@ -108,22 +109,7 @@ module.on_event = function(event)
...
-
-                vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(fallback, true, true, true), "t", false)
-                return
-            end
-
-            utils.notify(
-                "No object to continue! Make sure you're under an iterable item like a list or heading.",
-                vim.log.levels.WARN
-            )
+            vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<CR>", true, true, true), "n", false)
```
max397574, line 112: you could also search for mappings which have the plug mapping as rhs and then use that as fallback
so it would also work if users mapped a different key to this

vhyrro: Thats a good idea

max397574, line 112:
```suggestion
            vim.api.nvim_feedkeys(vim.keycode("<CR>"), "n", false)
```

vhyrro: thanks for this, I forgot that was a thing.

vhyrro:
Mode support is much more sophisticated now and should fix a lot of issues found during day to day usage. This is because modes can now have parents which they inherit behaviours like keybinds from.

vhyrro:
Note: general keybinds that operate outside of norg files will not trigger because neorg lazy loads itself. This should be mitigated by moving the core startup code to `plugin/` and optimizing Neorg's startup time.

max397574:
just a thought I had
somehow hook into this for more default keybindings?
Thinking about neorg-telescope right now. Because currently it can't define keymaps for itself afaict
And we shouldn't just add them when telescope isn't available

max397574:
another thing is custom keybinds
so the user can just assign custom functions which aren't neorg keybinds to things (even if they don't define plug mappings)

you can just use normal keybinds because of the different modes

vhyrro:
I earlier alluded to reimplementing the keybind hook for interoperability, but instead settled on (in my opinion) a much better system. Keybind hooks are very limiting and unnecessary DSLs for our keybind system.

Users will have to set their keymaps in a buffer-local fashion either:
- using `.config/nvim/ftplugin/norg.lua`, where they can freely use `vim.keymap.set` to override any key they'd like
- using an autocommand defined somewhere in their configuration, inside of which they can also run `vim.keymap.set`

This will, by default, override that key for all Neorg modes. To remap a key in a specific mode only, one can simply query the current mode using `require("neorg.modules").get_module("core.mode").get_mode()` and conditionally run their mappings depending on the return value.

This approach is much more inline with the vim philosophy and much easier to adapt to any configuration style.

vhyrro:
Actually, I'm highly considering nuking mode support. Shouldn't we integrate with e.g. hydra instead?

max397574:
I'm against it tbh

vhyrro:
I never wanted to add it as a direct dependency. But integration with existing tools is much wiser than trying to reinvent the wheel in my eyes. It simplifies the existing code so much that I'd say it's worth pursuing.

vhyrro:
Just documentation left now!

vhyrro:
I don't believe I'm missing anything here. Should be ready to merge into main soon as a first test wave. If any of you are not pinned to the latest stable release then I encourage you to do so asap :p

lua/neorg/modules/core/itero/module.lua
vhyrro: Since this isn't that easy to do I'll probably leave this for a different time :|

PR #1513, opened 2024-07-15
Disallow keybind clashes

vhyrro:
This PR introduces three key changes:
- Don't forcefully override the user's keys
- Add proper `:checkhealth` integration which looks for various conflicts and remaps
- Set global keybinds only once, and apply all norg-specific keys on `Filetype` instead of `BufEnter`

All in all, this concludes the major changes I wanted to make to the keybind system. I personally consider the system to be very close to perfect now, just like the Neovim overlords intended.

### DIFF
--- a/lua/neorg/health.lua
+++ b/lua/neorg/health.lua
@@ -90,18 +90,35 @@ return {
             if key_healthcheck.preset_exists then
                 vim.health.info(string.format("Neorg is configured to use keybind preset `%s`", keybinds_config.preset))
             else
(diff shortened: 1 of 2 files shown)


vhyrro:
(cc @benlubas, in case there's anything else you wanted to add)

benlubas:
@vhyrro  can we update the documentation that tells users to set keybinds in a "BufEnter" autocommand to instead recommend "FileType"?

vhyrro:
> @vhyrro can we update the documentation that tells users to set keybinds in a "BufEnter" autocommand to instead recommend "FileType"?

done!

PR #1522, opened 2024-07-17
fix(keybinds): correctly get keymap opts

ALVAROPING1:
Previous implementation was trying to get the keymap opts from the list of keybinds for each mode rather than from the table of each mapping. I guess that was a typo because none of the mode lists have an opts field, but all of the mapping tables do


vhyrro:
Oh, great catch. Thank you!

PR #1527, opened 2024-07-19
feat(dirman): add option `name_notes`

youngnix:
Add the `name_notes` option to control whether to show a popup to name new notes

### DIFF
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -133,6 +133,9 @@ module.config.public = {
     -- Whether to use core.ui.text_popup for `dirman.new.note` event.
     -- if `false`, will use vim's default `vim.ui.input` instead.
     use_popup = true,
+    -- Whether to show a popup to name created notes
+    -- if `false`, a random UUID will be provided
+    name_notes = true,
 }
 
 module.private = {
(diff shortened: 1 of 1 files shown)


max397574:
try to fix the formatting, makes it much easier to review

also
perhaps we should allow a function to be used for naming the notes and the default value being an input.
This would allow users to create what you did here but with more extensibility

youngnix:
> try to fix the formatting, makes it much easier to review
> 
> also perhaps we should allow a function to be used for naming the notes and the default value being an input. This would allow users to create what you did here but with more extensibility

I'm using the default formatting settings provided by Neorg, as far as I'm aware. Not sure how to fix it in that case.

As for using a function for this setting, I believe that solution could in fact work better. I'm not exactly sure how to implement it, but I can try.

benlubas:
You can accomplish the goal of "creating a new, time-stamp-named note" without changing core at all. Just use the `create_file` function
```lua
local unnamed_note = function()
	local title = "whatever timestamp generation you want"
	require("neorg.modules.core.dirman.module").public.create_file(title)
end
```

youngnix:
> You can accomplish the goal of "creating a new, time-stamp-named note" without changing core at all. Just use the `create_file` function
> 
> ```lua
> local unnamed_note = function()
> 	local title = "whatever timestamp generation you want"
> 	require("neorg.modules.core.dirman.module").public.create_file(title)
> end
> ```

That is definitely a solution, though I would have never thought about messing with Neorg's internals for runtime configuration.

benlubas:
If you would prefer to use the neorg module loader to get access to that function you can do that as well:
```lua
local nm = require("neorg").modules
local dirman = nm.get_module("core.dirman")
if not dirman then return end
dirman.create_file("same thing")
```

youngnix:
I'll be leaving this idea to someone better suited to handle it. I don't think I'll be able to provide a full implementation, so if anyone would like to take the wheel feel free to do so.

PR #1528, opened 2024-07-20
feat: proper quote concealing

vhyrro:
Please help test this PR before it's merged! This PR adds proper quote concealing to Neorg, by displaying the box chars for every separate line. Example:

```
> This is a quote
   and this still belongs to said quote!
```

Previously, the second line would not be highlighted, even though it absolutely should.

This change fundamentally changes some concealer functionality. If you find any concealer bugs that weren't present earlier, please let me know!

@benlubas would you have some time to try and break this thing?

### DIFF
--- a/doc/neorg.norg
+++ b/doc/neorg.norg
@@ -50,11 +50,6 @@ version: 0.1
 
 *** Unordered lists
 
-    |example
-    - Unordered list level 1
-    -- Unordered list level 2
-    |end
-
     - Unordered list level 1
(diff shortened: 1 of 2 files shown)


benlubas:
Haven't noticed anything crazy happening with a little bit of normal use

vhyrro:
Nice, I can reproduce that one. Time to fix it.

vhyrro:
Tried again today, still couldn't break it. Merging :)

PR #1529, opened 2024-07-20
fix: set monthly view of calendar reasonably buffer options

belltoy:
Set the `filetype` of monthly view of calendar buffer more reasonably. And turn cursor column and line on to help more continent locate the cursor in the dates.


max397574:
The prompt filetype is perhaps not 100% correct but makes perhaps sense

but I'm strongly against setting cursorline and cursor column
it makes the ui just look horrible

belltoy:
> The prompt filetype is perhaps not 100% correct but makes perhaps sense

Agreed! The `prompt` ft maybe not correct. Maybe the monthly view worth its own ft. In addition to this, a dedicated hl group should give users and colorschemes enough to customize its look, the cursorline and column is no necessary.

PR #1537, opened 2024-07-24
Automatically typed module.required dependencies

mtrajano:
Did a quick POC for automatic typing of module resolves. Wondering if something like this would be useful to us. Benefit is that doing `local mod = module.required[{dependency}]` mod will be automatically typed with the correct class and will give us luals type hints and typechecks without having to do something like this every time:

```lua
---@type core.integrations.treesitter
local ts = module.required["core.integrations.treesitter"]
```
Type hinting:
Screenshot 2024-07-24 at 12 42 50 PM

It already showed some incorrectly type methods in the codebase
Screenshot 2024-07-24 at 12 36 07 PM

We also get automatic suggestions on which modules can be imported as you can see below:
Screenshot 2024-07-24 at 12 34 48 PM

Downside is obviously that any new added modules will have to be updated in that new `neorg.module.resolver` class

Thoughts/feedback? @vhyrro @benlubas

### DIFF
--- a/lua/neorg/core/modules.lua
+++ b/lua/neorg/core/modules.lua
@@ -13,6 +13,47 @@ local utils = require("neorg.core.utils")
 
 --- @alias neorg.module.public { version: string, [any]: any }
 
+--- @class (exact) neorg.module.resolver
+--- @field ["core.autocommands"] core.autocommands
+--- @field ["core.clipboard"] core.clipboard
+--- @field ["core.completion"] core.completion
+--- @field ["core.concealer"] core.concealer
+--- @field ["core.dirman"] core.dirman
+--- @field ["core.esupports.hop"] core.esupports.hop
+--- @field ["core.esupports.indent"] core.esupports.indent
+--- @field ["core.esupports.metagen"] core.esupports.metagen
+--- @field ["core.export"] core.export
+--- @field ["core.export.markdown"] core.export.markdown
+--- @field ["core.fs"] core.fs
+--- @field ["core.highlights"] core.highlights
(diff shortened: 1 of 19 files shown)


max397574:
really like this change

benlubas:
I also really like this

vhyrro:
LGTM, thank you! I played around with this and tried my best to make some clever use of generics but couldn't figure it out. Hardcoded it is (for now, until LuaCATS generics get better :p).

vhyrro:
Are there any other modules that need to be annotated, or can this be marked as ready?

vhyrro:
Oops, impulsively pressed the button to let the CI run, made the PR "ready" instead. My bad.

mtrajano:
It pretty much is ready and fully functional. I think I managed to type everything but there were some where the class typing was on the module.private so I can update those. Also wasn't 100% unsure on the syntax, not sure if a class or an indexed table etc.. 

Other than that just fixing the incorrect typings that will pop up but that feels like a separate PR or we can just fix them as we come across them.

mtrajano:
> LGTM, thank you! I played around with this and tried my best to make some clever use of generics but couldn't figure it out. Hardcoded it is (for now, until LuaCATS generics get better :p).

That's interesting, didn't consider the generics route but agree not sure if we're there yet with LuaCATS 😄

mtrajano:
Ok I believe I got all of them but maybe worth a second pair of eyes, let me know if I missed anything

benlubas (approved):

vhyrro:
Looking good! Thanks again. Guess I'll now have to spend some time looking at the type errors :)

PR #1541, opened 2024-07-28
fix: Neorg return to most recent file

benlubas:
`:Neorg return` was bringing you to the first file you opened, this is not what the docs say, and also not intuitive.


NTBBloodbath (approved):
Thanks for the patch buddy :)

boltlessengineer:
You can use `:enumerate()` instead of `ipair(…:totable())`
```lua
for _, buffer in vim.iter(buffers):rev():enumerate() do
```

benlubas:
Even better, you can just iterate through `vim.iter(list):rev()` apparently

NTBBloodbath (approved):

PR #1549, opened 2024-08-05
fix(ToC): only try to open when still in norg buffer

benlubas:
When switching buffers really quickly with auto TOC enabled, we could enter and leave a norg buffer before the scheduled function gets to run, which results in a noisy error message from Neorg Command saying `:Neorg toc` only works from a norg buffer.

### DIFF
--- a/lua/neorg/modules/core/qol/toc/module.lua
+++ b/lua/neorg/modules/core/qol/toc/module.lua
@@ -48,8 +48,10 @@ module.load = function()
             pattern = "*.norg",
             callback = function()
                 vim.schedule(function()
-                    next_open_is_auto = true
-                    vim.cmd([[Neorg toc]])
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/qol/toc/module.lua
```diff
@@ -48,8 +48,12 @@ module.load = function()
             pattern = "*.norg",
             callback = function()
                 vim.schedule(function()
-                    next_open_is_auto = true
-                    vim.cmd([[Neorg toc]])
+                    if
+                        vim.api.nvim_get_option_value("filetype", { buf = vim.api.nvim_get_current_buf() }) == "norg"
```
vhyrro, line 52: nitpick: doesn't change anything, but maybe just `vim.bo.filetype` for brevity? :)

PR #1567, opened 2024-09-01
fix comment

krmpotic:
fixes documentation of https://github.com/nvim-neorg/neorg/wiki/Summary
by removing a leftover comment

### DIFF
--- a/lua/neorg/modules/core/summary/module.lua
+++ b/lua/neorg/modules/core/summary/module.lua
@@ -300,7 +300,9 @@ module.config.public = {
     -- - "default" - read the metadata to categorize and annotate files. Files
     --   without metadata will use the top level heading as the title. If no headings are present, the filename will be used.
     -- - "by_path" - Similar to "default" but uses the capitalized name of the folder containing a *.norg file as category.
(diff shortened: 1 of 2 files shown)


lua/neorg/modules/core/summary/module.lua
```diff
@@ -300,7 +300,6 @@ module.config.public = {
     -- - "default" - read the metadata to categorize and annotate files. Files
     --   without metadata will use the top level heading as the title. If no headings are present, the filename will be used.
     -- - "by_path" - Similar to "default" but uses the capitalized name of the folder containing a *.norg file as category.
```
benlubas, line 302: We should make it clear that the user can provide a function, something like:
```suggestion
    -- - "by_path" - Similar to "default" but uses the capitalized name of the folder containing a *.norg file as category.
    -- A custom function with the signature: `fun(files: PathlibPath[], ws_root: PathlibPath,
    -- heading_level: number?, include_categories: string[]?): string[]?`. Returning a list of lines
    -- that make up the summary
```

benlubas (changes requested):

krmpotic:
changed as suggested

benlubas (approved):

PR #1568, opened 2024-09-02
todo_item parent update behaviors

benlubas:
I'm not sure when this changed, or exactly what broke it, but parent items used to update when a keybind was used to change a child state, and sometime around the keybinds refactor, this broke (bisecting to an exact commit is tricky b/c there are multiple commits where the plugin fails to load around this time).

I took this opportunity to make this behavior configurable, b/c personally I've always hated it :P

### DIFF
--- a/lua/neorg/modules/core/qol/todo_items/module.lua
+++ b/lua/neorg/modules/core/qol/todo_items/module.lua
@@ -92,6 +92,20 @@ module.config.public = {
     -- ```
     create_todo_parents = false,
 
+    -- Automatically update the parent todo state when a child node is updated.
+    --
+    -- eg:
+    -- ```norg
+    -- - ( ) parent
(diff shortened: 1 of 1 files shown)


nate-credoai:
Looks like it changed in PR #1509 here: https://github.com/nvim-neorg/neorg/pull/1509/files#diff-000516599b7c7ce272aebe1a0b5b11118ae48d92eea59b6c76cc02cfa407a810L432-L438 .

nate-credoai:
> Looks like it changed in PR #1509 here: https://github.com/nvim-neorg/neorg/pull/1509/files#diff-000516599b7c7ce272aebe1a0b5b11118ae48d92eea59b6c76cc02cfa407a810L432-L438 .

Actually, this seems to be working fine for me now. Not sure what I changed.

PR #1574, opened 2024-09-08
fix: don't suggest leading chars for file completions

benlubas:
This caused problems with completion engines that didn't do extra work. As a result, completion worked fine with cmp, but caused extra chars to be completed for other engines like builtin, and now care.nvim. (probably mini-completion too but I've never tested).

Now it works with all of them!

### DIFF
--- a/lua/neorg/modules/core/completion/module.lua
+++ b/lua/neorg/modules/core/completion/module.lua
@@ -139,7 +139,7 @@ module.private = {
             if not file:samefile(Path.new(vim.api.nvim_buf_get_name(0))) then
                 local rel = file:relative_to(files[1], false)
                 if rel and rel:len() > 0 then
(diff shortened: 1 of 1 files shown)


lua/neorg/modules/core/completion/module.lua
```diff
@@ -160,14 +160,11 @@ module.private = {
...
-        -- return vim.iter(links)
-        --     :map(function(x)
-        --         return leading_whitespace .. x.title .. closing_chars
-        --     end)
-        --     :totable()
+        return vim.iter(links)
+            :map(function(x)
+                return leading_whitespace .. x.title .. closing_chars
+            end)
+            :totable()
```
benlubas, line 167: and since we're supporting 0.10+ now, I've dropped the soon to be deprecated tbl_map.

vhyrro (approved):

PR #1577, opened 2024-09-10
docs: getting started updated

apraga:
Be more clear how TODO status are shown with concealer. I found it confusing the first time that TODO are shown as 'x' but 'x' is translated as DONE.


apraga:
That would be less confusing :)

PR #1578, opened 2024-09-11
fix: intuitive default undone icon

benlubas:
This has come up enough times that I feel justified in making this PR even though you intentionally changed it to the x a while ago.


benlubas:
most recently in #1577

PR #1580, opened 2024-09-13
feat(esupports.hop): Add tab drop as option for open_mode

devansh08:
This adds `:tab drop` as an option for opening a file when clicking a file link. Tab drop as an option would be helpful to avoid opening a file a new tab, when it has already been open somewhere else.

IMO, this could be a replacement for the `tabnew` option itself, as tab drop will open a new tab normally if the same file is already not open. But it can be kept as a separate option as well. Also IMO, never bad to have multiple options :)

Do let me know your thoughts on this. If this looks good and needs any documenting, happy to work on that as well.

Thanks!

### DIFF
--- a/lua/neorg/modules/core/esupports/hop/module.lua
+++ b/lua/neorg/modules/core/esupports/hop/module.lua
@@ -42,6 +42,7 @@ module.load = function()
     dirman_utils = module.required["core.dirman.utils"]
     vim.keymap.set("", "<Plug>(neorg.esupports.hop.hop-link)", module.public.hop_link)
     vim.keymap.set("", "<Plug>(neorg.esupports.hop.hop-link.vsplit)", lib.wrap(module.public.hop_link, "vsplit"))
+    vim.keymap.set("", "<Plug>(neorg.esupports.hop.hop-link.tab-drop)", lib.wrap(module.public.hop_link, "tab-drop"))
 end
 
 module.config.public = {
@@ -216,9 +217,16 @@ module.public = {
                 end,
 
                 buffer = function()
-                    open_split()
+                    if open_mode ~= "tab-drop" then
+                        open_split()
+                    end
 
                     if located_link_information.buffer ~= vim.api.nvim_get_current_buf() then
+                        if open_mode == "tab-drop" then
+                            vim.cmd("tab drop " .. vim.api.nvim_buf_get_name(located_link_information.buffer))
+                            return
+                        end
+
                         vim.api.nvim_buf_set_option(located_link_information.buffer, "buflisted", true)
                         vim.api.nvim_set_current_buf(located_link_information.buffer)
                     end
--- a/lua/neorg/modules/core/esupports/hop/module.lua
+++ b/lua/neorg/modules/core/esupports/hop/module.lua
@@ -17,6 +17,8 @@ mapping them):
 
 - `neorg.esupports.hop.hop-link` - Follow the link under the cursor, seeks forward
 - `neorg.esupports.hop.hop-link.vsplit` - Same, but open the link in a vertical split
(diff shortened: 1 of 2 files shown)


lua/neorg/modules/core/esupports/hop/module.lua
```diff
@@ -208,9 +209,16 @@ module.public = {
                 end,
 
                 buffer = function()
-                    open_split()
+                    if open_mode ~= "tab-drop" then
+                        open_split()
+                    end
 
                     if located_link_information.buffer ~= vim.api.nvim_get_current_buf() then
+                        if open_mode ~= "tab-drop" then
```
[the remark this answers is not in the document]
devansh08: Fixed now

devansh08:
@benlubas Fixed the mentioned issue and added docs.
I accidentally broke my git while trying to rebase, so had to commit my changes again. My bad!

Let me know if the docs part is good. Could not think of any better keybind using CR, so went with C-t. Let me know if you have any better ideas for that.

Thanks!

devansh08:
I've just added another similar option that uses `:drop` instead. This works similar to `tab-drop` but would be useful for a buffer focused workflow (to which I have recently switched over :)). So cases where you have the destination already open in a split, it would just navigate to that instead of opening the destination in the current buffer (what default `hop` does today).

Let me know your thoughts. Thanks!

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -298,6 +298,21 @@ module.private = {
...
+                    -- If destination is already open in an existing tab, just navigate to it
+                    {
+                        "<C-t>",
+                        "<Plug>(neorg.esupports.hop.hop-link.tab-drop)",
+                        opts = { desc = "[neorg] Jump to Link (Tab Drop)" },
+                    },
+
+                    -- Same as `<CR>`, except if destination is already open in a buffer, just navigate to it
+                    {
+                        "<C-d>",
```
benlubas, line 311: `<c-d>` is the keybind to scroll down half a page. It's incredibly widely used, so this is not an acceptable default.

I'd say this doesn't have to have a default binding.

benlubas (changes requested):
Docs look good, but we can't override `<C-d>` that's a default vim bind.

lua/neorg/modules/core/keybinds/module.lua
devansh08: I use `<C-d>` and I completely forgot about it when writing this :smile:
Sure, will remove it from the defaults.

devansh08: Done

lua/neorg/modules/core/keybinds/module.lua
```diff
@@ -298,6 +298,14 @@ module.private = {
                         opts = { desc = "[neorg] Jump to Link (Vertical Split)" },
                     },
 
+                    -- Same as `<CR>`, except open the destination in a new tab
+                    -- If destination is already open in an existing tab, just navigate to it
+                    {
```
benlubas, line 303: Okay wait I'm so sorry, but `<c-t>` is a builtin tag list keybind, it takes you back to the previous spot in the tag list. I totally forgot about that. @devansh08

devansh08: Ah no worries. I did not realize it either.
Would something like `<M-t>` be fine ? Don't see that keybind do anything in base neovim normal mode. I'm kinda running out of logical keybinds to use :sweat_smile:

benlubas: yeah that would be fine, I don't think that neovim binds anything to alt by default b/c default terminals don't interpret it.

devansh08: Done :+1:

benlubas (approved):

PR #1587, opened 2024-09-29
Fixed tree-sitter-norg parser builds on MacOS.

pcekm:
- Added use_makefile, which is how nvim-treesitter builds the norg parser by default and which correctly adds the required compiler flags
 - Removed a macOS build kludge that is no longer needed (and seems not to work anymore)
 - Removed branch specification which is ignored when setting revision.

Tested on an Apple silicon machine running macOS 14.7, and an Intel machine running macOS running macOS 13.7.

Fixes nvim-neorg/tree-sitter-norg#45

### DIFF
--- a/lua/neorg/modules/core/integrations/treesitter/module.lua
+++ b/lua/neorg/modules/core/integrations/treesitter/module.lua
@@ -137,16 +137,16 @@ module.config.public = {
         norg = {
             url = "https://github.com/nvim-neorg/tree-sitter-norg",
             files = { "src/parser.c", "src/scanner.cc" },
-            branch = "main",
             revision = "6348056b999f06c2c7f43bb0a5aa7cfde5302712",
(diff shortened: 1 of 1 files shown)


satanx016:
Hey I think you're changes are already covered in this [PR](https://github.com/nvim-neorg/neorg/pull/1020/commits/03c8db64b48ba1567fb028e92c542dbf6cb24fdb)

benlubas:
#1020 is stalled though, I could see merging this in the meantime.

wjyoung65:
FWIW, I got a clean build with this PR on an M3 MacBook running MacOS 15, and also on an x86_64 laptop running Linux Mint.

saxon1964:
This is a must have! Compiled without issues on Neorg 9.2.0 and Mac 14.7.4.

PR #1597, opened 2024-10-20
fix: add coq_nvim completion integration

### LINKED ISSUE #1594: Errors when using coq_nvim for autocompletion
### Prerequisites

### Neovim Version

v0.10.2

### Neorg setup
```
require("neorg").setup({
  load = {
	["core.defaults"] = {},
	["core.concealer"] = {},
	["core.autocommands"] = {},
	["core.integrations.image"] = {},
	["core.integrations.coq_nvim"] = {},
	["core.pivot"] = {},
	["core.latex.renderer"] = {
		config = {
			conceal = true,
			render_on_enter = true,
			dpi = 450,
			scale = 1,
		},
	},
	["core.completion"] = {
		config = {
		 engine = "coq_nvim",
		},
	},
	["core.integrations.treesitter"] = {
		config = {
			install_parsers = false; -- handled by nix
		},
	},
	["core.dirman"] = {
		config = {
			workspaces = {
				uni = "~/Documents/Uni",
			},
		},
	},
  }
})
```

### Actual behavior

whenever I try linking to something in neorg, coq complains with an error like 
`coq_nvim/lua/coq/lsp-request.lua:373: NotImplementedError :: coq_3p.args->full_line` as soon as i type {:<any symbol>

### Expected behavior

coq should provide fitting auto completion suggestions for the links.

### Steps to reproduce

1. Install Neorg and setup coq_nvim
2. tell neorg to use coq_nvim for autocompletion
3. open a neorg file
4. try to link to something in neorg

_No response_

### Other information

I tried also with Neorg v8.6.0 and coq_nvim from 2024-04-26.
I use NixOS unstable.

### Help

No

### Implementation help

_No response_

TheLeoP:
Fixes #1594

### DIFF
--- a/lua/neorg/modules/core/integrations/coq_nvim/module.lua
+++ b/lua/neorg/modules/core/integrations/coq_nvim/module.lua
@@ -67,17 +67,39 @@ module.public = {
                     return callback()
                 end
 
-                local completion_cache = module.public.invoke_completion_engine(args)
+                local row, col = unpack(args.pos)
+                local line = args.line
+                local before_char = line:sub(#line, #line)
+                -- Neorg requires a nvim-compe like context nvim-compe defines
+                -- the following fields. Not all of them are used by Neorg, but
+                -- they are left here for future reference
+                local context = {
+                    start_offset = nil,
+                    char = col + 1,
+                    before_char = before_char,
+                    line = line,
+                    column = nil,
+                    buffer = nil,
+                    line_number = row + 1,
+                    previous_context = nil,
+                    full_line = line,
+                }
+                local completion_cache = module.public.invoke_completion_engine(context)
+                local prev = line:match("({.*)$")
 
                 if completion_cache.options.pre then
-                    completion_cache.options.pre(args)
+                    completion_cache.options.pre(context)
                 end
 
                 local completions = vim.deepcopy(completion_cache.items)
 
                 for index, element in ipairs(completions) do
-                    local word = element
-                    local label = element
+                    -- coq_nvim requries at least 2 exact prefix characters by
+                    -- default. Users could chagnge this setting, but instead
+                    -- we are providing the start of the match (for links) so
+                    -- coq_nvim doesnt' filter our results
+                    local word = (prev or "") .. element
+                    local label = (prev or "") .. element
                     if type(element) == "table" then
                         word = element[1]
                         label = element.label


lua/neorg/modules/core/integrations/coq_nvim/module.lua
```diff
@@ -67,17 +67,39 @@ module.public = {
...
 
-                local completion_cache = module.public.invoke_completion_engine(args)
+                local row, col = unpack(args.pos)
+                local line = args.line
+                local before_char = line:sub(#line, #line)
+                -- Neorg requires a nvim-compe like context nvim-compe defines
+                -- the following fields. Not all of them are used by Neorg, but
+                -- they are left here for future reference
+                local context = {
+                    start_offset = nil,
```
benlubas, line 77: What is the point of setting a key to nil here? In lua this is the same as not including the key at all

benlubas (changes requested):
I have no way to easily test this. I trust that the heart reaction means that it worked.

Just a few lines to remove and then I will merge

LoCrealloc:
the heart reaction meant something like "great that someone is putting effort into this", i did not test this pull request (yet)

lua/neorg/modules/core/integrations/coq_nvim/module.lua
TheLeoP: I left them there for future reference.

As far as i can tell, there is no formal definition/interface for what the completion context should have. When i originally added support for coq, I missed a lot of fields (and the type inference of lua-ls didn't help (because of how the modules are structured, I think. I didn't look into it).

For this PR i had to look into both nvim-cmp and nvim-compe source code to know what fields a context is supposed to have and what the value of each one is supposed to be. Not all of the context's fields were currently being used by neorg, but since the current implementation seems tightly coupled with what nvim-compe consideres a context I thought leaving them here will help with future changes (removing the need to check nvim-compe and nvim-cmp codebases)

benlubas:
Oh, please test when you get the chance and let me know.

LoCrealloc:
@benlubas just tested the changes and as far as I can tell the problem is solved - no  error messages anymore

lua/neorg/modules/core/integrations/coq_nvim/module.lua
benlubas: fair enough

benlubas (approved):

PR #1613, opened 2024-12-07
Add MINGW32_NT to get_os_info

GabrielFrigo4:
If you install neovim using [MSYS 2](https://www.msys2.org/), the operating system will be `mingw32_nt-<ver>`.
in my case it was `mingw32_nt-10.0`.

Therefore, it is necessary to add a check for the prefix `mingw32_nt` to see if the operating system is Windows

### DIFF
--- a/lua/neorg/core/config.lua
+++ b/lua/neorg/core/config.lua
@@ -36,7 +36,7 @@
 local function get_os_info()
     local os = vim.loop.os_uname().sysname:lower()
 
-    if os:find("windows_nt") then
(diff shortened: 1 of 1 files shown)


benlubas:
I'm curious if the `32` can also be `64`, and if so, if it would be better to just match against `mingw` instead?

GabrielFrigo4:
> I'm curious if the `32` can also be `64`, and if so, if it would be better to just match against `mingw` instead?

The `system` in `mingw32`, `mingw64`, `clang32`, `clang64`, `ucrt64` are `mingw32_nt-<ver>`

PR #1618, opened 2024-12-15
fix: Tangle should close fd for files

Jimmyscene:
The current implementation keeps the fd open for each file it is tangling, which results in tangled shell scripts not being able to be run until the instance is closed

### DIFF
--- a/lua/neorg/modules/core/tangle/module.lua
+++ b/lua/neorg/modules/core/tangle/module.lua
@@ -526,6 +526,7 @@ module.on_event = function(event)
                         )
                     end
                 end)
+                vim.uv.fs_close(fd)
             end)
         end
     end


Jimmyscene:
Small fix, but let me know if there's anything I should update :)

lua/neorg/modules/core/tangle/module.lua
```diff
@@ -526,6 +526,7 @@ module.on_event = function(event)
                         )
                     end
                 end)
+                vim.loop.fs_close(fd)
```
benlubas, line 529: `vim.loop` -> `vim.uv`

see `:h vim.loop`

benlubas (changes requested):

benlubas (approved):

PR #1623, opened 2024-12-23
feat(dirman): dynamically set default workspace

mtrajano:
Ability to dynamically set the default workspace based on the `cwd`.

This gives you the ability to open different workspaces based on context for example:
```lua
config = {
  workspaces = {
    dev = '~/notes/dev',
    personal = '~/notes/personal',
  },
  default_workspace = function()
    local git_root = vim.fs.root(0, '.git')
    local basename = vim.fs.basename(git_root)

    if basename == 'dotfiles' then
      return 'dev'
    end

    return 'personal'
  end,
},

```

### DIFF
--- a/lua/neorg/modules/core/dirman/module.lua
+++ b/lua/neorg/modules/core/dirman/module.lua
@@ -92,7 +92,7 @@ module.load = function()
 
     if module.config.public.open_last_workspace and vim.fn.argc(-1) == 0 then
         if module.config.public.open_last_workspace == "default" then
-            if not module.config.public.default_workspace then
+            if not module.public.get_default_workspace() then
                 log.warn(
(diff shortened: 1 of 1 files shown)


benlubas:
I like this addition. Can you document the type of cwd. Bc users would probably assume it's a string.

Honestly it's not necessary to pass the cwd at all. People can get it from neovim themselves. But idc either way

mtrajano:
@benlubas That's a great point, that way the user isn't tied to the more internal Path object and can resolve it however he/she wants. I updated the pr and the example comment of how this would look like using the nvim internal fs api. Let me know if you think this comment should be removed as well since the cwd isn't passed anymore `may be set dynamically based on cwd`

benlubas (approved):

PR #1634, opened 2025-01-11
docs: generate vim help pages in norg from wiki markdown

champignoom:
- Add `:help neorg-wiki`, generated with https://github.com/pysan3/minorg and several pandoc filters
- Rename `:help neorg` into `:help norg`
- Rename `:help neorg-cheatsheet` into `:help norg-cheatsheet`
- Remove `:help neorg-breaking-changes`, which didn't exist anyway
- Remove all other `:help neorg-*`, which are more noisy than helpful


doc/wiki/Autocommands.norg
```diff
@@ -0,0 +1,56 @@
...
+* Required By
+
+  - {https://github.com/nvim-neorg/neorg/wiki/Concealer}[`core.concealer`] - Enhances the basic Neorg experience by using icons instead of text. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Dirman}[`core.dirman`] - This module is be responsible for managing directories full of .norg files. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Indent}[`core.esupports.indent`] - A set of instructions for Neovim to indent Neorg documents. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Metagen}[`core.esupports.metagen`] - A Neorg module for generating document metadata automatically. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Core-Highlights}[`core.highlights`] - Manages your highlight groups with this module. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Core-Latex-Renderer}[`core.latex.renderer`] - An experimental module for rendering latex images inline. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Storage}[`core.storage`] - Deals with storing persistent data across Neorg sessions. 
+  - {https://github.com/nvim-neorg/neorg/wiki/Syntax}[`core.syntax`] - Handles interaction for syntax files for code blocks.
```
benlubas, line 55: Can we convert wiki links to relative path links to other help files?

doc/wiki/_Sidebar.norg
```diff
@@ -0,0 +1,52 @@
+
```
benlubas, line 1: we could change the name of this file to `module-list` maybe? or just remove it all together

doc/wiki/Setup-Guide.norg
```diff
@@ -0,0 +1,131 @@
...
+   Let’s adapt our configuration: 
+
+   @code lua
+   require("neorg").setup({
+   load = {
+   ["core.defaults"] = {},
+   ["core.concealer"] = {}, -- We added this line!
+   }
+   })
+   @end
```
benlubas, line 74: what is the bug that causes this? We don't have to re-indent, just copy the existing code which is correctly indented.

benlubas (commented):
I'm a fan of this PR. few questions below, and additionally: can you add a CI pipeline to generate these automatically on merge with main?

doc/wiki/Setup-Guide.norg
champignoom: The raw norg generated by minorg is badly formatted and not indented at all, which is why I used neorg's indenter, despite of its bug like this.

doc/wiki/_Sidebar.norg
champignoom: done

doc/wiki/Autocommands.norg
champignoom: done

doc/wiki/Setup-Guide.norg
benlubas: I'm pretty sure neorg's indenter relies on TS in code blocks, and it works for lua if I indent manually

champignoom:
> can you add a CI pipeline to generate these automatically on merge with main?

I don't really know how, but here's the generator https://github.com/champignoom/neorg-help-gen.

doc/wiki/Setup-Guide.norg
champignoom: you're right. I just tried `gg=G` manually and it works. Somehow `nvim -c "gg=G"` doesn't work.

benlubas: -c takes a command, so `nvim -c "norm gg==G"` should work, but I also see that it doesn't work for norg. It works for plain python though.

My guess is that something about TS injections doesn't get to load.

---

I wrote that and then went to go investigate further. It seems that I was right. and forcing a full parse computes the injections.

This works:
`nvim --headless -c "TSBufEnable indent" -c "lua vim.treesitter.get_parser(0, 'norg', {}):parse(true)"  -c "norm! gg=G" -c "wq" file`

benlubas: I'll note that this require nvim TS plugin, and neorg, so when we do this in CI we should probably use a neovim setup with that stuff via nix or something.

benlubas: > there's indeed a bug in neorg's indenter: gg=G doesn't indent code snippets outside currently visible region

I'm not seeing this behavior

champignoom: there's indeed a bug in neorg's indenter: `gg=G` doesn't indent code snippets outside currently visible region.

EDIT: it's likely treesitter's problem. See https://github.com/neovim/neovim/discussions/31970

champignoom: done

Issue #1624, opened 2024-12-30
Neorg generate-workspace-summary causes an error


erezak:
### Prerequisites

- [X] I am using the latest stable release of Neovim
- [X] I am using the latest version of the plugin

### Neovim Version

v0.10.2

### Neorg setup

```lua
  1   {                                                                                                                                                                                                 
  2     "nvim-neorg/neorg",                                                                                                                                                                             
  3     opts = {                                                                                                                                                                                        
  4       load = {                                                                                                                                                                                      
  5         ["core.defaults"] = {},                                                                                                                                                                     
  6         -- ["core.dirman"] = {                                                                                                                                                                      
  7         --  config = {                                                                                                                                                                              
  8         --    workspaces = {                                                                                                                                                                        
  9         --      notes = "~/notes/general",                                                                                                                                                          
 10         --      aidev = "~/notes/aidev",                                                                                                                                                            
 11         --      example_gtd = "~/notes/gtd",                                                                                                                                                        
 12         --    },                                                                                                                                                                                    
 13         --    default_workspace = "notes",                                                                                                                                                          
 14         --  },                                                                                                                                                                                      
 15         -- },                                                                                                                                                                                       
 16         -- -- ["core.integrations.nvim-cmp"] = {},                                                                                                                                                  
 17         -- ["core.completion"] = {                                                                                                                                                                  
 18         --  config = {                                                                                                                                                                              
 19         --    engine = "nvim-cmp",                                                                                                                                                                  
 20         --  },                                                                                                                                                                                      
 21         -- },                                                                                                                                                                                       
 22         -- ["core.concealer"] = {                                                                                                                                                                   
 23         --  config = {                                                                                                                                                                              
 24         --    folds = false,                                                                                                                                                                        
 25         --  },                                                                                                                                                                                      
 26         -- },                                                                                                                                                                                       
 27         -- ["core.integrations.image"] = {},                                                                                                                                                        
 28         -- ["core.latex.renderer"] = {},                                                                                                                                                            
 29         -- ["core.export"] = {},                                                                                                                                                                    
 30         ["core.summary"] = {},                                                                                                                                                                      
 31       },                                                                                                                                                                                            
 32     },                                                                                                                                                                                              
 33   },                                                                                                                                                                                                
```

### Actual behavior

Generated error when trying to generate summary:

![image](https://github.com/user-attachments/assets/727dd2a4-bc35-461a-9d4c-ec71654fff04)

```
[...]

PR #1637, opened 2025-01-15
fix(summary): Fix #1624

OmegaLambda1998:
Extract category / path args from generate_workspace_summary command. Tested with both strategies and it seems to work.

PR #1640, opened 2025-01-21
Fix https://github.com/nvim-neorg/neorg/issues/1624

erezak:
The include_categories parameter to generate_workspace_summary is assumed to contain only strings, but it doesn't.
I added sanitation code.

It works as expected with and without categories.


benlubas:
I chose to correct #1637 instead, as it's a lot simpler. Thank you for taking the time to file the issue and look into a fix though!

Issue #1652, opened 2025-02-17
Latex Issue on Mac?


TogiFerretFerret:
I was having trouble getting latex rendering to work on Mac - I found that one of the latex checks consistently fails due to an error code with the latex command, but the command produces 100% valid dvi files (I know since I forked the repo, removed that check and it works flawlessly.) Not sure what the issue was, I don't think its that I'm using the nightly (since it was an issue where the command did return a false flag but ran correctly) (I use MacTex on an Intel Macbook)
Not sure what the problem is, the change is minor so I'm pretty sure it can be merged.
[LINK](https://github.com/TogiFerretFerret/neorg)

benlubas:
Did you check std error of the latex command to see why it's exiting with a non zero exit code?

PR #1653, opened 2025-02-17
Latex Issue on Mac minor fix

TogiFerretFerret:
Issue I made: https://github.com/nvim-neorg/neorg/issues/1652
This pull request includes changes to the `lua/neorg/modules/core/latex/renderer/module.lua` file to fix the broken check.


benlubas:
have you inspected the command output to see why it's failing?

I feel like this is either a bug in the latex command, or something wrong with your setup. I've run this on macOS before.

I'd rather not merge this fix as it seems like it's just hiding a problem

benlubas:
closing for now

PR #1655, opened 2025-02-18
#1654: Use luarocks' GitHub actions

### LINKED ISSUE #1654: The "Automatically Generate Documentation" GitHub Action always fails
[The current GitHub Action](https://github.com/nvim-neorg/neorg/blob/main/.github/workflows/gendoc.yml) which generates the documentation appears to consistently fail in its current implementation. The last action that appears to have succeeded is for commit 09deca2ba5adcf1a729d1e3abf64438d7ce4d7ee.

From what I gather, the issue lies in the installing of lua and luarocks, which uses a GitHub Action which has been malfunctioning since the 17th of August, as per https://github.com/leafo/gh-actions-lua/issues/49

My recommendation would be to switch to a continuation of this action, as descibed in https://github.com/brimworks/lua-zlib/pull/69

MaxVanDeursen:
Fixes #1654

### DIFF
--- a/.github/workflows/gendoc.yml
+++ b/.github/workflows/gendoc.yml
@@ -36,10 +36,10 @@ jobs:
           neovim: true
           version: ${{ matrix.version }}
 
-      - uses: leafo/gh-actions-lua@v9
+      - uses: luarocks/gh-actions-lua@v10
         with:
           luaVersion: "luajit-2.1.0-beta3"
-      - uses: leafo/gh-actions-luarocks@v4
+      - uses: luarocks/gh-actions-luarocks@v5
 
       - name: Install all required modules
         run: |


MaxVanDeursen:
Please let me know if an alternative solution is preferred or whether some other factor should be considered prior to merging this PR!

NTBBloodbath (approved):
LGTM, thanks for addressing this!

Failed CIs can be ignored for merging, they are not related.

benlubas:
lovely, seems like this deleted half the wiki

MaxVanDeursen:
@benlubas Sorry to hear that the merge of this PR caused pages to be deleted. Is there something I/we can do to make sure this does not occur in the future?

benlubas:
Nothing you could have done. The doc gen cut is setup in weird way.

We can temporarily disable this ci job and manually restore the wiki. I'll probably do this later today

MaxVanDeursen:
@benlubas I don't mind focusing a bit on the nightly build and see how to solve this. I indeed have now confirmed that the last release of neovim does not have the error, while the nightly build does, locally.

I will submit a PR if I find the fix!

benlubas:
Oh interesting, I was unable to replicate the error locally. I guess the nightly flake was a little behind.

@vhyrro since you setup the original CI, is there an easy way to switch from nightly to stable?

MaxVanDeursen:
@benlubas You can set the version through a parameter. I will make a PR for this this evening (12 hours)

MaxVanDeursen:
@benlubas See #1662 for a PR which does this.

MaxVanDeursen:
@benlubas I have tracked down the issue: It appears the treesitter's parser is not loaded correctly in the `get_first_node` nor `get_first_node_recursive` function of the `core/integrations/treesitter` module. Calling `parser:parse()` prior to `for_each_tree(...)` seems to solve this problem. I do not know whether this is the most efficient solution so am currently looking for a relevant change in `neovim` that explains this.

benlubas:
Recently there was async TS parsing code merged. I think the highlight code now uses async, and the highlight code was previously what performed the full parse.

I think we should only call parse in neovim 0.11, and start to convert these queries to use the async TS API once it's better documented.

Btw I really appreciate that you dug into this. I will marge your fix after work today.

MaxVanDeursen:
@benlubas My pleasure. Regarding your comment about calling parse in 0.11, is that something we want to do instead of #1662, is that something we want to do alongside #1662 at this point in time, or do we want to an issue to track this and do this once 0.11 is released? If you expect or would like any action from me, let me know!

benlubas:
@MaxVanDeursen 

> calling parse in 0.11

I think we could/should do this before 0.11 releases. can just use vim.version to check versions. If you want to do this I will absolutely not stop you. I'm going to be really busy next week, but after that I'll have time and I can make the changes.

We also need to at some point either make the wiki generate but not push in PR, or make the wiki generation script fail (ie. don't push) if it doesn't generate the expected files. or both, so that something like this doesn't happen again.

Same deal here. I can do this, or if you're interested, I will not say no :)

PR #1685, opened 2025-04-22
fix(esupports.indent): keep indentation of verbatim blocks without an injected TS parser

ALVAROPING1:
This fixes `=` messing with the indentation of verbatim blocks when no language is injected (either because a parser doesn't exist/isn't installed, or because no language was specified). Sometimes this is used for diagrams, which get messed up if all lines are aligned with the tag start. The current implementation doesn't handle code blocks with `norg` as language (which i can't see a good use case for) and thus those won't get auto indented either

https://github.com/user-attachments/assets/c6d021cb-865c-4c3f-8df6-254d74096418

### DIFF
--- a/lua/neorg/modules/core/esupports/indent/module.lua
+++ b/lua/neorg/modules/core/esupports/indent/module.lua
@@ -149,7 +149,7 @@ module.public = {
     ---@param new_indent number
     buffer_set_line_indent = function(buffer, start_row, new_indent)
         local line = vim.api.nvim_buf_get_lines(buffer, start_row, start_row + 1, true)[1]
-        if line:match("^%s*$") then
+        if line:match("^%s*$") or new_indent == -1 then
             return
         end
 
--- a/lua/neorg/modules/core/esupports/indent/module.lua
+++ b/lua/neorg/modules/core/esupports/indent/module.lua
@@ -96,6 +96,11 @@ module.public = {
             end
         end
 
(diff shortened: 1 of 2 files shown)


benlubas:
You're amazing for this. Testing this PR made me realize that the visual promotion is broken

If you visually select the contents of this block and press `>`
```
@code
one
  - two
  - three
@end
```
You get this:
```
@code
  one
- two
- three
@end
```

The first line is indented, and all the rest are given 0 indent.

This feels related to this PR. I wonder if you have a guess as to why this happens and could maybe include the fix? Otherwise I will probably dig into the code a bit myself

ALVAROPING1:
> You're amazing for this. Testing this PR made me realize that the visual promotion is broken
> 
> If you visually select the contents of this block and press `>`
> 
> ```
> @code
> one
>   - two
>   - three
> @end
> ```
> 
> You get this:
> 
> ```
> @code
>   one
> - two
> - three
> @end
> ```
> 
> The first line is indented, and all the rest are given 0 indent.
> 
> This feels related to this PR. I wonder if you have a guess as to why this happens and could maybe include the fix? Otherwise I will probably dig into the code a bit myself

i have found a couple different bugs involved in the behavior of `<`/`>`, but i think i know how to fix it

ALVAROPING1:
should be fixed now. I also fixed range promotion/demotion not auto-indenting the first line in the range

PR #1688, opened 2025-04-29
Add the ability to export norg files to HTML

adc613:
Mirrors the methods used by markdown  exporter to export files to HTML

Provides an option to apply a custom handler for norg range tags


benlubas:
First off, wow thank you for going through the effort to do this. This is massive.

Second off, I'm so sorry for the amount of bugs that I'm about to point out. For all I know this all exists for the markdown exporter too, I don't use it myself.

I'll order them roughly by importance I guess:

- illegal chars are not escaped, this includes `<`, `>` [others](https://stackoverflow.com/questions/7381974/which-characters-need-to-be-escaped-in-html). This matters a _lot_ b/c it makes it possible to generate illegal/invalid HTML
- List items with text that spans many lines become multiple list items 
image
- generic links need to be resolved so they can correctly link to headings. Currently you get a link with `#generic-headingname` but the heading id is `#heading1-headingname`, so the link will never work.
- Code block indentation is off. This is a common mistake, the text in a codeblock all shares leading indent, but the TS node for the content doesn't include the first leading indent, so you have to get it from the buffer to remove it, otherwise you end up with this 
image
- inline verbatim and math are represented with pre tags which are block by default, and so text ends up getting split up when it shouldn't
- inline link targets aren't given ids to link to, and they create new paragraphs which places text on different lines when it shouldn't be. example: image
- Definition and footnote links don't work
- code blocks that have unrecognized languages still get `class="python"` for some reason, they should just not have a language class
- no class for spoilers, just a span (doesn't really matter honestly)
- duplicate headings aren't handled so you can end up with a heading with the same ID twice (I think this is safe to ignore for now)

---

let me know if you want help on any of this, I'm very interested in this feature myself

benlubas:
I also think that it would be nice to have an option to strip external file links out of the HTML (they break anyway). My use case is for emails; ideally I'd be able to copy something from my notes that may link elsewhere in my notes without having to manually remove the link (b/c the link is useless to the email recipients).

But this is out of scope of the PR, I'll add this feature myself at a later date

benlubas:
The math in norg is normally latex, but some people use typst. I think just leaving the content alone is fine (certainly for the initial PR), as long as it's displayed inline. Here's an example of what I'm talking about: 

this: ``Inline `verbatim`, and math: $|\frac{1}{4}|$.``
generates this: 
```html
<p>
  Inline
<pre>verbatim</pre>, and math: <pre class="inline-math">\frac{1}{4}</pre>.
</p>
```
which looks like this:
image when it should look like this 
image


---

re: footnotes, I think no need to do anything fancy, just put them all at the bottom of the document after an `<hr/>` tag. They don't even need to be styled. And then give them ids and link to them.

---

Other stuff:

- I've noticed that the html escape doesn't effect text inside of math tags, ranged verbatim (code blocks), inline verbatim, or inside of bold/italic/etc.
- The way you handled generic links was surprising, I expected you'd  resolve the links (using the hop module) to get the correct type and then use that. So instead of `#generic-whatever` you would resolve `#whatever` and figure out "oh it's a heading two" and use `#heading2-whatever` for the link. I think this method is much cleaner, and less error prone in the long run (you just implement the resolution code once vs adding a generic span to every single link target), but this works for now
- Every code block has a class of "python" even when it's not python code

adc613:
I think I fixed most the issues you mentioned, but let me know if you find anything. It's difficult to test exhaustively test the exporter, I suspect there are more edge cases that I missed.

I agree using the hop method would be less error prone. If I have time later, I may look into it.

adc613:
Hi, I just thought I'd follow up and ask is this pull request worth pursuing? I'm happy to address comments if its a worthwhile change, but I didn't want too much time on this if its never going to be used.

I also ran into an issue with the hop module ([link](https://github.com/nvim-neorg/neorg/issues/1717)) when I tried to us hop for link generation in [this branch](https://github.com/adc613/neorg/tree/export.html.hop).

PR #1701, opened 2025-05-29
docs(README): explain lazy loading is not supported

gagath:
After trying enable lazy-loading for lazy.nvim users [1], it seems like lazy-loading causes problems sometimes and is not supported nor encouraged.

Explain below the lazy.nvim section why lazy-loading is discouraged so that no other attempts are made to propose lazy load by other contributors.

[1] https://github.com/nvim-neorg/neorg/pull/1700


benlubas:
merged with slightly different wording in #1703

thanks!

PR #1706, opened 2025-06-11
fix(rockspec): add `nvim-treesitter-legacy-api` dependency

NTBBloodbath:
The original `nvim-treesitter` package has been removed from luarocks a year ago, however there is a `nvim-treesitter-legacy-api` package which exposes the legacy module system API from `nvim-treesitter` that neorg requires in order to work properly, e.g. being able to install the norg parsers and interacting with them.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   integration-test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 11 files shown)


NTBBloodbath:
I'll probably also commit fixes to the cancelled/failing CIs here, I need to verify that all workflows run properly with the changes.

benlubas (changes requested):
our CI has broken in the past b/c we've used un-pinned versions, so let's just pin

I'm lacking context on the luarocks changes (old packages and whatnot), but I trust that you've tested.

PR #1707, opened 2025-06-13
fix(ci): update actions version

NTBBloodbath:
Some (if not most) workflows had ancient versions of the used actions, which were causing failures or had deprecation warnings (e.g. `actions/checkout@2`). Additionally, the luarocks version in use has been bumped in the docgen workflow in order to address a failure caused by luarocks <=3.11 using an huge Lua table as manifest instead of a JSON file.


NTBBloodbath:
Merging as trivial

PR #1710, opened 2025-06-13
chore: fix luacheck warnings

NTBBloodbath:
This should in theory fix the codebase checks CI failing


benlubas:
Might as well address them. They're going to cause problems down the road

benlubas:
#1776 supersedes this

PR #1718, opened 2025-07-04
Add clause for the nil case in the treesitter module

adc613:
The nil case wasn't being handled which led to thrown errors in some edge cases.


benlubas:
ah, good catch

Issue #1720, opened 2025-07-17
`journal toc update` returns errors because of the recent change in Treesiter API


muwizayeon:
### Prerequisites

- [x] I am using the latest stable release of Neovim
- [x] I am using the latest version of the plugin

### Neovim Version

v0.11.1

### Neorg setup

```
return {
  "nvim-neorg/neorg",
  lazy = false, -- Disable lazy loading
  version = "*",
  dependencies = {
    { "nvim-lua/plenary.nvim" },
    { "nvim-neorg/neorg-telescope" },
  },
  opts = {
    load = {
      -- Default Modules
      ["core.clipboard.code-blocks"] = {},
      ["core.esupports.hop"] = {},
      ["core.esupports.indent"] = {},
      ["core.esupports.metagen"] = {},
      ["core.itero"] = {},
      ["core.journal"] = {},
      ["core.keybinds"] = {
        ["config"] = {
          ["default_keybinds"] = false,
          ["hook"] = function(keybinds)
            keybinds.map("norg", "n", "<localleader>o", ":Neorg return<CR>", { silent = true })
            keybinds.map("norg", "n", "<localleader>c", ":Neorg toggle-concealer<CR>", { slient = true })
          end,
        },
      },
      ["core.looking-glass"] = {},
      ["core.pivot"] = {},
      ["core.promo"] = {},
      ["core.qol.toc"] = {},
      ["core.qol.todo_items"] = {},
      ["core.tangle"] = {},
      ["core.todo-introspector"] = {},
      ["core.ui.calendar"] = {},
      -- Other Modules
      ["core.concealer"] = {},
      ["core.dirman"] = {
        config = {
          workspaces = {
            mainWS = "~/Documents/neorg-roam",
            dailySchedule = "~/Documents/neorg-roam/schedule_mgmt",
          },
          default_workspace = "mainWS",
        },
      },
      ["core.export"] = {},
      ["core.export.markdown"] = {},
      ["core.latex.renderer"] = {},
      ["core.integrations.image"] = {},
      ["core.integrations.treesitter"] = {},
      ["core.integrations.telescope"] = {},
      ["core.syntax"] = {},
    },
  },
}
```

### Actual behavior

If I try to update the table of contents (TOC) of a journal with `Neorg journal toc update`, multiple error messages appear as the update attempts to retrieve the titles of journals. One of the messages is

```Lua
[...]

Issue #1678, opened 2025-04-03
Neorg cannot update journal TOC after upgrading Neovim to 0.11.0


saxon1964:
### Prerequisites

- [x] I am using the latest stable release of Neovim
- [x] I am using the latest version of the plugin

### Neovim Version

0.11.0

### Neorg setup

```
    opts = {
      load = {
        ["core.defaults"] = {},
        ["core.concealer"] = {
          config = {
            icon_preset = "basic",
          },
        },
        ["core.dirman"] = {
          config = {
            workspaces = {
              notes = "~/notes",
            },
            default_workspace = "notes",
          },
        },
        ["core.keybinds"] = {
          config = {
            default_keybindings = true,
            preset = "neorg"
          },
        },
        ["core.export"] = {
          config = {
            export_dir = "~/notes/export",
          }
        },
        ["core.integrations.treesitter"] = {},
        ["core.highlights"] = {
          config = {
            highlights = {
            },
          },
        },
        ["core.neorgcmd"] = {},

```

### Actual behavior

Command :Neorg journal toc update generates a sequence of errors, one error for every journal entry on the filesystem. Only buffer number is different. After all errors are reported the TOC is just an empty file.

```
Error executing vim.schedule lua callback: .../n  
eovim/0.11.0/share/nvim/runtime/lua/vim/treesitt  
er.lua:104: Buffer 17 must be loaded to create p  
arser  
stack traceback:  
[C]: in function 'error'  
.../neovim/0.11.0/share/nvim/runtime/lua/vim/tr  
eesitter.lua:104: in function 'get_parser' 
...ua/neorg/modules/core/integrations/treesitte  
r/module.lua:922: in function 'get_ts_parser'  
...ua/neorg/modules/core/integrations/treesitte  
r/module.lua:741: in function 'get_document_meta  
data'  
...vim/lazy/neorg/lua/neorg/modules/core/journa  
l/module.lua:257: in function 'get_title'  
...vim/lazy/neorg/lua/neorg/modules/core/journa  
l/module.lua:300: in function <...vim/lazy/neorg  
/lua/neorg/modules/core/journal/module.lua:298> 
```

### Expected behavior

Regular TOC file with years, months and date links.

### Steps to reproduce

:Neorg journal toc update

### Help

None

PR #1680, opened 2025-04-06
fix(journal): journal index generation

benlubas:
fix #1678

PR #1721, opened 2025-07-17
Fix(journal): Use bufload to ensure the journal TOC buffer is loaded

muwizayeon:
As I reported at #1720, there is an issue to update a table of contents of Neorg Journal. Updating `get_parser` function in nvim-treesitter is mainly suspected as a cause. I simply add a one line of the code to ensure loading a buffer before returning the meta.title.


muwizayeon:
I think that @benlubas already committed another suggestion (#1680) to resolve #1678. As I mentioned in #1720, I am a novice in the Lua language, so please disregard this pull request if it is not more suitable than #1678. I will wait for a version that successfully merges #1678.

muwizayeon:
As I confirmed that PR #1680 resolves issue #1720, I will close this PR. Thank you, @benlubas, for letting me know and for the opportunity to learn the Lua language!

PR #1733, opened 2025-08-17
fix(todo-instrospector): use combine hl_mode

austinliuigi:


austinliuigi:
Yeah, sorry about that. Before, the todo introspector extmarks replace any highlighting that would otherwise be there, including the CursorLine highlight.

See how the `[0/3] (0%)` in this screenshot has a background different from the cursorline:
image

With this fix, it now blends with the cursorline:
image

PR #1734, opened 2025-08-24
fix(core.latex.renderer): rewrite with snacks.nvim to fix inline rendering

### LINKED ISSUE #1725: incorrect inline math rendering
### Prerequisites

- [x] I am using the latest stable release of Neovim
- [x] I am using the latest version of the plugin

### Neovim Version

v0.11.3

### Neorg setup

return {
    {
        "nvim-neorg/neorg",
        version = "*",
        event = "VeryLazy",
        config = function()
            local mod = {
                ["core.defaults"] = {},
                ["core.highlights"] = {
                    rendered = {
                        latex = "@Normal",
                    },
                },
                ["core.concealer"] = {},
                ["core.export"] = {},

                ["core.latex.renderer"] = {
                    --conceal=false,
                    --render_on_enter = true
                },
                ["core.dirman"] = {
                    config = {
                        workspaces = {
                            ai = "~/Documents/notes/ai",
                        },
                        index = "index.norg", -- The name of the main (root) .norg file
                    },
                },
            }

            if not vim.g.neovide then
                mod["core.integrations.image"] = {}
            end
      
            require("neorg").setup({
                load = mod,
                engine = "nvim-cmp",
            })
        end,
    },
}

### Actual behavior

Image

```norg
Lorem ipsum dolor sit $|\alpha|$, consectetur adipisicing elit

Lorem ipsum dolor sit $|\alpha|$, consectetur adipisicing elit

Lorem ipsum dolor $| \theta |$ amet, consectetur $| \alpha |$ elit, sed  $| \gamma |$ s
```
- inline math in first line didn't show up
- inline math is not in correct place

### Expected behavior

inline math should be in their right place

### Steps to reproduce

1. setup neorg
2. open kitty or wezterm
3. edit a norg file and write some inline math
4. `Neorg render-latex enable`

_No response_

### Other information

_No response_

### Help

None

### Implementation help

_No response_

imli700:
**Description:**

This pull request resolves issue #1725 by rewriting the `core.latex.renderer` module to use `snacks.nvim` instead of `image.nvim`.

**Problem:**

The previous implementation used `image.nvim` and a `virt_text` padding strategy to render inline LaTeX. This was unreliable and often caused the rendered math image to overlap with surrounding text, especially when the math was not on its own line. The image background was also not consistently transparent.

**Solution:**

Following the suggestion from @benlubas in the issue, this module has been refactored to use `snacks.nvim`. `snacks` utilizes a more robust rendering technique based on unicode placeholders, which correctly positions inline images without overlapping text.

**Key Changes:**

*   Replaced all `image.nvim` integration logic with the `snacks.nvim` placement API.
*   The module now internally configures `snacks.nvim`'s ImageMagick arguments to ensure rendered LaTeX has a **transparent background** by default.
*   The module now requires `pysan3/snacks.nvim` as a dependency.
*   Cursor-line concealment is handled by using the `:hide()` and `:show()` methods from `snacks` placements, improving user experience during editing.
*   The old `dpi` option has been replaced with `density`, which is passed to ImageMagick during conversion.

This change provides a much more stable and visually correct "out-of-the-box" experience for inline LaTeX rendering in Neorg.

Closes #1725

### DIFF
--- a/lua/neorg/modules/core/latex/renderer/module.lua
+++ b/lua/neorg/modules/core/latex/renderer/module.lua
@@ -144,7 +144,6 @@ module.public = {
                 local original_snippet = module.required["core.integrations.treesitter"].get_node_text(node, buf)
                 
                 -- FIX: Only clean the surrounding pipe delimeters ($|...|$).
-                -- We do NOT strip backslashes anymore, because that breaks LaTeX commands like \hat and \frac.
                 local clean_snippet = original_snippet:gsub("^%$|", "$"):gsub("|%$$", "$")
                 
                 if #clean_snippet - 2 < module.config.public.min_length then return end


imli700:
Hi @benlubas,

I've just pushed a new commit that adds support for multi-line math blocks, completing the feature set for the renderer.

**Key Enhancements in the new commit:**

*   **Multi-line Math Blocks:** The renderer now fully supports the `@math ... @end` syntax for multi-line equations.
*   **Indentation-Aware Placement:** Instead of trying to center the blocks (which caused a lot of headaches), they respect the document's indentation. The placement logic finds the indentation of the content within the `@math` block and aligns the rendered image accordingly. This feels much more natural and robust.

imli700:
I know I said I couldn't, but I fixed it:
image

Both in-line and block math renders normally:
image

benlubas (changes requested):
I really appreciate the effort, but I can't merge this.

I'm going to ask that we separate this out into it's own "renderer" module the same way the current image.nvim integration is setup. This way users currently using image.nvim don't break, and future users have the option to choose.

If this requires some changes to the actual `/latex/renderer/module.lua` file, let's discuss them. But theoretically, you don't have to change this file at all.

There are also some mistakes (along with the entirety of the PR missing that we've factored out the image rendering mechanism from the image placement logic) that make me think some or most of this PR is AI generated. Is that correct?

Vaisakhkm2625:
few years ago I assumed terminal math rendering is an easy topic.. 

now this went from hologram->image.nvim->snacks and become a multi-year sub project 😂
we can still see vhyrro old   [#latex-rendering](https://github.com/nvim-neorg/neorg/tree/latex-rendering) branch with hologram in it staying dormant..

benlubas:
I understand. Sorry we couldn't accept this as is. Hopefully this review can serve as a guide for whoever attempts this next

benlubas:
Hi, snacks has its own implementation of latex rendering for norg documents.

You can use that.

voidarclabs:
Sorry to be a bother, but is there any guides that you can point to? Neither the PR adding the feature to snacks nor the actual snacks docs explain how to set it up so that it works automatically

PR #1735, opened 2025-09-11
fix(concealer): apply correct extmark opts

austinliuigi:
This fixes two issues with the extmarks added by the concealer:

- hides the extmarks when they flow off the screen due to horizontally scrolling
- removes highlights on the extmarks that pads the left of headings/lists with whitespace

**Demonstration of the wrapping issue**: Notice how the extmarks stay on the left side of the screen even as the screen scrolls to the right. Also notice how the whitespace before the higher level headings has the heading highlight applied to it.

_Before_:

https://github.com/user-attachments/assets/d22c609e-4fff-43cb-99f2-1cd1c8206438

_After_:

https://github.com/user-attachments/assets/5dd93d17-8295-4a1b-b3cb-34163489abdd

### DIFF
--- a/lua/neorg/modules/core/concealer/module.lua
+++ b/lua/neorg/modules/core/concealer/module.lua
@@ -466,13 +466,18 @@ module.public = {
                     return
                 end
 
-                local text = (" "):rep(len - 1) .. icon
-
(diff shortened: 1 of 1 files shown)


benlubas:
I think arguably the heading whitespace having the heading highlight is intentional/desirable. Can we make it configurable?

Issue #1715, opened 2025-06-18
Unable to load nvim-treesitter.ts_utils


serpro69:
### Prerequisites

- [x] I am using the latest stable release of Neovim
- [x] I am using the latest version of the plugin

### Neovim Version

0.11.2

### Neorg setup

```lua
return {
  "nvim-neorg/neorg",
  lazy = false,
  version = "*",
  opts = {
    load = {
      ["core.defaults"] = {}, -- Loads default behaviour
      ["core.concealer"] = { config = { folds = true, icon_preset = "varied" } }, -- Adds pretty icons to your documents
      ["core.keybinds"] = {
        config = {
          neorg_leader = ",",
        },
      },
      ["core.dirman"] = {
        config = {
          workspaces = {
            notes = "~/notes",
          },
          default_workspace = "notes",
        },
      },
    },
  },
}
```

### Actual behavior

![Image](https://github.com/user-attachments/assets/89d30278-8702-4780-ad22-3726736ad9b0)

### Expected behavior

Should not throw errors

### Steps to reproduce

Switch to `main` branch of nvim-treesitter and nvim-treesitter-textobjects

### Help

None

IpadDilectio:
I think I have the same problem.

timhugh:
I ran into this as well. It's only an issue on the `main` branch of treesitter which introduced breaking API changes, but the `master` branch still works. I don't know the scope of the changes or how neorg interacts with treesitter, but the breaking changes are noted at the top of the treesitter readme: https://github.com/nvim-treesitter/nvim-treesitter?tab=readme-ov-file#nvim-treesitter

sheldonj84:
FYI: [LazyVim](https://www.lazyvim.org/news) just went to the `main` branch of treesitter which is going to break a bunch of peoples neorg

Necrom4:
@sheldonj84 Have you found a solution yet? l personally tried installing Neorg with Rocks but was unsuccesful, so I thought of pinning treesitter to the master branch but it conflicts with the LazyVim config that pins to the main branch.

imli700:
I'm willing to pay if necessary

benlubas:
Sounds like a lazy vim problem to me. They're shipping a broken plugin.

But yeah we should switch over the main. It's about time.

benlubas:
You can obviously downgrade lazy vim in the meantime.

But yes, we absolutely do need to upgrade.

benlubas:
[...]

PR #1738, opened 2025-09-19
[workaround-ts-main] workaround: copy nvim-treesitter utils for main …

randoentity:
…branch compatibility

ts_utils.lua and utils.lua were copied from nvim-treesitter

Just a quick hacky workaround to get things working for now. Relevant issue: https://github.com/nvim-neorg/neorg/issues/1715


benlubas:
Thank you for the workaround. I will leave this here for people to use. I plan to properly fix the issue this weekend.

PR #1816, opened 2026-05-11
feat: add exclude_dirs option for summary module and corresponding docs

manos00:
Hey everyone! I've had an issue with the summary module: I like using the journal module for daily to-dos, but thought they were kind of bloating the auto generated summary in my Index file. When I looked at the documentation, I only found the option to only include certain categories: 
> If arguments are provided then a partial summary is generated containing only categories that you have provided.

This however did not fit my use case. I thought it would be great to have the option to specifically exclude certain categories or folders. Implementing the option to exclude folders however seems more sensible to me, as it provides the most value, since it is the most different from everything implemented so far (e.g. the inclusion of only specific categories).

This perfectly fits my needs (as the journal files typically are within another folder) and is much easier than what I used to do (add the `journal` category to all journal files and then delete the journal list from the generated summary).
Albeit I implemented this as a solution to a specific problem, I can also see more general use cases for an option like this (e.g. excluding directories like  `scratchpad/`, `templates/`, `archive/` or `old_projects/`).
Let me know if there is anything I should change and thanks a lot for your work on Neorg! :)

### DIFF
--- a/lua/neorg/modules/core/summary/module.lua
+++ b/lua/neorg/modules/core/summary/module.lua
@@ -19,6 +19,8 @@ The way the summary is generated relies on the `strategy` configuration option,
 which by default consults the document metadata (see also
 [`core.esupports.metagen`](@core.esupports.metagen)) or the first heading title
 as a fallback to build up a tree of categories, titles and descriptions.
+The module also provides the option to exclude files from certain directories 
+from the summary using the `exclude_dirs` configuration option.
 --]]
 
 local neorg = require("neorg.core")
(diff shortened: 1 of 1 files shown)


manos00:
Now that I think about it: I think it would be good idea to omit the check, if the excluded directories exist because it adds no functionality, while adding another import. It stems from an earlier revision of this implementation. Let me know what you think :)

max397574:
just wondering
can't we these days use builtin neovim stuff for most things instead of pathlib?

though I also don't know how much of the functionality there is and if we're supporting neovim versions where this isn't there yet
@vhyrro

manos00:
I also just thought about using this option to exclude singular files, which could be useful (e.g to remove the `index.norg` file from the summary).

PR #1819, opened 2026-05-26
refactor: more straight-forward event handling

max397574:


max397574:
the a bit more extreme (but imo better) alternative would be https://github.com/nvim-neorg/neorg/pull/1063

PR #1820, opened 2026-06-13
fix(export/html): use link_description for anchor content

YsuOS:
The wrap_anchor() recollector used output[1] as the display text, but output ordering differs between link ({url}[text]) and anchor_definition ([text]{url}) nodes. Use state.link.link_description which is consistent regardless of syntax. Also fix the fallback from the nonexistent link_text field to link_location_text.


YsuOS:
Issue: https://github.com/nvim-neorg/neorg/issues/1822

PR #1821, opened 2026-06-13
fix: typos in core/export/html module

YsuOS:
Fix typos in comments (stackk, differnete, givne, specically, hte, could so here), variable names (link_desription), function names (add_closing_segement_tags), and an HTML class name (footnore-content).


NTBBloodbath:
Thank you so much for addressing this! :)

PR #1823, opened 2026-06-15
fix(export/markdown): avoid over-stripping indentation in ranged tags

YsuOS:
Use math.min to clamp the strip amount to the actual leading whitespace, preventing striping non-whitespace characters. when a line has less indentation than tag_indent.

### DIFF
--- a/lua/neorg/modules/core/export/markdown/module.lua
+++ b/lua/neorg/modules/core/export/markdown/module.lua
@@ -492,7 +492,9 @@ module.public = {
                 split_text[1] = string.rep(" ", ranged_tag_content_column_start - state.tag_indent) .. split_text[1]
 
                 for i = 2, #split_text do
-                    split_text[i] = split_text[i]:sub(state.tag_indent + 1)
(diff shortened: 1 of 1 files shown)


YsuOS:
# Test result
- Src norg file
```
$ cat /tmp/test.norg
* Title
  - bullet point
  @code
ABCDEFG
abcdefg
  @end
  @code
  ABCDEFG
  abcdefg
  @end
  @code
    ABCDEFG
    abcdefg
  @end
```
- Actual result: the "ab" is stripped in the first code block because this line start from column 0 but the tag_indent is 2.
````
$ cat /tmp/test-before.md
# Title

- bullet point
```
ABCDEFG
cdefg  <<<--- HERE!!
```
```
ABCDEFG
abcdefg
```
```
  ABCDEFG
  abcdefg
```
````
- Expected result (with this patch)
````
$ cat /tmp/test-after.md
# Title

- bullet point
```
ABCDEFG
abcdefg
```
```
ABCDEFG
abcdefg
```
```
  ABCDEFG
  abcdefg
```
````

